PR #6, opened 2018-04-06
Switch to cereal for serialization.

akleeman:
The first approach for serialization directly used `yaml` and `json` libraries, but `cereal` is far more flexible.

This is vaguely broken up into two steps, first I removed the existing dependences, then switched to using `cereal` and made the corresponding required changes.

Noteworthy changes:
- Any existing `to_string` methods were removed in favor of `pretty_string` to make it clear that these methods are for human readable representations, not serialized strings.
- Serialization now happens through helper functions in `serialize.h` such as `to_json(T)` and `from_json(T, string)`.


albatross/core/model.h
```diff
@@ -63,15 +67,15 @@ using FoldIndexer = std::map<FoldName, FoldIndices>;
  * A combination of training and testing datasets, typically used in cross
  * validation.
  */
-template <class Feature>
+template <class FeatureType>
 struct RegressionFold {
-  RegressionDataset<Feature> train;
-  RegressionDataset<Feature> test;
+  RegressionDataset<FeatureType> train;
```
kmdade: nit: `train` and `test` are verbs. maybe `train_data` and `test_data`

albatross/core/cereal_utils.h
```diff
@@ -0,0 +1,96 @@
...
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef ALBATROSS_CEREAL_UTILS_H
+#define ALBATROSS_CEREAL_UTILS_H
+
```
kmdade: NIT: Maybe this file is `cereal_eigen.h`? It serves a very clear purpose currently and it might be nice to codify that in the name.

albatross/core/model.h
```diff
@@ -63,35 +65,47 @@ using FoldIndexer = std::map<FoldName, FoldIndices>;
...
 /*
  * A model that uses a single Feature to estimate the value of a double typed
  * target.
  */
-template <class Feature>
+template <typename FeatureType>
 class RegressionModel : public ParameterHandlingMixin {
  public:
-  RegressionModel() : ParameterHandlingMixin(), has_been_fit_(false){};
+  typedef FeatureType Feature;
```
kmdade: doesn't look like you are using this type alias anywhere.

albatross/core/model.h
```diff
@@ -63,35 +65,47 @@ using FoldIndexer = std::map<FoldName, FoldIndices>;
...
+  RegressionModel() : ParameterHandlingMixin(), has_been_fit_() {};
   virtual ~RegressionModel(){};
 
+  template <typename OtherFeatureType>
+  bool operator == (const RegressionModel<FeatureType> &other) const {
+    return false;
+  }
+
+  bool operator == (const RegressionModel<FeatureType> &other) const {
```
kmdade: are they equal if they `has_been_fit()` on different datasets?

albatross/core/model.h
```diff
@@ -112,8 +126,8 @@ class RegressionModel : public ParameterHandlingMixin {
    * properly sized PredictionDistributions.
    */
   PredictionDistribution predict(
-      const std::vector<Feature> &features) const {
-    assert(has_been_fit_);
+      const std::vector<FeatureType> &features) const {
+    assert(has_been_fit());
```
kmdade: assert is fine for now, but in the long run this should be an acceptable failure mode and handled gracefully.

albatross/covariance_functions/covariance_functions.h
```diff
@@ -24,6 +24,10 @@ template <typename Term>
 struct CovarianceFunction {
   Term covariance_term;
 
+  CovarianceFunction() : covariance_term() {};
+  CovarianceFunction(Term &term) : covariance_term(term) {};
+  CovarianceFunction(Term &&term) : covariance_term(term) {};
```
kmdade: Do you need both of these? I think defining the & version should get you the && version via implicit casting. I could be wrong though.

albatross/core/model.h
akleeman: That's true for `albatross`, but it is being used in our private repo.

albatross/core/model.h
akleeman: Yeah I guess right now they are ... which definitely isn't ideal.  I'll remove this class level operator and switch it to a unit test specific equality function.

albatross/core/cereal_utils.h
akleeman: I moved it to `/cereal/eigen.h`.  That sound good?

albatross/core/model.h
akleeman: actually, I decided to change the equality to fail (hard) if the model has been fit and is being equality compared.  Idea being that if anyone is implementing a `RegressionModel` comparison will work fine as long as the model hasn't been fit (in which case parameter comparison should suffice) but if they want their model to be comparable after having been fit they'll need to write a custom equality operator which knows to compare any custom member variables which may have been modified during the fit.

albatross/covariance_functions/covariance_functions.h
akleeman: yep, you're right.  Thought to get it to work I had to make the `&` version `const`.

PR #8, opened 2018-04-18
Add parameter optimization.

akleeman:
Added the ability to create an arbitrary model, provide a dataset and an evaluation metric and be able to find the parameters which minimize the metric.

Includes:
- Addition of `tune.h` which deals with numerical optimization of model hyper parameters.
- A likelihood evaluation metric which can be used to do maximum likelihood optimization.
- Creates a new example `tune_example.cc`.

### DIFF
--- a/albatross/tune.h
+++ b/albatross/tune.h
@@ -23,20 +23,17 @@ namespace albatross {
 std::vector<ParameterValue> transform_parameters(
     const std::vector<ParameterValue>& x) {
   std::vector<ParameterValue> transformed(x.size());
-  for (std::size_t i = 0; i < x.size(); i++) {
-    transformed[i] = log(x[i]);
-  }
+  std::transform(x.begin(), x.end(), transformed.begin(), log);
   return transformed;
 }
 
 std::vector<ParameterValue> inverse_parameters(
     const std::vector<ParameterValue>& x) {
   std::vector<ParameterValue> inverted(x.size());
-  for (std::size_t i = 0; i < x.size(); i++) {
-    inverted[i] = exp(x[i]);
-  }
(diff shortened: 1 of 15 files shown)


albatross/tune.h
```diff
@@ -0,0 +1,118 @@
...
+#include <map>
+#include <vector>
+#include "core/model.h"
+#include "nlopt.hpp"
+
+namespace albatross {
+
+std::vector<ParameterValue> transform_parameters(
+    const std::vector<ParameterValue>& x) {
+  std::vector<ParameterValue> transformed(x.size());
```
benjamin0: nit: you could use `std::for_each` here and in the following functions below.

albatross/tune.h
```diff
@@ -0,0 +1,118 @@
...
+
+  TuneModelConfg(const RegressionModelCreator<Predictor> &model_creator_,
+                 const RegressionDataset<Predictor> &dataset_,
+                 const TuningMetric<Predictor> &metric_)
+      : model_creator(model_creator_), dataset(dataset_), metric(metric_){};
+};
+
+template <class Predictor>
+double objective_function(const std::vector<double> &x,
+                          std::vector<double> &grad, void *void_tune_config) {
```
benjamin0: Is `grad` const? or is this the nlopt api?

akleeman: That's the `nlopt` api.  If the optimization algorithm you choose requires gradients you're expected to have filled them in hence the non const.

albatross/tune.h
akleeman: done (though I had to use `transform` instead of `for_each` to avoid a copy).

PR #9, opened 2018-04-19
Generalize the AdaptedRegressionModel.

akleeman:
Generalize the AdaptedRegressionModel  so it defaults to deriving from RegressionModel, but could optionally derive from anything that has extended RegressionModel.

This is allows you to do things like:

```
class FancyRegressionModel : public RegressionModel<X> {
    ...something...
}

template <typename SubModelType>
using AdaptedFancyRegressionModel = AdaptedRegressionModel<X, SubModelType, FancyRegressionModel>
```


benjamin0:
looks good to me

PR #10, opened 2018-04-25
Allow adapted models to be serialized.

akleeman:
Give AdaptedRegressionModel the ability to extend anything that is derived from RegressionModel.  In particular this lets AdaptedRegressionModel wrap a SerializableRegressionModel and proives the appropriate cereal::base_class paths which allow polymorphic serialization.
    
This required using traits to conditionally add required pure virtual functions when an AdaptedRegressionModel extends a SerializableRegressionModel.


akleeman:
@kmdade yeah, I like that idea.

PR #12, opened 2018-05-02
Add ability specify measurement noise of the targets.

akleeman:


albatross/core/distribution.h
```diff
@@ -0,0 +1,77 @@
...
+
+  std::size_t size() const {
+    // If the covariance is defined it must have the same number
+    // of rows and columns which should be the same size as the mean.
+    assert_valid();
+    return mean.size();
+  }
+
+  void assert_valid() const {
+    if (covariance.size() > 0) {
```
benjamin0: Could we make `CovarianceType` optional?

albatross/core/indexing.h
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return out;
+}
+
+/*
+ * Takes a symmetric subset of an Eigen::Matrix.  Ie, it'll index the same rows and
+ * columns.
+ */
+template <typename SizeType>
+Eigen::MatrixXd subset(const std::vector<SizeType> &indices, const Eigen::MatrixXd &v) {
```
benjamin0: nit: name as `symmetric_subset`?

albatross/core/indexing.h
```diff
@@ -0,0 +1,89 @@
...
+                                           const Eigen::DiagonalMatrix<Scalar, Size> &v) {
+  return subset(indices, v.diagonal()).asDiagonal();
+}
+
+/*
+ * Extracts a subset of an Eigen::Matrix for the given row and column
+ * indices.
+ */
+template <typename SizeType>
+Eigen::MatrixXd subset(const std::vector<SizeType> &row_indices,
```
benjamin0: So this is for extracting non contiguous elements of the matrix? Eigen has `.block()`

akleeman: Right, as far as I could tell `.block()` was for contiguous, here we want to be able to pull out arbitrary chunks the way we do in cross validation.

albatross/core/distribution.h
akleeman: I was trying to weigh the benefit of this approach versus including another submodule.  Think it's worth adding https://github.com/akrzemi1/Optional ?

albatross/core/indexing.h
akleeman: yeah good idea.  meanwhile, do you think `subset` is a good name?  I could stick with `block` or `noncontiguous_block` or something along those lines.

albatross/core/distribution.h
benjamin0: Yeah, probably not necessary at the moment, but something to think about in the future.

akleeman: yeah I agree.  If we run into another situation where `optional` would be nice I think it'd be worth adding it (and changing this).

PR #13, opened 2018-05-09
Add ability to tune a model using multiple datasets.

akleeman:
This provides a way to run parameter optimization on a model by aggregating tuning metrics across multiple datasets.

In pseudocode, it lets us minimize a metric along the lines of:
```
double multiple_dataset_metric() {
  vector<double> metric_per_dataset;
  for (dataset : vector_of_datasets) {
    one_metric = compute_metric_for_single_dataset(dataset)
    metric_per_dataset.push_back(one_metric)
  }
  return compute_mean_of_all_metrics(metric_per_dataset);
}
```
where before we were only capable of minimizing:
```
compute_metric_for_single_dataset(dataset)
```

### DIFF
--- a/albatross/tune.h
+++ b/albatross/tune.h
@@ -20,36 +20,71 @@
 
 namespace albatross {
 
-std::vector<ParameterValue>
+inline std::vector<ParameterValue>
 transform_parameters(const std::vector<ParameterValue> &x) {
   std::vector<ParameterValue> transformed(x.size());
   std::transform(x.begin(), x.end(), transformed.begin(), log);
   return transformed;
 }
 
-std::vector<ParameterValue>
+inline std::vector<ParameterValue>
 inverse_parameters(const std::vector<ParameterValue> &x) {
   std::vector<ParameterValue> inverted(x.size());
   std::transform(x.begin(), x.end(), inverted.begin(), exp);
   return inverted;
 }
 
-template <class Predictor>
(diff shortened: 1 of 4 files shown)


albatross/tune.h
```diff
@@ -20,76 +20,111 @@
...
+                 const TuningMetric<FeatureType> &metric_,
+                 const TuningMetricAggregator &aggregator_ = mean_aggregator,
+                 std::ostream &output_stream_ = std::cout)
+      : model_creator(model_creator_), datasets({dataset_}), metric(metric_),
+        aggregator(aggregator_), output_stream(output_stream_){};
+
+  TuneModelConfg(const RegressionModelCreator<FeatureType> &model_creator_,
+                 const std::vector<RegressionDataset<FeatureType>> &datasets_,
+                 const TuningMetric<FeatureType> &metric_,
+                 const TuningMetricAggregator &aggregator_ = mean_aggregator,
```
benjamin0: Do we want to use a `mean_aggregator` over all the metrics? Seems like using median might be more robust here.

akleeman: Yeah I think median would be more robust, but that is probably not desirable for a default tuning metric.  For example, one might be tuning using negative log likelihood and with the median you may end up with some model that performs better on 51% of the data but really horribly on 49%.  I think in general such a model would be less desirable than one that performed well on average.

PR #15, opened 2018-05-14
Model Adapter Params

akleeman:
This changes lets an `AdaptedModel` take on higher level parameters.  This is useful for situations where converting from feature `X` to feature `Y` might change depending on some model parameter.

An example of this is given in the new added test,

```
class TestAdaptedModel : public TestAdaptedModelBase {
public:
  TestAdaptedModel() { this->params_["center"] = 0.; };

  const Eigen::VectorXd convert_feature(const double &x) const {
    Eigen::VectorXd converted(2);
    converted << 1., (x - this->params_.at("center"));
    return converted;
  }
```

### DIFF
--- a/albatross/core/model_adapter.h
+++ b/albatross/core/model_adapter.h
@@ -96,7 +96,9 @@ public:
 
   bool has_been_fit() const override { return sub_model_.has_been_fit(); }
 
-  ParameterStore get_params() const override { return sub_model_.get_params(); }
+  ParameterStore get_params() const override {
+    return map_join(this->params_, sub_model_.get_params());
+  }
 
   template <class Archive> void save(Archive &archive) const {
     archive(cereal::make_nvp(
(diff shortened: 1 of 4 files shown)


albatross/core/model_adapter.h
```diff
@@ -112,7 +114,12 @@ class AdaptedRegressionModel
 
   void unchecked_set_param(const std::string &name,
                            const double value) override {
-    sub_model_.set_param(name, value);
+
+    if (map_contains(this->params_, name)) {
+      this->params_[name] = value;
```
benjamin0: nit: What's the difference between `params_[name] = value` and `.set_param(name, value)`?

akleeman: `.set_param(name, value)` checks to make sure the parameter exists, then calls `unchecked_set_param`: https://github.com/swift-nav/albatross/blob/master/albatross/core/parameter_handling_mixin.h#L74-L77.  so if I called it from here we'd get a circular reference.

PR #16, opened 2018-05-22
Clang compatibility

seth-swiftnav:
Three separate commits for compatibility of Albatross with the Clang compiler.


akleeman:
looks great!

PR #18, opened 2018-05-24
Covariance matrix directly from CovarianceTerm.

akleeman:
Add a wrapper around covariance matrix functions to allow them to be called directly with a CovarianceTerm.

### DIFF
--- a/albatross/covariance_functions/covariance_functions.h
+++ b/albatross/covariance_functions/covariance_functions.h
@@ -22,6 +22,7 @@
 namespace albatross {
 
 template <typename Term> struct CovarianceFunction {
+  using TermType = Term;
   Term term;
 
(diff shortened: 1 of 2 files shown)


albatross/covariance_functions/covariance_functions.h
```diff
@@ -22,6 +22,7 @@
 namespace albatross {
 
 template <typename Term> struct CovarianceFunction {
+  using TermType = Term;
```
benjamin0: Is this used? Can we just change the template param name?

akleeman: It's used to be able to access the template parameter from an instantiation of a `CovarianceFunction`.  Ie, to be able to do: `CovarianceFunction::TermType new_term;`

PR #19, opened 2018-05-25
Angular Distance and Exponential Covariance

akleeman:
Add and an angular distance metric and exponential radial term.

### DIFF
--- a/albatross/covariance_functions/distance_metrics.h
+++ b/albatross/covariance_functions/distance_metrics.h
@@ -30,7 +30,6 @@ protected:
 
 class EuclideanDistance : public DistanceMetric {
 public:
-  EuclideanDistance(){};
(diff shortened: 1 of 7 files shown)


albatross/covariance_functions/distance_metrics.h
```diff
@@ -52,6 +52,18 @@ class RadialDistance : public DistanceMetric {
   }
 };
 
+class AngularDistance : public DistanceMetric {
+public:
+  AngularDistance(){};
```
benjamin0: nit: Do we need to define the default constructor?

akleeman: Nice catch, right we don't need it since it's inherited from `DistanceMetric`.

PR #21, opened 2018-06-07
Serializable LDLT Decomposition.

akleeman:
Adds an extention of Eigen's LDLT which is serializable and only serializes the lower triangular portion.

Changes include:
- Generalization of the existing Matrix serialization routines to allow different scalar types.  What were previously separate serializations routines for `Matrix3d` and `MatrixXd` are now handled in a single case which also allows for different internal types.
- Addition of the `SerializableLDLT` class which wraps `LDLT<MatrixXd>`.
- Use of `SerilizableLDLT` in `GaussianProcessFit`.
- Extended `test_serilalize` to test more than just the models.

### DIFF
--- a/albatross/cereal/eigen.h
+++ b/albatross/cereal/eigen.h
@@ -20,65 +20,81 @@
 
 namespace cereal {
 
-template <class Archive>
-inline void save(Archive &archive, const Eigen::VectorXd &v) {
-  size_type s = static_cast<size_type>(v.size());
-  archive(cereal::make_size_tag(s));
-  for (size_type i = 0; i < s; i++) {
-    archive(v[i]);
+/*
+ * The subsequent save/load methods catch the serialization methods
+ * for arbitrary Eigen::Matrix* types.  The general idea is that each
+ * row is saved as a Eigen::Vector.
+ */
+template <class Archive, typename _Scalar, int _Rows, int _Cols>
+inline void save(Archive &archive,
+                 const Eigen::Matrix<_Scalar, _Rows, _Cols> &v) {
+  size_type rows = static_cast<size_type>(v.rows());
+  archive(cereal::make_size_tag(rows));
+  for (size_type i = 0; i < rows; i++) {
+    Eigen::Matrix<_Scalar, _Cols, 1> row = v.row(i);
+    archive(row);
   }
 };
 
-template <class Archive>
-inline void load(Archive &archive, Eigen::VectorXd &v) {
-  size_type s = static_cast<size_type>(v.size());
-  archive(cereal::make_size_tag(s));
-  v.resize(s);
-  for (size_type i = 0; i < s; i++) {
-    archive(v[i]);
+template <class Archive, typename _Scalar, int _Rows, int _Cols>
+inline void load(Archive &archive, Eigen::Matrix<_Scalar, _Rows, _Cols> &v) {
+  size_type rows;
+  archive(cereal::make_size_tag(rows));
+  /*
+   * In order to determine the size of a matrix, we have to first determine
+   * how many rows, then inspect the size of the first row to get the
+   * number of columns.
+   */
+  if (rows > 0) {
+    Eigen::Matrix<_Scalar, _Rows, 1> first;
+    archive(first);
+    size_type cols = first.rows();
+    v.resize(rows, cols);
+    v.row(0) = first;
+
+    for (size_type i = 1; i < rows; i++) {
+      Eigen::Matrix<_Scalar, _Cols, 1> row;
+      archive(row);
+      v.row(i) = row;
+    }
+  } else {
+    // Serialized matrix is empty.
+    v.resize(0, 0);
   }
 };
 
 /*
- * For length 3 vectors, we wrap the length L call.
+ * The subsequent save/load methods catch the serialization methods
+ * for arbitrary Eigen::Vector* types through template specialization.
+ * In this case each scalar value is serialized.
  */
-template <typename Archive>
-inline void save(Archive &archive, const Eigen::Vector3d &v) {
-  Eigen::VectorXd as_xd(v.size());
-  as_xd << v;
-  save(archive, as_xd);
-}
-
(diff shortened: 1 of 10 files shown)


albatross/cereal/eigen.h
```diff
@@ -20,66 +20,82 @@
...
-template <class Archive>
-inline void save(Archive &archive, const Eigen::VectorXd &v) {
-  size_type s = static_cast<size_type>(v.size());
-  archive(cereal::make_size_tag(s));
-  for (size_type i = 0; i < s; i++) {
-    archive(v[i]);
+/*
+ * The subsequent save/load methods catch the serialization methods
+ * for arbitrary Eigen::Matrix* types.  The general idea is that each
+ * row is saved as a Eigen::Vector.
```
benjamin0: How was it done before?

albatross/cereal/eigen.h
```diff
@@ -20,66 +20,82 @@
...
-  archive(cereal::make_size_tag(s));
-  v.resize(s);
-  for (size_type i = 0; i < s; i++) {
-    archive(v[i]);
+template <class Archive, typename _Scalar, int _Rows, int _Cols>
+inline void load(Archive &archive, Eigen::Matrix<_Scalar, _Rows, _Cols> &v) {
+  size_type rows;
+  archive(cereal::make_size_tag(rows));
+  /*
+   * In order to determine the size of a matrix, we have to first determine
```
benjamin0: Would it be easier to serialize the dimensions instead of needing to scan through?

albatross/eigen/serializable_ldlt.h
```diff
@@ -0,0 +1,118 @@
...
+template <class Archive, typename _Scalar, int _Rows, int _Cols>
+inline void load_lower_triangle(Archive &archive,
+                                Eigen::Matrix<_Scalar, _Rows, _Cols> &v) {
+  cereal::size_type storage_size;
+  archive(cereal::make_size_tag(storage_size));
+  // TODO: understand why the storage size upon load is always augmented by two.
+  storage_size -= 2;
+  // We assume the matrix is square and compute the number of rows from the
+  // storage size.
+  double drows = (std::sqrt(static_cast<double>(storage_size * 8 + 1)) - 1) / 2;
```
benjamin0: Can we name these numbers so it's clearer what's going on?

albatross/eigen/serializable_ldlt.h
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+class SerializableLDLT : public LDLT<MatrixXd, Lower> {
+public:
+  SerializableLDLT() : LDLT<MatrixXd, Lower>(){};
+
+  SerializableLDLT(const LDLT<MatrixXd, Lower> &ldlt)
+      // Can we get around copying here?
+      : LDLT<MatrixXd, Lower>(ldlt){};
```
benjamin0: If this stores a unique pointer we could `move` it in which would save the copy?

albatross/eigen/serializable_ldlt.h
```diff
@@ -0,0 +1,118 @@
...
+    Eigen::VectorXd sqrt_diag = this->vectorD();
+    for (Eigen::Index i = 0; i < n; i++) {
+      sqrt_diag[i] = 1. / std::sqrt(sqrt_diag[i]);
+    }
+
+    Eigen::VectorXd inv_diag(n);
+    for (Eigen::Index i = 0; i < n; i++) {
+      const Eigen::VectorXd col_i =
+          inverse_cholesky.col(i).cwiseProduct(sqrt_diag);
+      ;
```
benjamin0: nit: extra semicolon

albatross/eigen/serializable_ldlt.h
```diff
@@ -0,0 +1,118 @@
...
+   * decomposition represents in O(n^2) operations.
+   */
+  Eigen::VectorXd inverse_diagonal() const {
+    Eigen::Index n = this->rows();
+    Eigen::MatrixXd inverse_cholesky =
+        this->transpositionsP() * Eigen::MatrixXd::Identity(n, n);
+    this->matrixL().solveInPlace(inverse_cholesky);
+
+    Eigen::VectorXd sqrt_diag = this->vectorD();
+    for (Eigen::Index i = 0; i < n; i++) {
```
benjamin0: nit: You could maybe use the `Eigen::Array` functions here.

albatross/cereal/eigen.h
akleeman: It was the same idea, but with explict use of `VectorXd`.  Now it's via template specialization with `_Cols=1`.

albatross/eigen/serializable_ldlt.h
akleeman: If I understand correct, the problem with that is that it would require changing the way `Eigen::LDLT` works.  As it stands all the storage is done in the base class.  We might be able to do something like instantiate an empty `Eigen::LDLT` then point the internal members to pointers that are passed in ... but that feels wrong.

albatross/eigen/serializable_ldlt.h
akleeman: Ah yeah interesting.  Despite the appearance it seems like `x.array().sqrt().matrix()` doesn't add any runtime overhead ?!?

albatross/eigen/serializable_ldlt.h
akleeman: done

albatross/eigen/serializable_ldlt.h
benjamin0: Yeah probably not worth it.

albatross/cereal/eigen.h
benjamin0: good point, I'm fine with whichever you think is easiest.

akleeman: alright, I changed it around so the rows and cols are stored explicitly.  Shouldn't add much overhead in storage and makes the code easier to read (I think).

PR #26, opened 2018-06-12
Temperature Example

akleeman:
Adds a temperature example along documentation andh the support scripts to download, prepare and reproduce the plots used in the documentation.

The resulting docs can be found here: https://swiftnav-albatross.readthedocs.io/en/akleeman-temperature-example/index.html


examples/sinc_example_utils.h
```diff
@@ -0,0 +1,175 @@
...
+  for (int i = 0; i < n; i++) {
+    xs.push_back(distribution(generator));
+  }
+  return xs;
+};
+
+/*
+ * Creates a grid of n points between low and high.
+ */
+std::vector<double> uniform_points_on_line(const int n, const double low,
```
benjamin0: I think Eigen has some `linspace` utils if they're useful here.

akleeman: ohh yeah nice, hadn't noticed those.

PR #27, opened 2018-06-13
Spherical Covariance Tests.

akleeman:
Added several tests for positive definite covariance functions on spheres.

Consolidated some of the Distance Metrics

Add testing of doc generation.


akleeman:
for some reason travis isn't running the tests, but I verified that commit 49e2ab8 builds here: https://travis-ci.org/akleeman/albatross/builds/392506464

PR #30, opened 2018-07-02
Add ability to serialize Distributions and RegressionDatasets.

akleeman:
This gives you the ability to serialize a `RegressionDataset` provided the `FeatureType` is also serializable.  This can come in handy when it's costly to parse data and convert it into the required `features` and `targets`.  For example, you can now you can do something like:

```
// Parse data and cache it.
std::vector<RegressionDataset<MyFeature>> datasets = parse_lots_of_data();
{
  std::ofstream ofs(cache_path);
  cereal::BinaryOutputArchive archive(ofs);
  archive(datasets);
}
```
Allowing you to read them in much faster next time,
```
// Read the cached datasets.
std::vector<RegressionDataset<MyFeature>> datasets;
{
  std::ifstream ifs(cache_path);
  cereal::BinaryInputArchive archive(ofs);
  archive(datasets);
}
```

### DIFF
--- a/albatross/core/serialize.h
+++ b/albatross/core/serialize.h
@@ -13,6 +13,7 @@
 #ifndef ALBATROSS_CORE_SERIALIZE_H
 #define ALBATROSS_CORE_SERIALIZE_H
 
+#include "core/traits.h"
 #include <cereal/archives/json.hpp>
 #include <cereal/types/polymorphic.hpp>
 #include <iostream>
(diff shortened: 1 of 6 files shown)


albatross/core/serialize.h
```diff
@@ -38,21 +39,52 @@ class SerializableRegressionModel : public RegressionModel<FeatureType> {
...
 
-  // todo: enable if ModelFit is serializable.
-  template <class Archive> void save(Archive &archive) const {
+  /*
+   * Include save/load methods conditional on the ability to serialize
+   * ModelFit.
+   */
+  template <class Archive>
+  typename std::enable_if<valid_output_serializer<ModelFit, Archive>::value,
+                          void>::type
```
kmdade: this hurts me :( I feel like we should extend the character limit on our formatter.

albatross/eigen/serializable_diagonal_matrix.h
```diff
@@ -0,0 +1,58 @@
...
+
+template <typename _Scalar, int SizeAtCompileTime>
+class SerializableDiagonalMatrix
+    : public Eigen::DiagonalMatrix<_Scalar, SizeAtCompileTime> {
+  using BaseClass = Eigen::DiagonalMatrix<_Scalar, SizeAtCompileTime>;
+
+public:
+  SerializableDiagonalMatrix() : BaseClass(){};
+
+  SerializableDiagonalMatrix(const BaseClass &other)
```
kmdade: you implicitly get `SerializableDiagonalMatrix(const BaseClass &&other)` I think, so that's the only scenario where you don't actually need to copy.

albatross/core/serialize.h
akleeman: yeah kinda annoying.  I guess it's already a pretty difficult line to read, so doesn't make it that much worse :)

PR #32, opened 2018-07-03
If param key not found, print list of existing param keys.

benjamin0:


albatross/core/parameter_handling_mixin.h
```diff
@@ -60,19 +60,27 @@ class ParameterHandlingMixin {
 
   virtual ~ParameterHandlingMixin(){};
 
+  void check_param_key(const ParameterKey &key) {
+    const ParameterStore current_params = get_params();
+    if (!map_contains(current_params, key)) {
+      std::cerr << "Error: Key `" << key << "` not found in parameters: "
+                << pretty_params(current_params);
+      exit(EXIT_FAILURE);
```
akleeman: How does this differ from a failed `assert`?

akleeman: cool thanks.

PR #39, opened 2018-07-24
Allow fixed parameters and direct NLL.

akleeman:
Allow for parameters which are fixed (so will not be tuned.)

Add a negative log likelihood for Gaussian processes.

### DIFF
--- a/albatross/core/priors.h
+++ b/albatross/core/priors.h
@@ -37,6 +37,7 @@ public:
   virtual std::string get_name() const = 0;
   virtual double lower_bound() const { return -LARGE_VAL; }
   virtual double upper_bound() const { return LARGE_VAL; }
+  virtual bool is_fixed() const { return false; }
   virtual bool operator==(const Prior &other) const {
     return typeid(*this) == typeid(other);
   }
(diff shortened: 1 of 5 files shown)


albatross/core/priors.h
```diff
@@ -166,10 +191,14 @@ class LogNormalPrior : public Prior {
 
 } // namespace albatross
 
-CEREAL_REGISTER_TYPE(albatross::UninformativePrior);
-CEREAL_REGISTER_TYPE(albatross::PositivePrior);
-CEREAL_REGISTER_TYPE(albatross::UniformPrior);
-CEREAL_REGISTER_TYPE(albatross::GaussianPrior);
-CEREAL_REGISTER_TYPE(albatross::LogNormalPrior);
+CEREAL_REGISTER_TYPE_WITH_NAME(albatross::UninformativePrior,
```
akleeman: @benjamin0 note that this change should make the serialized JSON files a *little* easier to read, but at the price of non-backward compatible serialization.  Let me know if you think it'll be too much of a pain to recreate the serialized models we're currently using.

PR #40, opened 2018-07-25
Pretty Printing

akleeman:
Fix some formatting and naming issues surrounding printing priors for debugging.

Now `pretty_param_details` looks like this:

```
    sigma_constant                   value: 100        prior: positive        bounds: [0, inf]
    sigma_independent_noise          value: 10         prior: positive        bounds: [0, inf]
    sigma_slope                      value: 100        prior: none            bounds: [-inf, inf]
    sigma_squared_exponential        value: 100        prior: positive        bounds: [0, inf]
    squared_exponential_length_scale value: 1.5        prior: positive        bounds: [0, inf]
```

### DIFF
--- a/albatross/core/parameter_handling_mixin.h
+++ b/albatross/core/parameter_handling_mixin.h
@@ -13,6 +13,7 @@
 #ifndef ALBATROSS_CORE_PARAMETER_HANDLING_MIXIN_H
 #define ALBATROSS_CORE_PARAMETER_HANDLING_MIXIN_H
 
+#include <algorithm>
 #include <assert.h>
 #include <iomanip>
 #include <map>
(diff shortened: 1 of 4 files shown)


albatross/core/parameter_handling_mixin.h
```diff
@@ -115,15 +116,22 @@ inline std::string pretty_priors(const ParameterStore &params) {
 
 inline std::string pretty_param_details(const ParameterStore &params) {
   std::ostringstream ss;
+
+  std::size_t max_name_length = 0;
```
benjamin0: nit: Could you use https://en.cppreference.com/w/cpp/algorithm/max_element here?

akleeman: I think I'd have to create a vector of sizes first:
```
std::vector<std::size_t> sizes;
for (const auto &pair : params) {
    sizes.push_back(pair.first.size());
}
max_name_length = std::max_element(sizes.begin(), sizes.end());
```
or is there an easier way?

benjamin0: I think you can pass in a comparator as a 3rd argument

PR #41, opened 2018-07-26
Speedup LOO

akleeman:
Remove unnecessarily recomputed information vector step from the fast leave one out computations.


benjamin0:
👍

PR #45, opened 2018-08-03
Cross Validation Refactor

akleeman:
WIP


albatross/core/dataset.h
```diff
@@ -0,0 +1,91 @@
...
+ * target.
+ */
+template <typename FeatureType> struct RegressionDataset {
+  std::vector<FeatureType> features;
+  MarginalDistribution targets;
+  std::map<std::string, std::string> metadata;
+
+  RegressionDataset(){};
+
+  RegressionDataset(const std::vector<FeatureType> &features_,
```
kmdade: the trailing underscore convention is backwards here from what we normally have. I'm going to assume its consistent with the rest of albatross however which is all that really matters. I think we talked about this ages ago... :)

albatross/core/indexing.h
```diff
@@ -47,6 +54,21 @@ inline Eigen::VectorXd subset(const std::vector<SizeType> &indices,
   return out;
 }
 
+/*
+ * Extracts a subset of columns from an Eigen::Matrix
+ */
+template <typename SizeType>
```
kmdade: Does this have to be templated? can it not just be hardcoded to use `size_t` or `unsigned` or whatever?

akleeman: Yeah this might be overuse of templating, in general the value type of `col_indices` can be anything that can be converted to an `Eigen::Index`, but now I think the only thing that is really used is `s32`.  Whatever the decision is the same argument would apply to all the rest of the `subset` functions in this file as well.

albatross/core/dataset.h
akleeman: yeah, the rest of albatross uses a trailing underscore to avoid shadowing.  I started to try to change them all ... then decided it was more work than it was worth.

PR #46, opened 2018-08-08
GP Cross Validation

akleeman:
Builds on the cross validation API from #45 by moving the fast Gaussian process cross validation code into specialized `RegressionModel.cross_validated_predict_` methods.

Includes the ability to do arbitrary leave one group (previously it was only leave one out).

### DIFF
--- a/albatross/core/model_adapter.h
+++ b/albatross/core/model_adapter.h
@@ -117,6 +117,38 @@ public:
     return sub_model_.get_fit();
   }
 
+  /*
+   * Cross validation specializations
+   */
+  virtual std::vector<JointDistribution> cross_validated_predictions_(
+      const RegressionDataset<FeatureType> &dataset,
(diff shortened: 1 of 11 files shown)


albatross/core/model_adapter.h
```diff
@@ -117,6 +117,38 @@ class AdaptedRegressionModel
     return sub_model_.get_fit();
   }
 
+  /*
+   * Cross validation specializations
+   */
+  virtual std::vector<JointDistribution> cross_validated_predictions_(
```
kmdade: Maybe leave an informational comment lying around somewhere explaining about why these methods have the trailing underscore -- name resolution / ADL rules yada yada yada. Comments like this don't really run the risk of turning "stale" if they are purely informational and not necessarly specific to the code at hand.

akleeman: yeah good call.

PR #49, opened 2018-08-20
Specialize `adjust_storage_size()` so it works correctly for cereal::JSONInputArchive.

benjamin0:
I needed to revert this line for the `cereal::BinaryInputArchive()` to work. Do we still need it for the JSON archive?


akleeman:
I would guess that we do need it still ... though if the tests pass I *think* you're good to go.

akleeman:
Awesome looks good.  Merge away.

If you want to be sure everything will always serialize to binary you could add another test to `test_serialize` along the lines of `TYPED_TEST(PolymorphicSerializeTest, test_roundtrip_serialize)` that roundtrips via Binary.

benjamin0:
closing in favor of https://github.com/swift-nav/albatross/pull/50

PR #59, opened 2018-10-26
Add Metadata to Models

akleeman:
Add `metadata_` to all RegressionModels which provides a way of returning arbitrary informative information about how a model was fit to the user.

### DIFF
--- a/albatross/core/model.h
+++ b/albatross/core/model.h
@@ -35,6 +35,8 @@ template <typename T> struct PredictTypeIdentity { typedef T type; };
 // predict variants for which you only want the mean.
 using PredictMeanOnly = Eigen::VectorXd;
 
+using Insights = std::map<std::string, std::string>;
+
 /*
  * A model that uses a single Feature to estimate the value of a double typed
  * target.
@@ -68,7 +70,7 @@ public:
     assert(features.size() > 0);
     assert(features.size() == static_cast<std::size_t>(targets.size()));
     has_been_fit_ = true;
-    metadata_["input_feature_count"] = std::to_string(features.size());
+    insights_["input_feature_count"] = std::to_string(features.size());
     fit_(features, targets);
   }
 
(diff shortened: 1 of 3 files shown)


albatross/models/ransac.h
```diff
@@ -229,6 +229,8 @@ class GenericRansac : public RegressionModel<FeatureType> {
     RegressionDataset<FeatureType> inliers =
         ransac(dataset, fold_indexer, sub_model_, metric_, inlier_threshold_,
                random_sample_size_, min_inliers_, max_iterations_);
+    this->metadata_["post_ransac_feature_count"] =
```
benjamin0: And `input_feature_count` is the `pre_ransac_feature_count`?

albatross/core/model.h
```diff
@@ -256,6 +259,7 @@ class RegressionModel : public ParameterHandlingMixin {
   }
 
   bool has_been_fit_;
+  std::map<std::string, std::string> metadata_;
```
benjamin0: nit: Should we call this `insights` to make it consistent and clear these are results from the model fit, not metadata of the inputs?

albatross/core/model.h
```diff
@@ -137,6 +138,8 @@ class RegressionModel : public ParameterHandlingMixin {
 
   virtual std::string get_name() const = 0;
 
+  std::map<std::string, std::string> get_metadata() const { return metadata_; }
```
benjamin0: nit: define this with a `using`?

albatross/models/ransac.h
akleeman: yeah, I just didn't want to call it that since ransac isn't necessarily run every time.

albatross/core/model.h
akleeman: yeah good call, I like that name better.

albatross/core/model.h
akleeman: done

PR #60, opened 2018-10-29
Switch to CRTP for Covariance Terms.

akleeman:
The existing implementation of `CovarianceFunction` and `CovarianceTerm` is unnecessary, a `CovarianceTerm` is only ever used if wrapped in a `CovarianceFunction`.  This PR removes the `CovarianceTerm` in favor of a fully CRTP `CovarianceFunction`.  Basically, instead of needing to do:

    CovarianceFunction<Constant> constant = {Constant()}

you can just do:

    Constant constant;

One unfortunate thing about flattening what was the `CovarianceTerm` into the `CovarianceFunction` via CRTP is that it's easy to get confused between the `operator()` methods on a derived class and the one inherited through CRTP.  That along with the fact that [overloaded functions get hidden via inheritance](https://stackoverflow.com/questions/1628768/why-does-an-overridden-function-in-the-derived-class-hide-other-overloads-of-the/1629074#1629074) makes it desirable to have a completely different name for any implementation of the call operator.  In turn, I've changed it so any implementation of a covariance function needs to add `call_impl_` methods.  This means the actual user facing calls remain unchanged.  Ie, following the examples above, `constant(x, y)` would still compute the covariance between `x` and `y`, while if you look at `Constant` you'll see `double call_impl_(X &x, Y &y)` instead of `double operator() (X &x, Y&y)`.

In summary defining and using covariance term used to be done like this:

```
class Foo : public CovarianceTerm {
    double operator() (X &x, Y&y) {...};
    string get_name() {return "foo";};
}
CovarianceFunction<Foo> foo = {Foo()};
foo(x, y)
```
and with this PR would be done,
```
class Foo : public CovarianceFunction<Foo> {
    double call_impl_() (X &x, Y&y) {...};
    name_ = "foo";
}
Foo foo;
foo(x, y);
```


albatross/covariance_functions/polynomials.h
```diff
@@ -28,17 +28,15 @@ struct ConstantTerm {};
  * underlying constant term would actually be a biased estimate
  * of the mean.
  */
-class Constant : public CovarianceTerm {
+class Constant : public CovarianceFunction<Constant> {
 public:
-  Constant(double sigma_constant = 10.) {
+  Constant(double sigma_constant = 10.) : name_("constant") {
```
kmdade: NIT suggestions: This and the initializer constant in the diagonal covariance class could be a named `#define` or constexpr just so that the numbers don't get lost along the way, and changing them at the top of the file is plain to see. I guess nobody should ever be using these classes without an argument anyway so it probably doesn't matter.

akleeman:
@kmdade Yeah I'm also 100% sure there are edge cases ... but feel pretty confident that everything behaves the same way after these changes as it did before.  We were already using a CRTP-like approach with the covariance terms and functions so with the exception of a few extra trait checks I don't think we're doing any additional template work that we weren't already doing.  If the test run times are any indication those haven't changed with this PR (despite adding more tests).

albatross/covariance_functions/polynomials.h
akleeman: Yeah, I'm not super satisfied with the `name_` member approach.  you have any opinions on an approach for `constexpr` strings along these lines: https://akrzemi1.wordpress.com/2017/06/28/compile-time-string-concatenation/

PR #64, opened 2018-11-10
Adapated Ransac Bug

akleeman:
Using ransac with the adapted model had an awful bug in which some of the input parameters were passed by reference but could then lose scope.

This is basically a two character change (ie switching a lambda function to a copy capture and removing a reference operator in a member declaration).  The rest are just tests which don't pass without those changes.

### DIFF
--- a/albatross/core/model_adapter.h
+++ b/albatross/core/model_adapter.h
@@ -141,7 +141,7 @@ public:
     GenericModelFunctions<FeatureType, FitType> funcs;
 
     decltype(funcs.fitter) fitter =
-        [&](const std::vector<FeatureType> &features,
+        [=](const std::vector<FeatureType> &features,
             const MarginalDistribution &targets) {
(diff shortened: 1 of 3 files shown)


albatross/core/model_adapter.h
```diff
@@ -141,7 +141,7 @@ class AdaptedRegressionModel
     GenericModelFunctions<FeatureType, FitType> funcs;
 
     decltype(funcs.fitter) fitter =
-        [&](const std::vector<FeatureType> &features,
+        [=](const std::vector<FeatureType> &features,
```
kmdade: Yes good catch.

kmdade: I would also propose that if you only intend to capture one or two variables, it could even be better to name them explicitly. Doesn't really make a difference, but it can communicate intent sometimes.

akleeman: Good call, I've updated it to be explicit.

PR #68, opened 2018-11-19
Add utilities for concatenating datasets and distributions.

akleeman:


albatross/core/concatenate.h
```diff
@@ -0,0 +1,86 @@
...
+    assert(target_has_covariance == d.targets.has_covariance());
+    for (std::size_t j = 0; j < d.features.size(); ++j) {
+      features[i] = d.features[j];
+      Eigen::Index ei = static_cast<Eigen::Index>(i);
+      Eigen::Index ej = static_cast<Eigen::Index>(j);
+      target_means[ei] = d.targets.mean[ej];
+      target_variance[ei] = d.targets.get_diagonal(ej);
+      ++i;
+    }
+    metadata = map_join(metadata, d.metadata);
```
seth-swiftnav: What happens if two metadatas have the same field, it keeps the latter?
There's no assurance that the metadata for a dataset will apply to the concatenation. This doesn't interfere with anything currently, but it may introduce user error down the line.

tests/test_eigen_utils.cc
```diff
@@ -45,3 +45,31 @@ TEST(test_eigen_utils, test_diag_block_diag) {
...
+
+  Eigen::VectorXd B(9);
+  B << 7, 8, 9, 10, 11, 12, 13, 14, 15;
+
+  std::vector<Eigen::VectorXd> input = {A, B};
+  const auto C = albatross::vertical_stack(input);
+  Eigen::VectorXd expected(15);
+  expected << 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15;
+  EXPECT_EQ(expected, C);
+}
```
seth-swiftnav: Is there are way to test that a function failed? Passing in matrices of different widths?

albatross/core/concatenate.h
akleeman: Yeah exactly, currently if all the metadata is not the same it'll overwrite with the last.  I think if we're explicit about that in a comment then it's OK.  My current philosophy on metadata is that it's around as a helper, but shouldn't really be used or relied on programmatically.  We had a similar approach when designing the merge algorithm for attributes in [`xarray.Dataset`](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.merge.html).  Similar to how that ended up we could eventually provide a strategy enum for concatenation to make the user need to explicitly say something like:
```
enum {
    INTERSECTION_OF_METADATA,
    UNION_OF_METADATA,
    IGNORE_METADATA
} concatenation_strategy_t

...

concatenate_datasets(datasets, INTERSECTION_OF_METADATA)
```
That said, If we're worried about confusion I'm fine just having it drop the metadata for now.

tests/test_eigen_utils.cc
akleeman: There are ways to [catch exceptions](https://stackoverflow.com/questions/23270078/test-a-specific-exception-type-is-thrown-and-the-exception-has-the-right-propert), though I think `assert` behaves different from an exception.  We could probably redefine the [`assert` macro](https://en.cppreference.com/w/cpp/error/assert)?  But that all seems a bit beyond the amount of effort I want to put into this PR, particularly since I'd say that if we're going to start unit testing that asserts fail (which isn't a bad idea) then that'll become a pretty sweeping addition.

PR #70, opened 2018-11-30
Consolidate Forward Declarations

akleeman:
Adds a bunch of the most common classes used in `albatross` into `declarations.h` which makes it easier to isolate the template heavy parts of albatross.


seth-swiftnav:
Looks good. Makes recompilation happy and I can't think of any other big types right now.

PR #73, opened 2019-01-09
Removes assertion when Ransac fails consensus. Sets has_been_fit() to…

seth-swiftnav:
… false. Postpones crash until predict step.

### DIFF
--- a/albatross/models/gp.h
+++ b/albatross/models/gp.h
@@ -156,7 +156,7 @@ public:
 
   void set_fit(const FitType &fit) {
     this->model_fit_ = fit;
-    this->has_been_fit_ = true;
+    this->has_been_fit_ = (fit.train_features.size() > 0);
   }
 
   /*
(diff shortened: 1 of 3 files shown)


akleeman:
OK, so with this change if you pass in a dataset for which a consensus is never found then instead of a hard fail when calling fit, you'd get a hard fail calling `predict` which gives us the opportunity to check `has_been_fit()` after calling `fit()` to determine if ransac succeeded or not.  Do I have that right?

seth-swiftnav:
> OK, so with this change if you pass in a dataset for which a consensus is never found then instead of a hard fail when calling fit, you'd get a hard fail calling `predict` which gives us the opportunity to check `has_been_fit()` after calling `fit()` to determine if ransac succeeded or not. Do I have that right?
The crash is postponed until calling predict. However the behavior of has_been_fit() is not intuitive here. It will still return `true` if called on the Ransac model. However it will now return `false` if called on the submodel, or model on which the Ransac was built. My attempts at making the desired behavior crashed simpler models such as Least Squares.

akleeman:
Seems reasonable to me.  I think the entire framework needs to be rethought a little at which point we can come up with a better solution.

PR #78, opened 2019-02-08
Fix Matrix Covariance Function Methods.

akleeman:
Fixes a bug in which evaluating a covariance function using a vector of valid types would land on the deleted call operator instead of the desired one.  This was due to a missing `const` in the deleted method.

The bug fix is the addition of `const` on line 224 of `covariance_function.h`. 

Other Changes Include:
- An additional set of operators which map `covariance_function(x)` to `covariance_function(x, x)` and the corresponding deleted function to help debugging.
- Test to prevent a regression

### DIFF
--- a/albatross/covariance_functions/covariance_function.h
+++ b/albatross/covariance_functions/covariance_function.h
@@ -142,6 +142,16 @@ public:
   template <typename X,
             typename std::enable_if<
                 has_defined_call_impl<Derived, X &, X &>::value, int>::type = 0>
+  double operator()(const X &x) const {
+    return derived().call_impl_(x, x);
+  }
+
+  /*
+   * Covariance between each element and every other in a vector.
+   */
+  template <typename X,
+            typename std::enable_if<
+                has_defined_call_impl<Derived, X &, X &>::value, int>::type = 0>
   Eigen::MatrixXd operator()(const std::vector<X> &xs) const {
     int n = static_cast<int>(xs.size());
     Eigen::MatrixXd C(n, n);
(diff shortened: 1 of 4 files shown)


albatross/covariance_functions/covariance_function.h
```diff
@@ -136,6 +136,16 @@ class CovarianceFunction : public ParameterHandlingMixin {
     return derived().call_impl_(y, x);
   }
 
+  /*
+   * Covariance between each element and every other in a vector.
+   */
+  template <typename X,
+            typename std::enable_if<
+                has_defined_call_impl<Derived, X &, X &>::value, int>::type = 0>
+  double operator()(const X &x) const {
```
kmdade: doesn't `Derived` need to have a `defined_call_impl` with `const X &` here?

jbangelo: I would think so.
I also think there's a syntax that would allow not repeating the type which would prevent this potential issue, but it'd be kinda weird looking. Something like this:
```
template<typename X>
auto operator()(const X &x) const -> std::enable_if<has_defined_call_impl<Derived, decltype(x), decltype(x)>::value, double>::type;
```

akleeman: Yeah good point.  I do think we want all arguments to be const references.

@jbangelo I think "kinda weird looking" is a good way to describe albatross.  One alternative (or maybe supplement) could be to change `has_defined_call_impl<X, T>` to only check for `X::call_impl_(const T &)`.

PR #79, opened 2019-02-09
Improve Traits

akleeman:
These changes are intended to give more valuable information when a `CovarianceFunction` is ill-formed.

In particular the addition of `has_possible_call_impl` in conjuction with `has_valid_call_impl` (formerly `has_defined_call_impl`) are used to create compiler errors when a CRTP `CovarianceFunction` is close to, but not quite, properly defined.

For example if we start with
```
class IllFormed : public CovarianceFunction<IllFormed> {
public:
  // valid
  double call_impl_(const X &, const X &) const { return 0.; };
  // invalid return type
  int call_impl_(const X &, const Y &) const { return 0; };
  // invalid arguments
  double call_impl_(Y &, const Y &) const { return 0.; };
};
```
Then use some test types,
```
  IllFormed cov;
  X x = {};
  Y y = {};
  Z z = {};
```
If we were to write `cov(x, x)` it will compile fine but if we use an invalid call such as, `cov(x, y)` we get:
```
error: use of deleted function ‘double albatross::CovarianceFunction<Derived>::operator()(const X&, const Y&) const [with X = albatross::X; Y = albatross::Y; typename std::enable_if<(((! albatross::has_valid_call_impl<Derived, X&, Y&>::value) && (! albatross::has_valid_call_impl<Derived, Y&, X&>::value)) && (albatross::has_invalid_call_impl<Derived, X&, Y&>::value || albatross::has_invalid_call_impl<Derived, Y&, X&>::value)), int>::type <anonymous> = 0; Derived = albatross::IllFormed]’
   cov(x, y);
           ^
In file included from ./albatross/covariance_functions/covariance_function.h:262:10: note: declared here
   double operator()(const X &, const Y &) const = delete; // Invalid call_impl_
```
as before if we try with a completely undefined function such as `cov(z, z)` we get a similar (but slightly different) error:
```
error: use of deleted function ‘double albatross::CovarianceFunction<Derived>::operator()(const X&, const Y&) const [with X = albatross::Z; Y = albatross::Z; typename std::enable_if<(((! albatross::has_valid_call_impl<Derived, X&, Y&>::value) && (! albatross::has_valid_call_impl<Derived, Y&, X&>::value)) && ((! albatross::has_possible_call_impl<Derived, X&, Y&>::value) && (! albatross::has_possible_call_impl<Derived, Y&, X&>::value))), int>::type <anonymous> = 0; Derived = albatross::IllFormed]’
   cov(z, z);
           ^
In file included from ./albatross/covariance_functions/covariance_function.h:250:10: note: declared here
   double operator()(const X &, const Y &) const = delete; // No call_impl_
```

### DIFF
--- a/albatross/core/traits.h
+++ b/albatross/core/traits.h
@@ -43,7 +43,7 @@ template <class T> struct call_impl_arg_type {
 
 /*
  * This determines whether or not a class has a method defined for,
- *   `operator() (X x, Y y, Z z, ...)`
+ *   `operator() (const X &x, const Y &y, const Z &z, ...)`
  * The result of the inspection gets stored in the member `value`.
  */
 template <typename T, typename... Args> class has_call_operator {
@@ -66,10 +66,11 @@ public:
 template <typename T, typename... Args> class has_valid_call_impl {
 
   template <typename C>
-  static auto test(C *) -> typename std::is_same<
+  static typename std::is_same<
       decltype(std::declval<const C>().call_impl_(
           std::declval<typename call_impl_arg_type<Args>::type>()...)),
-      double>::type;
+      double>::type
+  test(C *);
   template <typename> static std::false_type test(...);
 
 public:
--- a/albatross/core/traits.h
+++ b/albatross/core/traits.h
@@ -48,12 +48,11 @@ template <class T> struct call_impl_arg_type {
  */
 template <typename T, typename... Args> class has_call_operator {
 
-  template <typename C>
-  static constexpr auto test(C *) -> typename std::is_same<
-      decltype(std::declval<C>()(
-          std::declval<typename call_impl_arg_type<Args>::type>()...)),
-      double>::type;
-  template <typename> static constexpr std::false_type test(...);
+  template <typename C,
+            typename = decltype(std::declval<C>()(
+                std::declval<typename call_impl_arg_type<Args>::type>()...))>
+  static std::true_type test(C *);
+  template <typename> static std::false_type test(...);
 
 public:
   static constexpr bool value = decltype(test<T>(0))::value;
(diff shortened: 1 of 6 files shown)


albatross/core/traits.h
```diff
@@ -28,36 +28,82 @@ namespace albatross {
...
+ */
+template <class T> struct call_impl_arg_type {
+  typedef
+      typename std::add_lvalue_reference<typename std::add_const<T>::type>::type
+          type;
+};
+
 /*
  * This determines whether or not a class has a method defined for,
  *   `operator() (X x, Y y, Z z, ...)`
```
jbangelo: You updated other comments with the new const reference types, shouldn't these be updated as well?

albatross/core/traits.h
```diff
@@ -28,36 +28,82 @@ namespace albatross {
...
  * This determines whether or not a class has a method defined for,
- *   `call_impl_(X x, Y y, Z z, ...)`
+ *   `double call_impl_(const X &x, const Y &y, const Z &z, ...)`
  * The result of the inspection gets stored in the member `value`.
  */
-template <typename T, typename... Args> class has_defined_call_impl {
+template <typename T, typename... Args> class has_valid_call_impl {
+
+  template <typename C>
+  static auto test(C *) -> typename std::is_same<
```
jbangelo: Any particular reason you changed how you are defining the `test()` function trait thing? Everywhere else you have your SFINAE switch in the template and here it's a part of the return type declaration.

albatross/core/traits.h
```diff
@@ -28,36 +28,82 @@ namespace albatross {
...
+ * if a class has:
+ *   double call_impl_(const X x)
+ * or
+ *   double call_impl_(X &x)
+ * or
+ *   int call_impl_(const X &x)
+ * those are nearly correct but the required `const X &x` in which
+ * case this trait can be used to warn the user.
+ */
+template <typename T, typename... Args> class has_possible_call_impl {
```
jbangelo: So this only checks if the `call_impl_()` takes in references as arguments? Will is do anything if there's not return type? Is it an issue if it doesn't?

albatross/core/traits.h
akleeman: good catch

albatross/core/traits.h
akleeman: This checks if it takes anything that *could* be a reference (so it'll return true for `call_impl_(X x)` or `call_impl_(X &x)` ...).  I don't think it'll deal with `call_impl_(X &&x)` but I'm ok with that at the moment.

In terms of return type, you're right it doesn't check the return type on purpose.  Since it's used as a way to generate more useful compiler errors I want it to catch some of the more common bugs, things like `int call_impl_(...)` which won't pass the `has_valid_call_impl` test in order to point the user towards the culprit function definition quicker.

albatross/core/traits.h
akleeman: No great reason, I'll swtich back for consistency.

PR #83, opened 2019-03-05
CRTP: ModelBase

akleeman:
This change replaces the `RegressionModel<FeatureType` virtual inheritance approach with a CRTP style `ModelBase<ModelType>` approach.

This is absolutely not going to pass tests and may contain declarations to parts of the implementation that don't exist yet.  But should give a representative idea of the changes.

My general philosophy on this process is that I'd like to make a first pass through this refactor which will let us successfully reproduce all the existing behavior in `albatross`.  I'd then like to (re)integrate it with our projects and then loop back and put some finishing touches and/or extend functionality in a second pass.  I'll break the first pass into smallish commits (like this one) in which I'm primarily hoping to avoid some design decisions that could lead to dead ends or add tech debt without focusing so much on edge cases.

A top level summary is this.  With the existing version of `albatross` models could be used along these lines:
```
void print_mean_prediction(RegressionModel<FeatureType> &model,
                                             RegressionDataset<FeatureType> &dataset) {
    model.fit(dataset);
    const auto pred = model.template predict<MarginalDistribution>(dataset.features);
    std::cout << pred.mean.norm() << std::endl;
}
```
You would now do:
```
template <typename FeatureType, typename ModelType>
void print_mean_prediction(ModelBase<ModelType> &model,
                                             RegressionDataset<FeatureType> &dataset) {
    const auto fit_model = model.get_fit_model(dataset);
    const auto pred = get_prediction(dataset.features).marginal();
    std::cout << pred.mean.norm() << std::endl;
}
```
I would recommend glancing at `ModelBase` first to get a sense of how that's wired together, then looking at how it was used in `LeastSquares` and subsequently extended in `LinearRegression`.

### DIFF
--- a/albatross/core/fit_model.h
+++ b/albatross/core/fit_model.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2018 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef ALBATROSS_CORE_FIT_H
+#define ALBATROSS_CORE_FIT_H
+
+namespace albatross {
+
+template <typename ModelType, typename Fit>
+class FitModel {
+
+  template <typename X, typename Y, typename Z>
+  friend class Prediction;
+
+ public:
+
+  static_assert(std::is_move_constructible<Fit>::value,
+                "Fit type must be move constructible to avoid unexpected copying.");
+
+  FitModel(const ModelType &model,
+           const Fit &&fit)
+      : model_(model), fit_(std::move(fit)) {}
+
+ public:
+
+  template <typename PredictFeatureType>
+  Prediction<ModelType, PredictFeatureType, Fit>
+  get_prediction(const std::vector<PredictFeatureType> &features) const {
+    return Prediction<ModelType, PredictFeatureType, Fit>(*this, features);
+  }
+
+  const ModelType &model_;
+  const Fit fit_;
+
+};
+
+}
+#endif
--- a/albatross/core/fit_model.h
+++ b/albatross/core/fit_model.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Swift Navigation Inc.
+ * Copyright (C) 2019 Swift Navigation Inc.
  * Contact: Swift Navigation <dev@swiftnav.com>
  *
  * This source is subject to the license found in the file 'LICENSE' which must
@@ -10,27 +10,23 @@
  * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
  */
 
-#ifndef ALBATROSS_CORE_FIT_H
-#define ALBATROSS_CORE_FIT_H
+#ifndef ALBATROSS_CORE_FIT_MODEL_H
+#define ALBATROSS_CORE_FIT_MODEL_H
 
 namespace albatross {
 
 template <typename ModelType, typename Fit>
 class FitModel {
-
+ public:
   template <typename X, typename Y, typename Z>
   friend class Prediction;
-
- public:
 
   static_assert(std::is_move_constructible<Fit>::value,
                 "Fit type must be move constructible to avoid unexpected copying.");
 
   FitModel(const ModelType &model,
-           const Fit &&fit)
+           Fit &&fit)
       : model_(model), fit_(std::move(fit)) {}
-
- public:
 
   template <typename PredictFeatureType>
   Prediction<ModelType, PredictFeatureType, Fit>
@@ -38,7 +34,8 @@ class FitModel {
     return Prediction<ModelType, PredictFeatureType, Fit>(*this, features);
   }
 
-  const ModelType &model_;
+ private:
+  const ModelType model_;
   const Fit fit_;
 
 };
--- a/albatross/core/model.h
+++ b/albatross/core/model.h
@@ -53,7 +53,7 @@ template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
   auto
   fit_(const std::vector<FeatureType> &features,
        const MarginalDistribution &targets) const {
-    const auto fit = derived().fit(features, targets);
+    auto fit = derived().fit(features, targets);
     return FitModel<ModelType, decltype(fit)>(derived(), std::move(fit));
   }
 
--- a/albatross/core/model.h
+++ b/albatross/core/model.h
@@ -13,266 +13,110 @@
 #ifndef ALBATROSS_CORE_MODEL_H
 #define ALBATROSS_CORE_MODEL_H
 
-#include "core/dataset.h"
-#include "core/indexing.h"
-#include "core/parameter_handling_mixin.h"
-#include "map_utils.h"
-#include "traits.h"
-#include <Eigen/Core>
-#include <cereal/archives/json.hpp>
-#include <map>
-#include <vector>
-
 namespace albatross {
 
-namespace detail {
-// This is effectively just a container that allows us to develop methods
-// which behave different conditional on the type of predictions desired.
-template <typename T> struct PredictTypeIdentity { typedef T type; };
-}
-
-// This can be used to make intentions more obvious when calling
-// predict variants for which you only want the mean.
-using PredictMeanOnly = Eigen::VectorXd;
 
 using Insights = std::map<std::string, std::string>;
 
-/*
- * A model that uses a single Feature to estimate the value of a double typed
- * target.
- */
-template <typename FeatureType>
-class RegressionModel : public ParameterHandlingMixin {
-public:
-  using Feature = FeatureType;
-  RegressionModel() : ParameterHandlingMixin(), has_been_fit_(){};
-  virtual ~RegressionModel(){};
+template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
 
-  virtual bool operator==(const RegressionModel<FeatureType> &other) const {
-    // If the fit method has been called it's possible that some unknown
-    // class members may have been modified.  As such, if a model has been
-    // fit we fail hard to avoid possibly unexpected behavior.  Any
-    // implementation that wants a functional equality operator after
-    // having been fit will need to override this one.
-    assert(!has_been_fit());
-    return (get_name() == other.get_name() &&
-            get_params() == other.get_params() &&
-            has_been_fit() == other.has_been_fit());
-  }
+  template <typename X, typename Y, typename Z>
+  friend class Prediction;
 
-  /*
-   * Provides a wrapper around the implementation `fit_` which performs
-   * simple size checks and makes sure the fit method is called before
-   * predict.
-   */
-  void fit(const std::vector<FeatureType> &features,
-           const MarginalDistribution &targets) {
-    assert(features.size() > 0);
-    assert(features.size() == static_cast<std::size_t>(targets.size()));
-    has_been_fit_ = true;
-    insights_["input_feature_count"] = std::to_string(features.size());
-    fit_(features, targets);
-  }
+  template <typename T, typename FeatureType>
+  friend class fit_model_type;
 
-  /*
-   * Convenience function which assumes zero target covariance.
-   */
-  void fit(const std::vector<FeatureType> &features,
-           const Eigen::VectorXd &targets) {
-    return fit(features, MarginalDistribution(targets));
-  }
+  template <typename T, typename FitModelType>
+  friend struct fit_type_from_fit_model_type;
 
-  /*
-   * Convenience function which unpacks a dataset into features and targets.
-   */
-  void fit(const RegressionDataset<FeatureType> &dataset) {
-    return fit(dataset.features, dataset.targets);
-  }
+  template <typename T, typename FeatureType>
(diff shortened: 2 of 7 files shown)


akleeman:
Some decisions I've made that might be worth questioning.

1. `ModelBase` defines methods like `fit` and `predict`, but (due to [name hiding in inheritance](https://stackoverflow.com/questions/1628768/why-does-an-overridden-function-in-the-derived-class-hide-other-overloads-of-the)) that means that when implementing a specialization of these methods for a certain model and `FeatureType` you need to implement methods like `fit_impl_`.  I'm not a huge fan of how that works out ... but can't think of a better option.
2. There's been some discussion about CRTP versus free functions.  One assumption I've made is that if we get CRTP working then we should have a decent inventory of trait inspection that would make switching to free functions relatively easy.

albatross/core/model.h
```diff
@@ -13,266 +13,110 @@
...
-    set_params(params);
-  }
+  template <
+      typename FeatureType,
+      typename std::enable_if<has_valid_fit<ModelType, FeatureType>::value,
+                              int>::type = 0>
+  auto
+  fit_(const std::vector<FeatureType> &features,
+       const MarginalDistribution &targets) const {
+    const auto fit = derived().fit(features, targets);
```
jbangelo: `fit` can't be `const` if you're going to be moving it into the `FitModel` constructor below.

albatross/core/fit_model.h
```diff
@@ -0,0 +1,47 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef ALBATROSS_CORE_FIT_H
+#define ALBATROSS_CORE_FIT_H
+
+namespace albatross {
+
+template <typename ModelType, typename Fit>
+class FitModel {
```
akleeman: This `FitModel` class is the return type of `ModelBase.get_fit_model(dataset)`.  It's basically just a container which keeps a model and a fit produced by that model together.  This is also used to enforce that you can only produce a prediction after having called fit.  Notice that `ModelBase` doesn't have a `get_predictions` method, you have to first get a fit, then make predictions from it.  Ie, `ModelBase.get_fit_model(dataset).get_predictions(features)`.

albatross/core/model.h
```diff
@@ -13,266 +13,110 @@
...
+    return FitModel<ModelType, decltype(fit)>(derived(), std::move(fit));
+  }
+
+  template <typename FeatureType,
+            typename std::enable_if<
+                    has_possible_fit<ModelType, FeatureType>::value &&
```
jbangelo: I'm wondering if you could just use `!has_valid_fit<ModelType, FeatureType>::value` here and only have the one deleted overload. Or is there an edge case that wouldn't be caught?

albatross/core/fit_model.h
```diff
@@ -0,0 +1,47 @@
...
+
+ public:
+
+  template <typename PredictFeatureType>
+  Prediction<ModelType, PredictFeatureType, Fit>
+  get_prediction(const std::vector<PredictFeatureType> &features) const {
+    return Prediction<ModelType, PredictFeatureType, Fit>(*this, features);
+  }
+
+  const ModelType &model_;
```
akleeman: For now I'm storing a reference to the model.  So far though, these models are pretty lightweight, I could see it being "OK" to copy them each time you call fit.  On the other hand the `Fit` could be pretty large so we for sure don't want to copy it, hence the strict enforcement of `is_move_constructible` above.

albatross/core/prediction.h
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright (C) 2018 Swift Navigation Inc.
```
jbangelo: 2019 :)

albatross/core/fit_model.h
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2018 Swift Navigation Inc.
```
jbangelo: 2019

albatross/core/model.h
```diff
@@ -13,266 +13,110 @@
...
-  }
+template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
 
-  std::string pretty_string() const {
-    std::ostringstream ss;
-    ss << get_name() << std::endl;
-    ss << ParameterHandlingMixin::pretty_string();
-    return ss.str();
-  }
+  template <typename X, typename Y, typename Z>
```
akleeman: These friend classes are required in order to allow things like `Prediction` to inspect a `ModelType` and decide which `predict_` methods are available.  This way we can keep the CRTP created `fit_` and `predict_` methods private to avoid confusion.

albatross/core/model.h
```diff
@@ -13,266 +13,110 @@
...
+    const auto fit = derived().fit(features, targets);
+    return FitModel<ModelType, decltype(fit)>(derived(), std::move(fit));
+  }
+
+  template <typename FeatureType,
+            typename std::enable_if<
+                    has_possible_fit<ModelType, FeatureType>::value &&
+                    !has_valid_fit<ModelType, FeatureType>::value,
+                int>::type = 0>
+  FitModel<ModelType, FeatureType>
```
akleeman: At this point, the only thing CRTP is really adding is the ability to inspect whether or not the user has properly defined `fit` and `predict` in their implementations.  Eventually I plan on adding more informative inspection to make sure `fit` and `predict` methods work together, etc ...

albatross/core/prediction.h
```diff
@@ -0,0 +1,144 @@
...
+
+public:
+  Prediction(const FitModel<ModelType, FitType> &fit_model,
+             const std::vector<FeatureType> &features)
+      : fit_model_(fit_model), features_(features) {}
+
+  /*
+   * MEAN
+   */
+  template <
```
akleeman: These methods define all the logic about things like "If a `predict` method exists for some `Distribution<T>` but not for the mean (`Eigen::VectorXd`) we can get the mean from `Distribution<T>.mean`.  Typically the complexity of producing a full `JointDistribution` is larger than a `MarginalDistribution` which is larger than the mean only so these are designed to define `mean`, `marginal` and `joint` using the least computationally complex method available.

albatross/models/least_squares.h
```diff
@@ -36,22 +30,22 @@ struct LeastSquaresFit {
...
  * first creating a design matrix, A, then solving least squares.  Ie,
  *
  *   min_x |y - Ax|_2^2
  *
  * The FeatureType in this case is a single row from the design matrix.
  */
-class LeastSquaresRegression
-    : public SerializableRegressionModel<Eigen::VectorXd, LeastSquaresFit> {
+template <typename ImplType>
+class LeastSquares : public ModelBase<LeastSquares<ImplType>> {
```
akleeman: The `ImplType` here is a way to extend `LeastSquares`.  In the example of `LinearRegression` it just provides `fit` and `predict` methods that work with `double` types instead of `Eigen::VectorXd` types, but could for example include things like routines for converting weigthed least squares problems into regular least squares etc ...

albatross/models/least_squares.h
```diff
@@ -62,36 +56,56 @@ class LeastSquaresRegression
...
+          features[i].dot(least_squares_fit.coefs);
     }
-    return mean;
+    return JointDistribution(mean);
   }
 
-  JointDistribution
-  predict_(const std::vector<Eigen::VectorXd> &features) const override {
-    return JointDistribution(predict_mean_(features));
+  template <typename FeatureType, typename FitType, typename PredictType,
```
akleeman: These heavily templated methods "forward" on any `predict` methods defined in `ImplType` to this base class.  This is required since `ModelBase<LeastSquares<ImplType>>` can't actually inspect `ImplType` directly.

albatross/core/prediction.h
```diff
@@ -0,0 +1,144 @@
...
+ */
+
+#ifndef ALBATROSS_CORE_PREDICTION_H
+#define ALBATROSS_CORE_PREDICTION_H
+
+namespace albatross {
+
+// This is effectively just a container that allows us to develop methods
+// which behave different conditional on the type of predictions desired.
+template <typename T> struct PredictTypeIdentity { typedef T type; };
```
jbangelo: So this is effectively used as a tag type? If that's the case you don't strictly need to be declaring it as a reference in the function signatures where it is used. Empty classes like this take the minimum amount of space (usually a byte) and it's the same cost to copy the object as it is to copy a reference to it.

albatross/models/least_squares.h
```diff
@@ -103,39 +117,60 @@ class LeastSquaresRegression
...
-using LinearRegressionBase =
-    AdaptedRegressionModel<double, LeastSquaresRegression>;
-
-class LinearRegression : public LinearRegressionBase {
+class LinearRegression : public LeastSquares<LinearRegression> {
 
 public:
-  LinearRegression(){};
-  std::string get_name() const override { return "linear_regression"; };
+  //  std::string get_name() const { return "linear_regression"; };
```
jbangelo: Does this commented line still need to stick around?

albatross/models/least_squares.h
```diff
@@ -103,39 +117,60 @@ class LeastSquaresRegression
...
-  template <class Archive> void save(Archive &archive) const {
-    archive(cereal::make_nvp("linear_regression",
-                             cereal::base_class<LinearRegressionBase>(this)));
-  }
-
-  template <class Archive> void load(Archive &archive) {
-    archive(cereal::make_nvp("linear_regression",
-                             cereal::base_class<LinearRegressionBase>(this)));
-  }
+  //  template <class Archive> void save(Archive &archive) const {
```
jbangelo: Does this commented code need to stick around?

albatross/core/fit_model.h
jbangelo: Like we discussed, this is fine as long as the object `model_` references remains valid for as long as you're using the `FitModel` objects. This is highly dependent on how Albatross is used, and is probably not something that easily enforced from inside of Albatross.

When we talked I, for some reason, was thinking that we didn't want to use any dynamic memory in Albatross. In hindsight I don't think that's strictly the case, it's not being used in an embedded context. This might be a good place to consider using `std::shared_ptr` and `std::weak_ptr` to more gracefully handle making sure pointers/references are valid before use.

albatross/core/fit_model.h
```diff
@@ -0,0 +1,47 @@
...
+  template <typename X, typename Y, typename Z>
+  friend class Prediction;
+
+ public:
+
+  static_assert(std::is_move_constructible<Fit>::value,
+                "Fit type must be move constructible to avoid unexpected copying.");
+
+  FitModel(const ModelType &model,
+           const Fit &&fit)
```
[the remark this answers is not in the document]
akleeman: oh interesting, so is it resorting to a copy the way it's written now?

albatross/core/fit_model.h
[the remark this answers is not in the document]
akleeman: nope, good catch.

albatross/core/fit_model.h
```diff
@@ -0,0 +1,47 @@
...
+      : model_(model), fit_(std::move(fit)) {}
+
+ public:
+
+  template <typename PredictFeatureType>
+  Prediction<ModelType, PredictFeatureType, Fit>
+  get_prediction(const std::vector<PredictFeatureType> &features) const {
+    return Prediction<ModelType, PredictFeatureType, Fit>(*this, features);
```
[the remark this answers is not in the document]
akleeman: I haven't included them in the PR ... but I've been using an approach similar to Eigen for the includes:

https://github.com/eigenteam/eigen-git-mirror/blob/master/Eigen/Core

which I agree makes it a little more difficult to trace back where things come from, but on the plus side it consolidates some of the trickiness of include ordering.  You have opinions on that?

albatross/core/model.h
jbangelo: If I'm understanding you right, you'd like to allow users to write a single `predict` method and for Albatross to extract the return type and use that as a proxy for `PredictTypeIdentity<T>` when selecting which function to use? That should be possible, and pretty snazzy!

albatross/core/model.h
akleeman: Yeah I could definitely do that, what I was going for here was to differentiate between two different cases.  One where no `fit` method was found (`!has_possible_fit`) and another where one was found, but it wasn't valid (`has_possible_fit && !has_valid_fit`) which happens frequently when things like a const is missing, etc ...)

albatross/models/least_squares.h
akleeman: That'll be coming back soon ... but I'll remove it from this PR.

albatross/core/prediction.h
akleeman: Right, it's just a tag type. So would you suggest switching to a signature for predict that looks like:

```
  template <typename FeatureType, typename FitType, typename PredictType>
  PredictType predict(const std::vector<FeatureType> &features,
                                  const FitType &least_squares_fit,
                                  PredictTypeIdentity<PredictType>) const
```
Is it worth throwing `const` in there even though there's nothing to modify?

albatross/core/fit_model.h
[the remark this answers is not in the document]
jbangelo: Yes, I believe the compiler will decay the `const Fit &&` into a `const Fit &` when it can't find an appropriate move constructor and end up using a copy constructor if it's available.

albatross/core/fit_model.h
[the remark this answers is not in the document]
jbangelo: Ah yeah, I later noticed there's a bit of a circular dependency between classes. Having a single include that has everything you need is a fine approach as well.

albatross/core/model.h
akleeman: Yeah exactly!  In some situations you want to be able to specialize method for each `PredictType`.  For example, you may have a much more efficient way of returning only the mean value instead of a full `JointDistribution` in which case using a `PredictIdentity` in the signature is (I think) unavoidable.  But for a basic model implementation you may not want to worry about that, in which case we could infer the `PredictIdentity`.

albatross/core/fit_model.h
akleeman: For now, what if we just made copies of `model_`? I'll look into what switching to smart pointers would involve.

PR #85, opened 2019-03-11
CRTP Refactor

akleeman:
This change switches `albatross` models from using runtime polymorphism (aka virtual inheritance) to using the Curiously Recurring Template Pattern (CRTP).

The motivation for the change was two fold:

1) We found that we often wanted models which built upon other models (`AdaptedModel` and `SerializableRegressionModel` for example) each of these variants on a `RegressionModel` needed to carefully forward on method calls and often the variants needed to be aware of each other.  (if you adapted a serializable model it too should be serializable etc ..) which became quite messy.

2) We found that we wanted to start using a model that was fit with one `FeatureType` to make predictions for another `FeatureType`.  This broke the `RegressionModel` interface.  Effectively, we wanted to be able to create an interface that looked like:
```
struct RegressionModel {
    template <typename OtherFeatureType>
    virtual auto predict(std::vector<OtherFeatureType> &features);
}
```
But [virtual templated functions aren't possible](https://stackoverflow.com/questions/2354210/can-a-class-member-function-template-be-virtual).

Switching to CRTP let us (mostly) flatten the models.  This change completely removes `AdaptedModel` and `SerializableRegressionModel` as well as most of the duplicated variations on ransac.

Note: The code this this PR has already gone through code review in pieces.

- [x] Core model functionaltiy.
- [x] Traits
- [x] Gaussian Processes
- [x] Cross validation.
- [x] Ransac.
- [x] Serialization.
- [x] Tuning.
- [x] Reproduce existing tests.
- [ ] Standardize traits style.
- [x] Update Examples
- [ ] Update Docs

The few outstanding pieces of work have been turned into issues since they (probably) aren't blockers.


seth-swiftnav:
Congrats on getting all the tests to pass. I approve of the refactor. It does mean OE and master albatross will be incompatible for a little while.

PR #87, opened 2019-03-11
CRTP: Gaussian Processes

akleeman:
This converts `gp.h` from the old model format to the new one.  It also shows some examples of how things fit together in `test_models.cc`.

A lot of the previous tests in `test_models` require ransac and cross validation which are coming up soon.  Those tests will be added back in that PR.

### DIFF
--- a/albatross/models/gp.h
+++ b/albatross/models/gp.h
@@ -124,14 +124,14 @@ class GaussianProcessBase
 public:
   GaussianProcessBase()
       : covariance_function_(), model_name_(covariance_function_.get_name()){};
-  GaussianProcessBase(CovFunc &covariance_function)
+  GaussianProcessBase(const CovFunc &covariance_function)
       : covariance_function_(covariance_function),
         model_name_(covariance_function_.get_name()){};
   /*
    * Sometimes it's nice to be able to provide a custom model name since
    * these models are generalizable.
    */
-  GaussianProcessBase(CovFunc &covariance_function,
+  GaussianProcessBase(const CovFunc &covariance_function,
                       const std::string &model_name)
       : covariance_function_(covariance_function), model_name_(model_name){};
   GaussianProcessBase(const std::string &model_name)
--- a/albatross/models/gp.h
+++ b/albatross/models/gp.h
@@ -13,43 +13,49 @@
 #ifndef ALBATROSS_MODELS_GP_H
 #define ALBATROSS_MODELS_GP_H
 
-#include "evaluate.h"
-#include "stdio.h"
-#include <functional>
-#include <memory>
-
-#include "cereal/eigen.h"
-#include "core/model.h"
-#include "core/serialize.h"
-#include "eigen/serializable_ldlt.h"
-
 namespace albatross {
 
-using InspectionDistribution = JointDistribution;
+/*
+ * The Gaussian Process needs to store the training data and the
+ * prior covariance for those data points.  One way to do that would
+ * be to simply store the entire RegressionDataset (features and targets)
+ * as well as the covariance matrix between all features.  The way this
+ * ends up getting used however makes it more efficient to store the
+ * information, defined as:
+ *
+ *   information = prior_covariance^{-1} * y
+ *
+ * where
+ *
+ *   y = targets.mean
+ *
+ * and
+ *
+ *   prior_covariance = cov(features, features) + targets.covariance.
+ *
+ * Furthermore, rather than store the entire prior covariance we just
+ * store the cholesky (LDLT).
+ */
+template <typename CovFunc, typename ImplType, typename FeatureType>
+struct Fit<GaussianProcessBase<CovFunc, ImplType>, FeatureType> {
 
-/* Forward Declarations */
-template <typename FeatureType, typename CovarianceType>
-class GaussianProcessRansac;
-template <typename FeatureType, typename CovarianceType>
-class GaussianProcessRegression;
-template <typename FeatureType, typename CovarianceType>
-inline std::unique_ptr<GaussianProcessRansac<FeatureType, CovarianceType>>
-make_gp_ransac_model(
-    GaussianProcessRegression<FeatureType, CovarianceType> *model,
-    double inlier_threshold, std::size_t min_inliers,
-    std::size_t random_sample_size, std::size_t max_iterations,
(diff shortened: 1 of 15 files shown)


albatross/models/gp.h
```diff
@@ -59,295 +65,252 @@ template <typename FeatureType> struct GaussianProcessFit {
...
+
+  GaussianProcessRegression() : Base(){};
+  GaussianProcessRegression(CovFunc &covariance_function)
+      : Base(covariance_function){};
+  GaussianProcessRegression(CovFunc &covariance_function,
+                            const std::string &model_name)
+      : Base(covariance_function, model_name){};
+
+  // The only reason these are here is to hide the base class implementations.
+  void fit() const = delete;
```
jbangelo: Are these are intentionally hiding the `fit()` and `predict()` functions in `GaussianProcessBase`? If so, why? I would have thought we wanted those to be usable/visible.

albatross/models/gp.h
```diff
@@ -59,295 +65,252 @@ template <typename FeatureType> struct GaussianProcessFit {
...
-  typedef CovarianceFunction CovarianceType;
+  template <typename FitFeatureType>
+  using GPFitType = Fit<GaussianProcessBase<CovFunc, ImplType>, FitFeatureType>;
 
-  GaussianProcessRegression()
+public:
+  GaussianProcessBase()
       : covariance_function_(), model_name_(covariance_function_.get_name()){};
-  GaussianProcessRegression(CovarianceFunction &covariance_function)
+  GaussianProcessBase(CovFunc &covariance_function)
```
jbangelo: Should this be `const CovFunc &covariance_function`? We're copying `covariance_function` anyways it looks like.

albatross/models/gp.h
akleeman: Yeah, it has to intentionally hide them so that `has_valid_fit` and `has_valid_predict` in base class don't end up finding themselves and causing infinite recursion!  There might be a different way around it, but this was all I could immediately think of.

albatross/models/gp.h
akleeman: yeah, good catch.

akleeman:
@seth-swiftnav can you take a pass at the math.  Nothing much has changed from `master` but it has been reorganized a bit.

albatross/models/gp.h
```diff
@@ -59,295 +65,252 @@ template <typename FeatureType> struct GaussianProcessFit {
...
+                       const Eigen::VectorXd &prior_variance,
+                       const Eigen::VectorXd &information,
+                       const Eigen::SerializableLDLT &train_ldlt) {
+  const Eigen::VectorXd pred = gp_mean_prediction(cross_cov, information);
+  // Here we efficiently only compute the diagonal of the posterior
+  // covariance matrix.
+  Eigen::MatrixXd explained = train_ldlt.solve(cross_cov);
+  Eigen::VectorXd explained_variance =
+      explained.cwiseProduct(cross_cov).array().colwise().sum();
+  Eigen::VectorXd marginal_variance = prior_variance - explained_variance;
```
seth-swiftnav: Should this be addition?

tests/test_models.cc
```diff
@@ -10,247 +10,140 @@
...
+TYPED_TEST(RegressionModelTester, test_predict_variants) {
+  auto dataset = this->test_case.get_dataset();
+  auto model = this->test_case.get_model();
 
-  auto loo_indexers = leave_one_out_indexer(dataset);
-  auto loo_predictions =
-      model->template cross_validated_predictions<MarginalDistribution>(
-          dataset, loo_indexers);
+  const auto fit_model = model.get_fit_model(dataset.features, dataset.targets);
+  silly_function_to_increment_stack_pointer();
```
seth-swiftnav: Why is this function needed?

tests/test_models.cc
```diff
@@ -10,247 +10,140 @@
...
-      cross_validated_scores(rmse, dataset, loo_indexers, model.get());
+  const Eigen::VectorXd pred_mean = pred.mean();
 
-  // Here we make sure the cross validated mean absolute error is reasonable.
-  // Note that because we are running leave one out cross validation, the
-  // RMSE for each fold is just the absolute value of the error.
-  EXPECT_LE(cv_scores.mean(), 0.1);
-}
+  const MarginalDistribution marginal = pred.marginal();
+  EXPECT_LE((pred_mean - marginal.mean).norm(), 1e-8);
```
seth-swiftnav: What is the type of `pred` here?
Is this testing that the predict types are consistent, or that conversion from a joint prediction to lower information forms is consistent.

albatross/models/gp.h
```diff
@@ -13,43 +13,49 @@
...
   Eigen::SerializableLDLT train_ldlt;
   Eigen::VectorXd information;
 
-  GaussianProcessFit(){};
+  Fit(){};
 
-  GaussianProcessFit(const std::vector<FeatureType> &features,
-                     const Eigen::MatrixXd &train_cov,
-                     const MarginalDistribution &targets);
+  Fit(const std::vector<FeatureType> &features,
```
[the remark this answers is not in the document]
akleeman: Yeah, I think after this change we could start to allow `fit` methods to take `JointDistribution` targets.  It'd be basically the same as the current version, namely the covariance of the targets would get added to prior.

tests/test_models.cc
akleeman: Heh, yeah.  That's there to try and catch a tricky bug that resulted from `FitModel` and `Prediction` storing references.  Basically there was a situation where everything worked fine *unless* you called any function in between `get_fit_model` and `get_prediction` at which point the stack pointer would increment and overwrite memory.  The bug was a result of me improperly dealing with references to objects.  At the moment I've switched to just copying everything, but eventually we'll want to be smarter about it and when we do I'd like to make sure we don't hit the same issue.

tests/test_models.cc
akleeman: `pred` will be of type `Prediction<ModelType, Feature, FitType>` (see `prediction.h`).  The `Prediction` class is responsible for determining which `predict` methods are available for the given model, feature type and the way it was fit.  The "consistency" test is exactly what you mentioned.  That if you ask for a `JointDistribution` the mean and variance match what you'd get if you asked for the `MarginalDistribution` which has a mean identical to if you'd just asked for the mean.

PR #89, opened 2019-03-15
CRTP: Ransac

akleeman:
Adds the new version of `Ransac`.

This now does away with `GenericRansac` in favor of a `Ransac` class that is itself derived from ModelBase.

Most of this is just re-wiring things, but the evaluation metrics were changed.  They are now templated structs that effectively just act like functions (but allow inspection of the required predict type).


albatross/models/ransac.h
```diff
@@ -127,143 +121,128 @@ ransac(const typename RansacFunctions<FitType>::Fitter &fitter,
...
+  static_assert(
+      std::is_base_of<EvaluationMetric<Eigen::VectorXd>, MetricType>::value ||
+          std::is_base_of<EvaluationMetric<MarginalDistribution>,
+                          MetricType>::value ||
+          std::is_base_of<EvaluationMetric<JointDistribution>,
+                          MetricType>::value,
+      "MetricType must be an EvaluationMetric.");
+
+  std::string get_name() const {
+    return "ransac[" << sub_model_->get_name() << "]";
```
jbangelo: Does this actually work? Shouldn't this be using a `std::stringstream`?

albatross/evaluation/evaluate.h
```diff
@@ -0,0 +1,113 @@
...
+ *   template <typename RequiredPredictType>
+ *   using EvaluationMetric =
+ *         std::function<double(const RequiredPredictType &,
+ *                              const MarginalDistribution &);
+ *
+ * the problem with this is that when you pass the std::function
+ * object as an argument the RequiredPredictType cannot be
+ * inferred.
+ */
+template <typename RequiredPredictType> struct EvaluationMetric {
```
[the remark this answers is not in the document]
akleeman: Yeah, I like that.  I'll give it a shot.

albatross/models/ransac.h
akleeman: heh, good catch.  I actually just fixed that in a subsequent commit, I'll bring it into this one.

albatross/evaluation/evaluate.h
```diff
@@ -68,7 +80,24 @@ negative_log_likelihood(const JointDistribution &prediction,
   return albatross::negative_log_likelihood(mean, covariance);
 }
 
-} // namespace evaluation_metrics
+static inline double
+negative_log_likelihood(const MarginalDistribution &prediction,
+                        const MarginalDistribution &truth) {
+  const Eigen::VectorXd mean = prediction.mean - truth.mean;
+  Eigen::VectorXd variance(prediction.covariance.diagonal());
+  if (truth.has_covariance()) {
```
seth-swiftnav: Can we have a Marginal where `has_covariance()` would be false? By definition, isn't a marginal distribution a mean and a diagonal-only covariance?

akleeman: I think we're *very* close to being able to make it so a `Distribution` by definition has a covariance.  But at the moment we support it to allow for things like a `target` used to fit a model which is mean only.

I might be able to do away with `has_covariance()` entirely if we default to zero variance.  I'll see how hard that'd be.

seth-swiftnav: My question is more about how the function arguments are a `MarginalDistribution`, and not a `Distribution`.

akleeman: If you create a marginal distribution with a mean only `has_covariance()` will return `false`.

PR #91, opened 2019-03-18
CRTP: Tuning

akleeman:
Modifies the tuning scripts to work with the new model design.

The largest difference was switching the `TuneModelConfig` to become a `ModelTuner` which now contains the logic that actually runs tuning.  The new syntax looks like:

```
auto tuner = get_tuner(model, metric, dataset);
tuner.optimizer.set_maxeval(100);
auto params = tuner.tune();
```

instead of:
```
TuneModelConfig<double> config(model_creator, dataset, metric);
config.optimizer.set_maxeval(100);
auto params = tune_regression_model(config);
```

### DIFF
--- a/albatross/tune.h
+++ b/albatross/tune.h
@@ -13,14 +13,10 @@
 #ifndef ALBATROSS_TUNE_H
 #define ALBATROSS_TUNE_H
 
-#include "core/model.h"
-#include "evaluate.h"
-#include "nlopt.hpp"
-#include "tuning_metrics.h"
-#include <map>
-#include <vector>
-
 namespace albatross {
+
+template <typename ModelType, typename MetricType, class FeatureType>
+struct ModelTuner;
 
 inline std::string to_string(const nlopt::result result) {
   std::map<nlopt::result, std::string> result_strings = {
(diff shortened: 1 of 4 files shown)


albatross/tune.h
```diff
@@ -38,122 +34,127 @@ inline std::string to_string(const nlopt::result result) {
...
+    return model.get_params();
+  }
+
+  void
+  initialize_optimizer(const nlopt::algorithm &algorithm = nlopt::LN_PRAXIS) {
+    // The various algorithms in nlopt are coded by the first two characters.
+    // In this case LN stands for local, gradient free.
+    auto tunable_params = model.get_tunable_parameters();
+
+    optimizer = nlopt::opt(algorithm, (unsigned)tunable_params.values.size());
```
seth-swiftnav: Since the algorithm is part of the optimizer, it looks like we can set it without changing the default in Albatross, but it requires knowing the number of dimensions. I've been changing the hardcoded default to NelderMead.
Do you think it's worth either changing the default algorithm, or adding a helper function to change the algorithm of the optimizer without the caller specifying (perhaps wrongly) the number of tunable parameters?

akleeman: What I've been doing is:
```
auto tuner = get_tuner(model, metric, dataset);
tuner.initialize_optimizer(nlopt::LN_NEALDERMEAD);
```
which will take care of the parameter size.  We could rename `initialize_optimizer` to `set_algorithm`? Though that's actually a little misleading since it does do a lot of other non-algorithm specific initializations.

seth-swiftnav: It didn't dawn on me that I could call `initialize_optimizer` again. But I prefer that name to `set_algorithm`. Looks like my request is already solved. (Assuming I can remember the solution).

PR #93, opened 2019-03-19
CRTP: Serialization

akleeman:
This enables things like:
```
  const auto fit_model = model.get_fit_model(dataset);
  // Serialize it
  std::ostringstream os;
  {
    cereal::JSONOutputArchive oarchive(os);
    oarchive(fit_model);
  }
  // Deserialize it.
  std::istringstream is(os.str());
  decltype(fit_model) deserialized;
  {
    cereal::JSONInputArchive iarchive(is);
    iarchive(deserialized);
  }
  fit_model == deserialized;
```
Same with `model.ransac().get_fit_model(dataset)`;

Note: I'm going to add versions to the serialization, but I'm keeping that for a subsequent PR to avoid merging issues in PRs I have lined up already.

### DIFF
--- a/albatross/core/model.h
+++ b/albatross/core/model.h
@@ -92,6 +92,24 @@ private:
   }
 
 public:
+  template <class Archive> void save(Archive &archive) const {
+    archive(cereal::make_nvp("params", derived().get_params()));
+    archive(cereal::make_nvp("insights", insights_));
+  }
+
+  template <class Archive> void load(Archive &archive) {
+    ParameterStore params;
+    archive(cereal::make_nvp("params", params));
+    derived().set_params(params);
+    archive(cereal::make_nvp("insights", insights_));
+  }
+
+  bool operator==(const ModelType &other) const {
+    return (derived().get_params() == other.get_params() &&
+            derived().get_name() == other.get_name() &&
+            derived().insights_ == other.insights_);
+  }
+
   template <typename DummyType = ModelType,
             typename std::enable_if<!has_name<DummyType>::value, int>::type = 0>
   std::string get_name() {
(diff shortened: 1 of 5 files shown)


tests/test_serialize.cc
```diff
@@ -211,156 +194,114 @@ struct DatasetWithMetadata
...
-  RepresentationType create() const override {
-    return std::make_unique<MockModel>(log(2.));
-  }
-
-  bool are_equal(const RepresentationType &lhs,
-                 const RepresentationType &rhs) const override {
-    return *lhs == *rhs;
-  };
-};
+REGISTER_TYPED_TEST_CASE_P(SerializeTest, test_roundtrip_serialize_json,
```
jbangelo: Neat! Didn't know gtest had this available.

albatross/core/model.h
```diff
@@ -92,6 +92,24 @@ template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
   }
 
 public:
+  template <class Archive> void save(Archive &archive) const {
```
jbangelo: This is probably a cereal thing, but it's not immediately obvious what the difference is between `save()` and `serialize()` below.

akleeman: Yeah ... that's a cereal thing.  https://uscilab.github.io/cereal/serialization_functions.html

Basically, if you can get away with directly serializing all members of an object you can usually just define `serialize`, but if you need to go through `get_` and `set_` methods then you end up needing `save` and `load`.

tests/test_serialize.cc
akleeman: Yeah!  This works out quite nice.  I like how it lets you define a set of things you want tested, then you can end up using those in specialized tests (the way we're doing now with `ExampleModels`).  It's led me to some bugs that I would have been too lazy to have tested for otherwise.

PR #94, opened 2019-03-19
CRTP: Rename

akleeman:
Renames:
```
   get_fit_model -> fit
   get_prediction -> predict
   fit -> _fit_impl
   predict -> _predict_impl
```
So the new syntax will look like:
```
model.fit(dataset).predict(features).mean();
```
This makes creating a new model a little less intuitive since you need to write methods for `_fit_impl` and `_predict_impl`.  BUT this mimics the syntax used for writing a new `CovarianceFunction`, matches Eigen and makes it more obvious that you shouldn't do, `model._fit_impl(dataset)`.

### DIFF
--- a/albatross/core/fit_model.h
+++ b/albatross/core/fit_model.h
@@ -37,7 +37,7 @@ public:
 
   template <typename PredictFeatureType>
   Prediction<ModelType, PredictFeatureType, Fit>
-  get_prediction(const std::vector<PredictFeatureType> &features) const {
+  predict(const std::vector<PredictFeatureType> &features) const {
     return Prediction<ModelType, PredictFeatureType, Fit>(model_, fit_,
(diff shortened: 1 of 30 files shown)


seth-swiftnav:
I'm onboard with the renames. But is there a list of other changes made besides the 5 mentioned?
There's other changes I spotted like comment deletion and other renames `_fit<->fit_`.

akleeman:
@seth-swiftnav looks like some additional changes to tests and examples slipped in.  It'll be a lot easier on me to just leave them in which, since they're concentrated on the tests and examples and are going into a branch (not master) I hope is OK.

re other renames, those are all side effects of the other renames.  In particular the private methods of models were switched around to not shadow class members and to match the `_foo` private method syntax.

PR #95, opened 2019-03-20
CRTP: Includes

akleeman:
With CRTP some of the include dependencies aren't super obvious.  In particular, the order in which headers are included is slightly more important.  This PR switches to the style Eigen uses in which block of functionality have their own "header" which ensures it includes what it needs and does so only once.

The way it's organized is that a user can include the basics of albatross (the model base, distributions and datasets) by doing:
```
     #include "albatross/Core"
```
after which additional functionality can be included such as:
```
     #include "albatross/GP"
     #include "albatross/Ransac"
```


CMakeLists.txt
```diff
@@ -34,32 +34,51 @@ CACHE INTERNAL ""
...
+    albatross/tuning_metrics.h
+    examples/example_utils.h
+    examples/sinc_example_utils.h
+    examples/temperature_example/temperature_example_utils.h
+    tests/mock_model.h
+    tests/test_core_distribution.h
+    tests/test_models.h
+    tests/test_serialize.h
     tests/test_utils.h
     CACHE INTERNAL ""
```
jbangelo: I'd recommend that this not be a `CACHE` variable. It'll make it difficult when you dev because you'll need to remember to wipe out your cache file when you add a header.

albatross/Core
```diff
@@ -0,0 +1,30 @@
...
+#include "core/declarations.h"
+#include "core/traits.h"
+#include "core/priors.h"
+#include "core/parameter_handling_mixin.h"
+#include "core/parameter_macros.h"
+#include "core/fit_model.h"
+#include "core/prediction.h"
+#include "core/model.h"
+
+//#include <cereal/types/map.hpp>
```
jbangelo: Missing this commented line?

albatross/Common
```diff
@@ -0,0 +1,40 @@
...
+#include <functional>
+#include <memory>
+#include <vector>
+#include <numeric>
+
+#include "map_utils.h"
+#include "random_utils.h"
+#include "cereal/eigen.h"
+
+#endif
```
jbangelo: Incredibly minor nit: It's good to have a newline at the end of files. It's no longer required by the C++ standard (it was pre-C++11!), but git and other tools show a little indicator and really dumb text manipulating tools (sed/grep/etc) sometimes rely on the last line of a file ending with a newline.

albatross/Evaluation
```diff
@@ -0,0 +1,9 @@
+
```
jbangelo: Missed the copyright and include guards

tests/CMakeLists.txt
```diff
@@ -3,40 +3,40 @@ add_definitions(-DCSV_IO_NO_THREAD)
...
+    test_traits_covariance_functions.cc
+    test_traits_evaluation.cc
+    test_tune.cc
+    test_tuning_metrics.cc
 )
 
 add_dependencies(albatross_unit_tests
     albatross
-#    clang-format-all
+    clang-format-all
```
jbangelo: This seems like an odd dependency to have. Are you sure about it?

akleeman: Nah, I'm not sold on it.  I find it nice to automatically have formatting run all the time, but I can achieve that other ways.

albatross/Common
akleeman: done

albatross/Core
akleeman: done

CMakeLists.txt
akleeman: good call

tests/CMakeLists.txt
jbangelo: In the past I've usually had things like this run as a git hook, but that's more easily done when it's a shell script and not part of CMake. Plus that approach is more difficult to enforce.

akleeman: Yeah, and in this case the unit tests enforce that formatting has been run, so I'll just remove the dependency.

PR #98, opened 2019-03-26
Return Maps from Cross Validation

akleeman:
Cross validation now returns a map from the fold name to a predictions, instead of a vector of predictions.  This makes it so we don't need to rely on preserved order.


akleeman:
yeah, `operator []` doesn't work for `const` maps: https://stackoverflow.com/questions/1474894/why-isnt-the-operator-const-for-stl-maps

PR #99, opened 2019-03-27
Improve cmake and directory organization.

akleeman:
Move all the code to an `include` directory to allow for more conventional use.

Thanks @martin-swift for the cmake makeover!


CMakeLists.txt
```diff
@@ -17,68 +14,22 @@ endif()
...
+  "${PROJECT_SOURCE_DIR}/include"
+  )
+target_include_directories(albatross SYSTEM INTERFACE
+  "${PROJECT_SOURCE_DIR}/third_party/eigen/"
+  "${PROJECT_SOURCE_DIR}/third_party/cereal/include"
+  "${PROJECT_SOURCE_DIR}/third_party/fast-cpp-csv-parser/"
+  )
+
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -fexceptions -frtti")
```
martin-swift: want to fix that too?  take a look at what I did in orion...

include/albatross/Ransac
```diff
@@ -13,6 +13,10 @@
 #ifndef ALBATROSS_RANSAC_H
 #define ALBATROSS_RANSAC_H
 
-#include "models/ransac.h"
+#include "Core"
```
martin-swift: Hmm... interesting.  Totally legit, but I always like to make things consistent; looks like we've done <albatross/Core> in other places, like Tune.

martin-swift:
Also, not seeing where the dependency on the math library is captured.  How are the examples linking?  Do those specific examples just happen to not use the math library?  Trying pulling "m" out of the unit test link libraries and see what happens...

CMakeLists.txt
akleeman: done

include/albatross/Ransac
akleeman: which is better.  If it's a relative path should it be `"Core"`? or `<albatross/Core>`

martin-swift: I think "Core", since it actually is local.  Unless you think these files might move into different directories.

CMakeLists.txt
martin-swift: just not pushed...?

CMakeLists.txt
```diff
@@ -17,68 +14,76 @@ endif()
...
-    tests/test_models.h
-    tests/test_serialize.h
-    tests/test_utils.h
-)
-
-add_custom_target(albatross DEPENDS ${albatross_HEADERS} COMMENT "albatross dependencies")
-
 add_subdirectory(tests)
 add_subdirectory(examples)
+
```
martin-swift: I would set the compile flags above the add_subdirectory(*) and the move the actually setting of the compile options to the CMake where those targets are created because what could (is going to) happen eventually is someone is going to add a target in one of those subdirectories an not realize they also need to add it to albatross_BINARIES.

cmake/ClangTools.cmake
```diff
@@ -21,27 +21,22 @@ endif(CMAKE_CROSSCOMPILING)
...
-        clang-format)
-    message(STATUS "Using clang format: ${CLANG_FORMAT}")
-    if("${CLANG_FORMAT}" STREQUAL "CLANG_FORMAT-NOTFOUND")
-        message(WARNING "Could not find 'clang-format' please set CLANG_FORMAT_PATH:STRING")
-    else()
-        set(CLANG_FORMAT_PATH ${CLANG_FORMAT})
-        message(STATUS "Found: ${CLANG_FORMAT_PATH}")
-    endif()
+  find_program(CLANG_FORMAT NAMES
+    clang-format-3.8
```
martin-swift: ?  I thought we went 4.0 and 6.0?  I can't even get 3.8 on 18.04 without sacrificing chickens...

examples/sinc_example_utils.h
```diff
@@ -13,8 +13,8 @@
 #ifndef ALBATROSS_SINC_EXAMPLE_UTILS_H
 #define ALBATROSS_SINC_EXAMPLE_UTILS_H
 
-#include "GP"
 #include "example_utils.h"
```
martin-swift: ordering?

tests/CMakeLists.txt
```diff
@@ -1,51 +1,54 @@
...
-
-target_link_libraries(albatross_unit_tests m gtest gtest_main pthread gflags nlopt)
-
-add_custom_target(
-run_albatross_unit_tests ALL
-COMMAND albatross_unit_tests
-COMMENT "Running unit tests"
-)
-
+  EXCLUDE_FROM_ALL
```
martin-swift: I prefer to put this on the previous line; that way it's not mingled with the source files.

martin-swift:
on the whole, looking pretty good!

cmake/ClangTools.cmake
akleeman: Yeah, it was originally 3.8 and below, then at the beginning of this PR it was switched and I just reverted back to 3.8 since I can't get 4.0 to install on travis.  https://github.com/swift-nav/albatross/issues/100

PR #103, opened 2019-03-30
Make macros work outside albatross namespace.

akleeman:
As it stands you'd need to do something like:
```
using albatross::ParameterKey;
using albatross::ParameterValue;
using albatross::Parameter;
```
before you can do
```
ALBATROSS_DECLARE_PARAMS(foo, bar);
```
This change makes the `using` statements unnecessary.


akleeman:
@jbangelo good idea.

PR #105, opened 2019-03-30
Rename Metrics

akleeman:
Renames `EvaluationMetric` to `PredictionMetric` and `TuningMetric` to `ModelMetric`.  These names were picked to disambiguate between metrics which require the signature:
```
prediction_metric(prediction, truth);
```
and
```
model_metric(dataset, model);
```


akleeman:
@seth-swiftnav I'm not 100% sold on `ErrorMetric`.  What I do want to do is differentiate between the various types of evaluation metrics we can have.  In particular there are metrics which take an arbitrary prediction and truth and return a value.  Then there are metrics which take a dataset and a model and perhaps do something like cross validation then apply an `ErrorMetric` to get a final value.

The existing names for these are `EvaluationMetric` and `TuningMetric`.  I was going to change them to `ErrorMetric` and `ModelMetric`.  But open to other suggestions.  `PredicitonScore` and `ModelScore`?  The thing I like about `Error` is that it implies larger is worse, where `Score` might imply the opposite.

Kinda seems like `sklearn` mixes and matches the terms metrics and scores: https://scikit-learn.org/stable/modules/model_evaluation.html

seth-swiftnav:
We could have evaluation metrics (prediction and truth) where higher is better, such as likelihood.
If the differentiation in the name is about what the input arguments are, I think that might be the way to pick the names.

PredictionMetric and ModelMetric could work

PR #112, opened 2019-04-09
Creating a Model Based on Predictions From a Model

pgrgich:
Our current architecture is based around a large model that takes in a lot of data (in the AM). To reduce CPU load, we then query this at select locations (in the AQ) and in turn create a small model from this subset of information that we then query repeatedly from the CG.

This PR implements a method that will ensure that the queries from the smaller model will return the same predictions and prediction variances for the data points queried from the larger model (though there is no guarantee that other points will have comparable values, we assume that they will as long as they are sufficiently close to the originally queried values). The current implementation will give a different  distribution to the original in general (though the mean may be the same).

### DIFF
--- a/include/albatross/models/gp.h
+++ b/include/albatross/models/gp.h
@@ -58,8 +58,36 @@ struct Fit<GaussianProcessBase<CovFunc, ImplType>, FeatureType> {
     information = train_ldlt.solve(targets.mean);
   }
 
-  template <typename Archive>
-  void serialize(Archive &archive, const std::uint32_t) {
+  // Create a fit that will replicate the given prediction states and
+  // covariances (targets) for the given features on the basis that k_uu is the
+  // prior covariance matrix relating them. This is useful because it allows us
+  // to create a model based on all the data we wish to include (and therefore
+  // be most accurate), then predict a few selected points based on this full
+  // model (getting the most accuracy possible) but after that only require a
+  // small amount of CPU load and memory to compute predictions close to those
+  // initial selected points while still, presumably, getting excellent
+  // accuracy. At least, predictions arbitrarily close to the initially
+  // requested predictions in the feature domain will also be arbitrarily close
+  // in the prediction distribution domain.
+  Fit(const std::vector<FeatureType> &features,
+      const JointDistribution &targets, const Eigen::MatrixXd &k_uu) {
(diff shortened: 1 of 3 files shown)


tests/test_model_metrics.cc
```diff
@@ -41,4 +41,20 @@ TYPED_TEST(ModelMetricTester, test_sanity) {
   EXPECT_FALSE(std::isnan(metric));
 }
 
-} // namespace albatross
+TEST(test_gp_from_prediction, test_model_from_prediction) {
```
pgrgich: Is there an existing file this should go in or maybe a new one? Doesn't seem to belong here...

akleeman: `test_gp.cc` would probably be a good spot for it.

include/albatross/models/gp.h
```diff
@@ -58,7 +58,21 @@ struct Fit<GaussianProcessBase<CovFunc, ImplType>, FeatureType> {
     information = train_ldlt.solve(targets.mean);
   }
 
-  template <typename Archive> void serialize(Archive &archive) {
+  Fit(const std::vector<FeatureType> &features,
+      const JointDistribution &targets, const Eigen::MatrixXd &k_uu) {
+    train_features = features;
```
akleeman: want to add a brief comment explaining the rationale behind the math here?

tests/test_model_metrics.cc
akleeman: note: `test_gp.cc` doesn't exist yet ... but I'm about to add it in another PR.  Feel free to create the file and add your test, or wait for my PR.

include/albatross/models/gp.h
pgrgich: Added some comments, feel free to suggest improvements. :)

tests/test_model_metrics.cc
pgrgich: I can wait for your PR. :)

akleeman: Actually ... I lied, my tests ended up in `test_sparse_gp.cc` which probably doesn't make sense for these.  Want to either make a new test file (`test_gp.cc`) or throw these in `test_models.cc`?

akleeman:
lgtm!

PR #114, opened 2019-04-10
Sparse Gaussian Processes

akleeman:
This change introduces an approach for approximating a Gaussian process using the Fully Independent Training Conditoinal (FITC) method.

An excellent survey of FITC in the context of other approximation methods can be found [here](http://www.jmlr.org/papers/volume6/quinonero-candela05a/quinonero-candela05a.pdf
).

This particular implementation works by first defining an `InducingPointStrategy`.  These strategies take a vector of features and will return a new vector of features which represent the inducing points, `u`.  For example, in `sparse_example.cc` we start with the same toy problem as `sinc_example.cc`.  The example has training data that is randomly sampled on the x-axis between 0 and 13.  The inducing point strategy in this example finds the bounds of the data and samples k linearly spaced points in the range.

FITC then works by effectively finding the posterior distribution of `u` conditional on the training data `y`, `[u|y]`.  For some new features, `x*` the corresponding predictions, `y*`, are then made by finding the posterior distribution `[y*|u]`.

The process is then:
- Start with some training data (features and targets)
- Find the distribution of the inducing points conditioned on the training data.
- Build a new representative Gaussian process over `u` which results in a `Fit<GaussianProcessBase<...>>` type equivalent to if one had been created using a full Gaussian process.
- Predictions are then made using the default Gaussian process `_predict_impl`.

Because `u` is typically much smaller than `y` this significantly accelerates inference.

### DIFF
--- a/include/albatross/models/sparse_gp.h
+++ b/include/albatross/models/sparse_gp.h
@@ -148,7 +148,7 @@ public:
     // P is such that:
     //     Q_ff = K_fu K_uu^-1 K_uf
     //          = K_fu L^-T L^-1 K_uf
-    //          = P P^T
+    //          = P^T P
     Eigen::MatrixXd P = K_uu_llt.matrixL().solve(K_fu.transpose());
 
     // Efficiently compute the diagonal diag[Q_ff].
--- a/include/albatross/models/sparse_gp.h
+++ b/include/albatross/models/sparse_gp.h
@@ -185,14 +185,17 @@ public:
      *     C^-1 = (K_uu^-1 - S)^-1
      *                                                  (Expansion of S)
      *          = (K_uu^-1 - (K_uu + K_uf A^-1 K_fu)^-1)^-1
-     *                                          (Matrix Inversion Lemma)
-     *          = (K_uu^-1 K_uf (A + K_fu K_uu^-1 K_uf)^-1 K_fu K_uu-1)
+     *                                        (Woodbury Matrix Identity)
+     *          = (K_uu^-1 K_uf (A + K_fu K_uu^-1 K_uf)^-1 K_fu K_uu^-1)
      *                                   (LL^T = K_uu and P = L^-1 K_uf)
      *          = L^-T P (A + P^T P)^-1 P^T L^-1
      *                                        (Searle Set of Identities)
-     *          = L^-T C A^-1 P^T (I + P A^-1 P)^-1 L^-1
-     *                           (B = (I + P A^-1 P) and R = A^-1/2 P^T)
+     *          = L^-T P A^-1 P^T (I + P A^-1 P^T)^-1 L^-1
+     *                         (B = (I + P A^-1 P^T) and R = A^-1/2 P^T)
      *          = L^-T R^T R B^-1 L^-1
+     *
+     *  taking the inverse of that then gives us:
+     *      C   = L B (R^T R)^-1 L^T
      *
      *  reusing some of the precomputed values there leads to:
      *
@@ -212,8 +215,8 @@ public:
     Eigen::MatrixXd L_uu_inv =
         K_uu_llt.matrixL().solve(Eigen::MatrixXd::Identity(m, m));
     Eigen::MatrixXd RtRBiLi = RtR * B_ldlt.solve(L_uu_inv);
-    Eigen::MatrixXd C =
-        (K_uu_llt.matrixL().transpose().solve(RtRBiLi)).inverse();
+    Eigen::MatrixXd LT = K_uu_llt.matrixL().transpose();
+    Eigen::MatrixXd C = K_uu_llt.matrixL() * B * RtR.ldlt().solve(LT);
 
     return typename Base::template GPFitType<FeatureType>(u, C.ldlt(), v);
   }
--- a/include/albatross/models/sparse_gp.h
+++ b/include/albatross/models/sparse_gp.h
@@ -228,7 +228,8 @@ public:
   IndexingFunction independent_group_indexing_function;
 };
 
-template <typename CovFunc, typename InducingPointStrategy, typename IndexingFunction>
+template <typename CovFunc, typename InducingPointStrategy,
+          typename IndexingFunction>
 auto sparse_gp_from_covariance(CovFunc covariance_function,
                                InducingPointStrategy &strategy,
                                IndexingFunction &index_function,
@@ -237,6 +238,6 @@ auto sparse_gp_from_covariance(CovFunc covariance_function,
                                          IndexingFunction>(
       covariance_function, strategy, index_function, model_name);
 };
-}
+} // namespace albatross
 
 #endif /* INCLUDE_ALBATROSS_MODELS_SPARSE_GP_H_ */
--- a/include/albatross/models/sparse_gp.h
+++ b/include/albatross/models/sparse_gp.h
@@ -1,0 +1,242 @@
+/*
+ * Copyright (C) 2019 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef INCLUDE_ALBATROSS_MODELS_SPARSE_GP_H_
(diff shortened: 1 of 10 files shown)


include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,242 @@
...
+    Eigen::Index n = targets.mean.size();
+    Eigen::Index m = static_cast<Eigen::Index>(u.size());
+
+    // Create the covariance matrices we need.
+    Eigen::VectorXd K_ff_diag = this->covariance_function_.diagonal(features);
+    if (targets.has_covariance()) {
+      K_ff_diag += targets.covariance.diagonal();
+    }
+    Eigen::MatrixXd K_fu = this->covariance_function_(features, u);
+    Eigen::MatrixXd K_uu = this->covariance_function_(u);
```
pgrgich: Do we need to reference `this`?

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,242 @@
...
+
+    auto K_uu_llt = K_uu.llt();
+    // P is such that:
+    //     Q_ff = K_fu K_uu^-1 K_uf
+    //          = K_fu L^-T L^-1 K_uf
+    //          = P P^T
+    Eigen::MatrixXd P = K_uu_llt.matrixL().solve(K_fu.transpose());
```
pgrgich: Either the comment or the variable name should change, this is P^T.

include/albatross/models/sparse_gp.h
akleeman: Yeah, `covariance_function_` is inherited so the compiler gets confused otherwise.

include/albatross/models/sparse_gp.h
akleeman: nice catch.

akleeman:
@pgrgich Just added a test to make sure the the sparse version is faster.  On a test set with 1000 data points and 100 inducing points it's fitting in about 15% of the time the direct method requires (0.25s instead of 1.8).

pgrgich:
Ok, so that'd cut a ~72 second fit down to 10 seconds. That'll help. :)

akleeman:
@pgrgich Yeah, it'll depend a LOT on the number of inducing points, but should definitely help.

Re:linear algebra.  I did borrow most of that from other implementations ... but not exactly.  I had to wing it through a few parts there, in particular the final steps where I create a new `Fit<GP<...>>`.  If you have time to look closely at that and see if I'm doing anything inefficient that'd be awesome.  On the other hand, this is clearly faster already, so we could wait to optimize the details until needed.

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,243 @@
...
+
+    Eigen::VectorXd v = P * (A.asDiagonal().inverse() * targets.mean);
+    v = B_ldlt.solve(v);
+    v = K_uu_llt.matrixL().transpose().solve(v);
+
+    Eigen::MatrixXd L_uu_inv =
+        K_uu_llt.matrixL().solve(Eigen::MatrixXd::Identity(m, m));
+    Eigen::MatrixXd RtRBiLi = RtR * B_ldlt.solve(L_uu_inv);
+    Eigen::MatrixXd C =
+        (K_uu_llt.matrixL().transpose().solve(RtRBiLi)).inverse();
```
pgrgich: I think we can save this `inverse` by recognizing that `C=(L^-T R^T R B^-1 L^-1)^{-1}`=L B (R^T R)^{-1}  L^T`. Considering `R` is made up of `P`, which is made up of `K_uf`, I don't think `R` will be invertible, so we can't get rid of the inner `ldlt`. Nonetheless, an `ldlt` is a little cheaper than an inversion...

pgrgich: Otherwise, this comes back to the issue I raised in the other PR whereby we could simply transport `C^{-1}` rather than `C.ldlt()` (which in this case requires an inversion and an `ldlt` decomposition as opposed to nothing).

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,243 @@
...
+     *  and to get C we need to do some algebra,
+     *
+     *     K_** - K_*u * C * K_u* = K_** - Q_** + K_*u S K_u*
+     *                            = K_** - K_*u (K_uu^-1 - S) K_u*
+     *  which leads to:
+     *     C^-1 = (K_uu^-1 - S)^-1
+     *                                                  (Expansion of S)
+     *          = (K_uu^-1 - (K_uu + K_uf A^-1 K_fu)^-1)^-1
+     *                                          (Matrix Inversion Lemma)
+     *          = (K_uu^-1 K_uf (A + K_fu K_uu^-1 K_uf)^-1 K_fu K_uu-1)
```
pgrgich: `K_uu-1` should be `K_uu^-1`

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,243 @@
...
+     *
+     *  and to get C we need to do some algebra,
+     *
+     *     K_** - K_*u * C * K_u* = K_** - Q_** + K_*u S K_u*
+     *                            = K_** - K_*u (K_uu^-1 - S) K_u*
+     *  which leads to:
+     *     C^-1 = (K_uu^-1 - S)^-1
+     *                                                  (Expansion of S)
+     *          = (K_uu^-1 - (K_uu + K_uf A^-1 K_fu)^-1)^-1
+     *                                          (Matrix Inversion Lemma)
```
pgrgich: Which matrix inversion lemma?

include/albatross/models/sparse_gp.h
akleeman: Yeah, good catch.  For now I removed the inverse following your suggestion.

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,243 @@
...
+     *          = (K_uu^-1 K_uf (A + K_fu K_uu^-1 K_uf)^-1 K_fu K_uu-1)
+     *                                   (LL^T = K_uu and P = L^-1 K_uf)
+     *          = L^-T P (A + P^T P)^-1 P^T L^-1
+     *                                        (Searle Set of Identities)
+     *          = L^-T C A^-1 P^T (I + P A^-1 P)^-1 L^-1
```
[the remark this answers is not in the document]
akleeman: Nice catch.

include/albatross/models/sparse_gp.h
akleeman: https://en.wikipedia.org/wiki/Woodbury_matrix_identity

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,246 @@
...
+     *  and to get C we need to do some algebra,
+     *
+     *     K_** - K_*u * C * K_u* = K_** - Q_** + K_*u S K_u*
+     *                            = K_** - K_*u (K_uu^-1 - S) K_u*
+     *  which leads to:
+     *     C^-1 = (K_uu^-1 - S)^-1
+     *                                                  (Expansion of S)
+     *          = (K_uu^-1 - (K_uu + K_uf A^-1 K_fu)^-1)^-1
+     *                                        (Woodbury Matrix Identity)
+     *          = (K_uu^-1 K_uf (A + K_fu K_uu^-1 K_uf)^-1 K_fu K_uu^-1)
```
pgrgich: I'm pretty sure this is missing a `^-1` at the end, since I think the Woodbury identity gives:

`K_uu^-1 - (K_uu + K_uf A^-1 K_fu )^-1 = K_uu^-1 K_uf ( A + K_fu K_uu^-1 K_uf )^-1 K_fu K_uu^-1`

and we want the inverse of the left hand side.

akleeman: I think this is OK right?  The left hand side is `C^-1`.

pgrgich: Yeah, I was talking about the final inverse. I would have assumed that passing the test would mean it's right, but I struggle to see how it can be, considering the code seems to match the comments. Unless `C^-1=C` for some odd reason.

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,246 @@
...
+     *
+     *     [f*|f=y] ~ N(K_*u S K_uf^-1 A^-1 y, K_** − Q_** + K_*u S K_u*)
+     *
+     *  we can find v easily,
+     *
+     *     v = S K_uf^-1 A^-1 y
+     *
+     *  and to get C we need to do some algebra,
+     *
+     *     K_** - K_*u * C * K_u* = K_** - Q_** + K_*u S K_u*
```
pgrgich: `C` should be `C^-1`

include/albatross/models/sparse_gp.h
```diff
@@ -0,0 +1,246 @@
...
+     *                            = K_** - K_*u (K_uu^-1 - S) K_u*
+     *  which leads to:
+     *     C^-1 = (K_uu^-1 - S)^-1
```
pgrgich: `C^-1 = K_uu^-1 - S)`

PR #115, opened 2019-04-15
Forwarded Inheritence

akleeman:
This PR switches from using an inheritance pattern that looks like `Base<Derived<Impl>>` to one that looks like `Base<Impl>`.  Or more specifically:
```
template <typename X>
struct Base {
  X &derived() { return *static_cast<X *>(this); }
};

template <typename Y>
struct Derived : public Base<Derived<Y>> {
  Y &impl() { return *static_cast<Y *>(this); }
};
```
To one that looks like:
```
template <typename X>
struct Base {
  X &derived() { return *static_cast<X *>(this); }
};

template <typename Y>
struct Derived : public Base<Y> {};
```
The reason for this is that in the first situation (the way it is now) calling `derived()` gives you an instance of `Derived<Y>` which is *almost* useful, but means that the `Derived` class needs to know about anything that could possibly be implemented in `Y` and needs to be sure to subsequently define and forward on those methods.  Ie, if we want to have `foo` available to the `Base` class methods then we'd need to define:
```
template <typename Y>
struct Derived : public Base<Derived<Y>> {
  auto foo() {
    // specialized to use a default value.
    return impl()._foo_impl(0.);
  }
  Y &impl() { return *static_cast<Y *>(this); }
};
```
This pattern also led to some confusing name hiding situations (see the `GaussianProcessRegression` class).  By switching to direct (or forwarded?) inheritance we avoid these problems since `derived()` in the new approach will return an object of type `Y`.

Effectively this ends up pushing the responsibility of what methods should be available from intermediary classes (such as `Derived`) to the final implementations `Y` so that before some final implementation might have looked like,
```
template <typename Y>
struct Derived : public Base<Derived<Y>> {
  auto foo() {
    // specialized to use a default value.
    return impl()._foo_impl(0.);
  }
  Y &impl() { return *static_cast<Y *>(this); }
};

struct Impl : public Derived<Impl> {
  double _foo_impl(double x) {return x};
}
```
But can now look like:
```
template <typename Y>
struct Derived : public Base<Derived<Y>> {
  auto foo() {
    // specialized to use a default value.
    return impl()._foo_impl(0.);
  }
  Y &impl() { return *static_cast<Y *>(this); }
};

struct Impl : public Derived<Impl> {
  using Base = Derived<Impl>;
  // unhide foo
  using Base::foo;

  double foo(double x) {return x};
}
```

### DIFF
--- a/include/albatross/models/gp.h
+++ b/include/albatross/models/gp.h
@@ -151,8 +151,7 @@ gp_joint_prediction(const Eigen::MatrixXd &cross_cov,
  * to define a custom ImplType.  See test_models.cc for an example.
  */
 template <typename CovFunc, typename ImplType>
-class GaussianProcessBase
-    : public ModelBase<GaussianProcessBase<CovFunc, ImplType>> {
+class GaussianProcessBase : public ModelBase<ImplType> {
 
 protected:
   template <typename FitFeatureType>
@@ -220,8 +219,7 @@ public:
   // FeatureType but the CovarianceFunction does.
   template <typename FeatureType,
             typename std::enable_if<
-                has_call_operator<CovFunc, FeatureType, FeatureType>::value &&
-                    !has_valid_fit<ImplType, FeatureType>::value,
+                has_call_operator<CovFunc, FeatureType, FeatureType>::value,
                 int>::type = 0>
   GPFitType<FeatureType> _fit_impl(const std::vector<FeatureType> &features,
                                    const MarginalDistribution &targets) const {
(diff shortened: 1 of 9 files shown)


include/albatross/models/gp.h
```diff
@@ -411,17 +381,16 @@ class GaussianProcessRegression
     : public GaussianProcessBase<CovFunc, GaussianProcessRegression<CovFunc>> {
 public:
   using Base = GaussianProcessBase<CovFunc, GaussianProcessRegression<CovFunc>>;
+  using Base::Base;
```
jbangelo: This pulls forward the constructors from `Base` in place of defining constructors for `GaussianProcessRegression` that just call the equivalent constructor in `Base`?

akleeman: Yeah exactly.

jbangelo: Nice, didn't know that this could be done so elegantly!

PR #120, opened 2019-04-17
Generalize Sparse Gaussian Processes to support PITC

akleeman:
This change lets the `SparseGaussianProcess` model take an arbitrary grouping function where each group is treated independent from each other group.  This is referred to as "Partially Independent Training Conditional" or PITC.

### DIFF
--- a/include/albatross/models/sparse_gp.h
+++ b/include/albatross/models/sparse_gp.h
@@ -81,7 +81,7 @@ struct UniformlySpacedInducingPoints {
  * that it is block diagonal.  These assumptions lead to an efficient way of
  * inferring the posterior distribution for some new location f*,
  *
- *     [f*|f=y] ~ N(K_*u S K_uf A^-1 y, K_** − Q_** + K_*u S K_u*)
+ *     [f*|f=y] ~ N(K_*u S K_uf A^-1 y, K_** - Q_** + K_*u S K_u*)
  *
  *  Where S = (K_uu + K_uf A^-1 K_fu)^-1 and A = diag(K_ff - Q_ff) and "diag"
  * may mean diagonal or block diagonal.  Regardless we end up with O(m^2n)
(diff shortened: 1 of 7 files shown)


include/albatross/models/sparse_gp.h
```diff
@@ -202,14 +224,15 @@ class SparseGaussianProcessRegression
      *
      *     v = L^-T B^-1 P * A^-1 y
      */
-    Eigen::VectorXd A_sqrt = A.array().sqrt();
-    Eigen::MatrixXd RtR = A_sqrt.asDiagonal().inverse() * P.transpose();
+    auto A_llt = A.llt();
```
pgrgich: Could this be less efficient if it's diagonal? I assume `Eigen` will make sure it isn't...

include/albatross/models/sparse_gp.h
```diff
@@ -202,14 +224,15 @@ class SparseGaussianProcessRegression
...
+    Eigen::MatrixXd Pt = P.transpose();
+    Eigen::MatrixXd RtR = A_llt.matrixL().llt().solve(Pt);
```
pgrgich: Is this correct? It should be `A^-1/2`, right?

tests/test_sparse_gp.cc
```diff
@@ -70,17 +94,17 @@ TEST(test_sparse_gp, test_scales) {
...
+      sparse_gp_from_covariance(covariance, strategy, indexer, "sparse");
 
   start = high_resolution_clock::now();
   auto sparse_fit = sparse.fit(large_dataset);
   end = high_resolution_clock::now();
   auto sparse_duration = duration_cast<microseconds>(end - start).count();
 
   // Make sure the sparse version is a lot faster.
-  EXPECT_LT(sparse_duration, 0.2 * direct_duration);
+  EXPECT_LT(sparse_duration, 0.3 * direct_duration);
```
pgrgich: It got a little slower?

include/albatross/models/sparse_gp.h
akleeman: Yeah, it might be slightly less efficient ... but I didn't notice any significant difference in the unit tests.

include/albatross/models/sparse_gp.h
akleeman: Yeah the matrixL part is the square root, could make that more obvious with naming.

tests/test_sparse_gp.cc
akleeman: Only the PITC version got slower, full FITC end up the same (at least as far as I could tell)

include/albatross/models/sparse_gp.h
```diff
@@ -124,43 +124,61 @@ class SparseGaussianProcessRegression
...
+      auto P_cols = P.block(0, i, P.rows(), cols);
+      Q_ff.blocks.emplace_back(P_cols.transpose() * P_cols);
+      i += cols;
+    }
 
-    Eigen::VectorXd A = K_ff_diag - Q_ff_diag;
+    auto A = K_ff - Q_ff;
 
-    if (A.minCoeff() < 1e-6) {
+    if (A.diagonal().minCoeff() < 1e-6) {
```
seth-swiftnav: Magic numbers. Does this lower bound make sense for all GPs, or just the OE use case?

PR #121, opened 2019-04-18
Directory Restructure

akleeman:
This change creates a `src` sub-directory to `albatross` which includes all the actual code.  The primary reason for this re-org was to avoid name collisions on OSX when files with names like `Evaluation` were next to directories named `evaluation`.

Pros of this choice:
- It makes it very clear that users should focus on including things like `<albatross/Core>`.
- It mimics Eigen's include pattern.

Cons
- If a specific file is needed it looks a little awkward: `#include <albatross/src/utils/map_utils.h>`.


akleeman:
Build on this is failing ... but it looks to be an issue with python building the docs due to and updated package so I'm ignoring it for now.

PR #124, opened 2019-04-30
Reintroduce metadata to the Distributions

akleeman:
In the recent refactor I removed the metadata field from the `Distribution` class.  Turns out that's actually pretty useful from time to time, so now it's back!


seth-swiftnav:
Looks good. Any example on when the Distribution metadata is useful? I'm just curious.

PR #126, opened 2019-05-09
Import standard cmake modules, standardise dependency resolution [ESD-1246] [ESD-1247]

woodfell:
Integrates with https://github.com/swift-nav/cmake/pull/2


examples/CMakeLists.txt
```diff
@@ -85,5 +85,6 @@ set(albatross_example_BINARIES
   )
 foreach(TARGET IN LISTS albatross_example_BINARIES)
   target_compile_options(${TARGET} PRIVATE ${albatross_COMPILE_OPTIONS})
+  set_target_properties(${TARGET} PROPERTIES CXX_STANDARD 14)
```
jbangelo: Do we want to also enforce no compiler extensions?

tests/CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ target_link_libraries(albatross_unit_tests
   nlopt
   z
   )
+set_target_properties(albatross_unit_tests PROPERTIES CXX_STANDARD 14)
```
jbangelo: Same here, are compiler extensions wanted?

akleeman:
lgtm

examples/CMakeLists.txt
woodfell: Done

tests/CMakeLists.txt
woodfell: Done

.gitmodules
```diff
@@ -1,18 +1,21 @@
 [submodule "third_party/eigen"]
 	path = third_party/eigen
-	url = https://github.com/PX4/eigen.git
+	url = https://github.com/swift-nav/eigen.git
```
martin-swift: what is the motivation for this?  we will pick up Peddie's optimizations, but rewind a lot of upstream bug fixes.  Do we want to pull upstream into our fork...?   @anth-cole @akleeman @scarcanague 

Also, I understand that Peddie's optimizations were primarily for starling, no? Also, why are we using the PX4 fork vs. the official mirror? https://github.com/eigenteam/eigen-git-mirror

benjamin0: I think we should use the same fork we use in Starling. @akleeman sound good to you?

akleeman: Yeah same as starling makes sense

akleeman: Yeah same as starling makes sense

woodfell: I made this change only to get the URLs synchronised and purposefully left the SHA the same. I'll update it so that albatross and starling point at the same commit

PR #127, opened 2019-05-10
Memory Allocation Bug.

akleeman:
Avoid `malloc` in favor of a vector when serializing eigen data.


mfine:
Nice!

mfine:
Just flagging another instance: https://github.com/swift-nav/albatross/blob/c15e637d726870fd18827f77ca19aea57066d4d0/include/albatross/src/cereal/eigen.hpp#L67

PR #128, opened 2019-05-13
Seth/nll logo metric

seth-swiftnav:
Adds a LeaveOutGroupOutLikelihood ModelMetric class.

### DIFF
--- a/include/albatross/src/evaluation/model_metrics.hpp
+++ b/include/albatross/src/evaluation/model_metrics.hpp
@@ -15,8 +15,9 @@
 
 namespace albatross {
 
-template <typename MetricType> class ModelMetric {
-public:
+template <typename MetricType>
+class ModelMetric {
+ public:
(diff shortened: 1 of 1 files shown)


include/albatross/src/evaluation/model_metrics.hpp
```diff
@@ -64,6 +65,27 @@ struct LeaveOneOutLikelihood
   }
 };
 
+template <typename PredictType, typename FeatureType>
```
akleeman: did you want to do the thing where you go:
```
template <typename FeatureType, typename PredictType = JointDistribution>
```
?

seth-swiftnav: I had some templating errors when I tried the first time... but I think I just missed one of the three spots that are affected. I'll update with my next try to get clang formatting correct.

PR #131, opened 2019-06-03
Add truncated_psd_solve method

akleeman:
The `truncated_psd_solve` method let's you solve symmetric but numerically negative (or nearly negative) definite systems by implicitly using the pseudo inverse.

Ie, given a matrix with eigen decomposition `A = V D V^T` and set of equations `A x = b` this will solve for `x` using `x = V D'^-1 V^T b` where `D'^-1` is a diagonal matrix which is equal to `D^-1` but where any values in `D` which are below `threshold` are zeroed.

### DIFF
--- a/include/albatross/SparseGP
+++ b/include/albatross/SparseGP
@@ -16,6 +16,7 @@
 #include "GP"
 
 #include "src/utils/block_utils.hpp"
+#include "src/utils/eigen_utils.hpp"
 #include "src/models/sparse_gp.hpp"
 
 #endif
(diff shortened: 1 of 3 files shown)


pgrgich:
I think that should be `x = V^-T D'^-1 V^-1 b`, shouldn't it?

akleeman:
Yeah exactly, since we're dealing with PSD matrices the eigen vectors are all perpendicular and you get: `V^-1 = V^T`.

PR #134, opened 2019-06-05
Fit From Prediction Stability

akleeman:
This fixes an issue in which `fit_from_prediction` becomes unstable when the prediction used as unobservable direction.

The `fit_from_prediction` method produces a matrix `C` and vector `v` when take the form,
```
v = prior^-1 y
C = prior (prior - prediction)^-1 prior
```
These are then used in the standard Gaussian process prediction formula,
```
m* = cross * v
S* = prior_S - cross * C^-1 * cross^T
``` 
such that when `cross = prior` and `prior_S = prior` you end up recovering the original prediction.  As mentioned this all works fine until `prediction` has an unobserved direction.  As an extreme case consider when `prediction = prior` (ie, nothing was observed).  Computing `C` in this case would require inverting a zero and since the current code computes `C` then uses it's cholesky decomposition to perform the resulting inversion we end up with an extremely unstable situation.

This change fixes this by allowing various options for storing the `C` matrix.  These are named `CovarianceRepresentations` and their only requirement (for now) is that they contain a `solve` method which computes the inversion.  This lets us preserve the existing approach for direct Gaussian processes, then when creating a fit from a prediction we can use a different `CovarianceRepresentation` which I've named the `ExplainedVariance` representation which stores both `prior` and `(prior-prediction)`.  The resulting solve can then use the formula `C^-1 = prior^-1 (prior - prediction) prior^-1` and avoids inverting the potentially singular inner matrix.

Doing so required a few larger changes:

- The `Fit` class is no longer templated on the `ModelType` and `FeatureType` and instead takes a single  template parameter.
- Because `Fit` is no longer tied to a specific model the `is_valid_fit` traits had to be reworked.
- The `Ransac` and in turn `AdaptedRansac` test models had to be modified.
- A test case was set up which fails on master but passes with this modification.

### DIFF
--- a/tests/test_eigen_utils.cc
+++ b/tests/test_eigen_utils.cc
@@ -1,5 +1,5 @@
 /*
-h * Copyright (C) 2018 Swift Navigation Inc.
+ * Copyright (C) 2018 Swift Navigation Inc.
  * Contact: Swift Navigation <dev@swiftnav.com>
  *
  * This source is subject to the license found in the file 'LICENSE' which must
--- a/tests/test_eigen_utils.cc
+++ b/tests/test_eigen_utils.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Swift Navigation Inc.
+h * Copyright (C) 2018 Swift Navigation Inc.
  * Contact: Swift Navigation <dev@swiftnav.com>
  *
  * This source is subject to the license found in the file 'LICENSE' which must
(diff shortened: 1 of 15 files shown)


tests/test_eigen_utils.cc
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Swift Navigation Inc.
+h * Copyright (C) 2018 Swift Navigation Inc.
```
pgrgich: Looks like a problem. :)

akleeman:
@seth-swiftnav, correct it'll still fail with duplicates. This just deals with the fit from predict.

PR #135, opened 2019-06-07
Add Measurement Noise

akleeman:
This change started as an attempt to kill a few birds with one stone.  In particular, I was hoping to fix Gaussian processes from breaking when duplicate observations are present AND add the ability to differentiate between the underlying process being modeled and measurements taken of the process.  In the end this change only accomplishes the later, but I think it's still worth bringing in.

This lets you know wrap a `FeatureType` with a `Measurement<FeatureType>` tag which let's you then create covariance functions which will only evaluate non-zero when building a training covariance, but which will then become null operations when testing.  See `test_measurement_noise_wrapper` for example behavior.

The motivation for this is that often when you make a prediction of some quantity you actually care about the underlying process NOT a measurement of that process.  Consider the temperature example.  You start with a bunch of stations all measuring the temperature then want to build a map of your best guess of the temperature.  Of course each station has measurement noise and it's important to  model this when training.  When predicting however you likely want to covariance to reflect the model's confidence in it's estimate of the actual temperature, not what a station would measure.

There are some exceptions to this (evaluation for example) in which you want to ask how well your model predicts held out measurements in which case you want the measurement noise to be a part of the prediction.  As a result the cross validation routines keep the measurement noise throughout.

### DIFF
--- a/include/albatross/src/covariance_functions/measurement.hpp
+++ b/include/albatross/src/covariance_functions/measurement.hpp
@@ -15,15 +15,32 @@
 
 namespace albatross {
 
+template <typename X> struct Measurement {
+
+  Measurement(){};
+
+  Measurement(const X &x) { value = x; }
+
+  X value;
+};
+
 template <typename SubCovariance>
 class MeasurementOnly
     : public CovarianceFunction<MeasurementOnly<SubCovariance>> {
 
 public:
-  ~MeasurementOnly(){};
+  MeasurementOnly(){};
+  MeasurementOnly(const SubCovariance &sub_cov) : sub_cov_(sub_cov){};
 
   std::string name() const {
     return "measurement[" + sub_cov_.get_name() + "]";
+  }
+
+  ParameterStore get_params() const override { return sub_cov_.get_params(); }
+
+  void unchecked_set_param(const ParameterKey &name,
+                           const Parameter &param) override {
+    sub_cov_.set_param(name, param);
   }
 
   /*
@@ -53,6 +70,14 @@ public:
 private:
   SubCovariance sub_cov_;
 };
+
+/*
+ * Utility function to act as a constructor but with template param resolution.
+ */
+template <typename SubCovariance>
+MeasurementOnly<SubCovariance> measurement_only(const SubCovariance &cov) {
+  return MeasurementOnly<SubCovariance>(cov);
+}
 
 } // namespace albatross
 
--- a/include/albatross/src/covariance_functions/measurement.hpp
+++ b/include/albatross/src/covariance_functions/measurement.hpp
@@ -17,7 +17,7 @@ namespace albatross {
 
 template <typename X> struct Measurement {
 
-  Measurement() : value() {};
+  Measurement() : value(){};
 
   Measurement(const X &x) { value = x; }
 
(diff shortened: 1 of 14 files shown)


include/albatross/src/covariance_functions/measurement.hpp
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+#ifndef INCLUDE_ALBATROSS_SRC_COVARIANCE_FUNCTIONS_MEASUREMENT_HPP_
+#define INCLUDE_ALBATROSS_SRC_COVARIANCE_FUNCTIONS_MEASUREMENT_HPP_
+
+namespace albatross {
+
+template <typename X> struct Measurement {
+
+  Measurement(){};
```
jbangelo: Wouldn't it make sense to call `X`'s default constructor here as well? Otherwise it won't be initialized.

include/albatross/src/covariance_functions/measurement.hpp
```diff
@@ -0,0 +1,84 @@
...
+  /*
+   * This will create a scaled identity matrix, but only between
+   * two different observations defined by the Observed type.
+   */
+  template <
+      typename X, typename Y,
+      typename std::enable_if<
+          has_valid_call_impl<SubCovariance, X &, Y &>::value, int>::type = 0>
+  double _call_impl(const X &x, const Y &y) const {
+    return 0.;
```
jbangelo: Does this match the comment above?

include/albatross/src/covariance_functions/measurement.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+  X value;
+};
+
+template <typename SubCovariance>
+class MeasurementOnly
+    : public CovarianceFunction<MeasurementOnly<SubCovariance>> {
+
+public:
+  MeasurementOnly(){};
```
jbangelo: Again, should probably be calling the subcovariance's default constructor here.

tests/test_sparse_gp.cc
```diff
@@ -73,6 +76,9 @@ TYPED_TEST(SparseGaussianProcessTest, test_sanity) {
   double really_sparse_cov_diff =
       (really_sparse_pred.covariance - direct_pred.covariance).norm();
 
+  std::cout << direct_pred.covariance << std::endl;
```
jbangelo: Is this still needed?

include/albatross/src/covariance_functions/measurement.hpp
akleeman: yes, good catch.

include/albatross/src/covariance_functions/measurement.hpp
akleeman: Ha! not even close.  I'll delete it thanks.

akleeman:
In this change I opted for adding the `predict_with_measurement_noise` method.  Ie, defaulting to predicting without measurement noise included ... but I'm torn about whether that's the way to go or not.  As another example [this package](https://github.com/SheffieldML/GPy/blob/devel/GPy/core/gp.py) uses `predict_noiseless`, so picked the opposite default.

PR #136, opened 2019-06-07
Add Serialization to the ExplainedCovariance class

akleeman:
This makes it so a Gaussian process built from `fit_from_prediction` can be serialized.

### DIFF
--- a/tests/test_serialize.cc
+++ b/tests/test_serialize.cc
@@ -113,11 +113,11 @@ struct LDLT : public SerializableType<Eigen::SerializableLDLT> {
   Eigen::Index n = 3;
 
   RepresentationType create() const override {
-    // This looks weird, but without forcing this to a Eigen::MatrixXd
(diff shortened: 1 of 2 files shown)


tests/test_serialize.cc
```diff
@@ -136,6 +136,29 @@ struct LDLT : public SerializableType<Eigen::SerializableLDLT> {
   };
 };
 
+struct ExplainedCovarianceRepresentation
+    : public SerializableType<ExplainedCovariance> {
+  Eigen::Index n = 3;
+
+  RepresentationType create() const override {
```
seth-swiftnav: Does Eigen::MatrixXd::Random()not return an Eigen::MatrixXd?
If it's having trouble, why not skip using "auto " instead?

akleeman: Yeah I could probably just skp auto.  I'll give that a shot.

PR #137, opened 2019-06-20
Use mapbox/variant

akleeman:
This change let's you use `variant<X, Y>` as an argument to a covariance function as long as that function supports operations on both `X` and `Y`.

### DIFF
--- a/include/albatross/src/covariance_functions/covariance_function.hpp
+++ b/include/albatross/src/covariance_functions/covariance_function.hpp
@@ -13,6 +13,8 @@
 #ifndef ALBATROSS_COVARIANCE_FUNCTIONS_COVARIANCE_FUNCTION_H
 #define ALBATROSS_COVARIANCE_FUNCTIONS_COVARIANCE_FUNCTION_H
 
+using mapbox::util::variant;
+
 namespace albatross {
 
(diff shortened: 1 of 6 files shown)


include/albatross/src/covariance_functions/covariance_function.hpp
```diff
@@ -135,15 +137,71 @@ template <typename CovFunc, typename Caller> struct MeasurementForwarder {
...
+      typename A, typename B, typename C,
+      typename std::enable_if<
+          !has_valid_cov_caller<CovFunc, Caller, variant<A, B>, C>::value &&
+              has_valid_cov_caller<CovFunc, Caller, A, C>::value &&
+              has_valid_cov_caller<CovFunc, Caller, B, C>::value,
+          int>::type = 0>
+  static double call(const CovFunc &cov_func, const variant<A, B> &x,
+                     const C &y) {
+
+    return x.match([&y, &cov_func](A xx) { return call(cov_func, xx, y); },
```
jbangelo: Have you tried having these lambdas take a parameter of `const A&`? That would avoid a copy of `A`.

akleeman: Nice!  That works. I tried it before without luck ... but must have been a different issue.

PR #140, opened 2019-06-27
Fit With Multiple Datasets

akleeman:
This introduces a new method to `ModelBase` which let's you call:

 `model.fit(dataset_x, dataset_y)`

where the two datasets don't necessarily need to be the same type.  If they are the same type then it's equivalent to fitting with a concatenation of the features and targets.  If they are different types a new dataset is created with type `variant<X, Y>`.

This should let us do things like fit to a bunch of training data and apply pseudo-observation constraints to any other type.

### DIFF
--- a/include/albatross/src/core/declarations.hpp
+++ b/include/albatross/src/core/declarations.hpp
@@ -13,6 +13,12 @@
 #ifndef ALBATROSS_CORE_DECLARATIONS_H
 #define ALBATROSS_CORE_DECLARATIONS_H
 
+namespace mapbox {
+namespace util {
+template <typename... Ts> class variant;
+}
+} // namespace mapbox
+
 using mapbox::util::variant;
 
 namespace Eigen {
--- a/include/albatross/src/core/distribution.hpp
+++ b/include/albatross/src/core/distribution.hpp
@@ -113,6 +113,25 @@ void set_subset(const Distribution<CovarianceType> &from,
   }
 }
 
+inline MarginalDistribution
+concatenate_marginals(const MarginalDistribution &x,
+                      const MarginalDistribution &y) {
+  Eigen::VectorXd mean(x.mean.size() + y.mean.size());
+  mean.block(0, 0, x.mean.size(), 1) = x.mean;
+  mean.block(x.mean.size(), 0, y.mean.size(), 1) = y.mean;
+
+  Eigen::VectorXd variance = Eigen::VectorXd::Zero(mean.size());
+  if (x.has_covariance()) {
+    variance.block(0, 0, x.mean.size(), 1) = x.covariance.diagonal();
+  }
+  if (y.has_covariance()) {
+    variance.block(x.mean.size(), 0, y.mean.size(), 1) =
+        y.covariance.diagonal();
+  }
+
+  return MarginalDistribution(mean, variance.asDiagonal());
+}
+
 } // namespace albatross
 
 #endif
--- a/tests/test_models.cc
+++ b/tests/test_models.cc
@@ -255,4 +255,54 @@ TEST(test_models, test_model_from_prediction_low_rank) {
       model_pred.covariance, 1e-8));
 }
 
+TEST(test_models, test_model_from_different_datasets) {
+  Eigen::Index k = 10;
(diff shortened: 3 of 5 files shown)


include/albatross/src/core/declarations.hpp
```diff
@@ -13,6 +13,12 @@
 #ifndef ALBATROSS_CORE_DECLARATIONS_H
 #define ALBATROSS_CORE_DECLARATIONS_H
 
+namespace mapbox {
+namespace util {
+template <typename... Ts> class variant;
```
seth-swiftnav: I'm not familiar with ellipsis in template names. What does this mean?

include/albatross/src/core/distribution.hpp
```diff
@@ -113,6 +113,25 @@ void set_subset(const Distribution<CovarianceType> &from,
...
 }
 
+inline MarginalDistribution
+concatenate_marginals(const MarginalDistribution &x,
+                      const MarginalDistribution &y) {
+  Eigen::VectorXd mean(x.mean.size() + y.mean.size());
+  mean.block(0, 0, x.mean.size(), 1) = x.mean;
+  mean.block(x.mean.size(), 0, y.mean.size(), 1) = y.mean;
+
+  Eigen::VectorXd variance = Eigen::VectorXd::Zero(mean.size());
```
seth-swiftnav: This logic looks weird to me. Do x and y have covariances if they are marginals? If only one of them has a covariance, do we get a concatenation with no uncertainty in some means but uncertainty in others? If neither has a covariance, is the concatenation a marginal with a diagonal of 0?

tests/test_models.cc
```diff
@@ -255,4 +255,54 @@ TEST(test_models, test_model_from_prediction_low_rank) {
...
+  }
+
+  ConstantEverywhere constant;
+  ConstantPerInterval per_interval;
+
+  // First we fit a model directly to the training data and use
+  // that to get a prediction of the inducing points.
+  auto model = gp_from_covariance(constant + per_interval, "unobservable");
+  const auto fit_model = model.fit(train_features, targets);
+  const auto inducing_points = create_inducing_points(train_features);
```
seth-swiftnav: This test really hinges on what this function does and whether `InducingFeature` is or is not a double. I cannot tell that from the test alone, but assume that it's testing the fit using two different datasets.

include/albatross/src/core/declarations.hpp
akleeman: It let's you provide an arbitrary number of template parameters.  Ie, `variant<int>` or `variant<int, double>` ...

Here's a good intro: https://eli.thegreenplace.net/2014/variadic-templates-in-c/

include/albatross/src/core/distribution.hpp
akleeman: yes, yes, yes :)

the `has_covariance` is sort of a poor name.  It's checking to make sure that an actual matrix has been allocated, in the case of a marginal distribution that "covariance" has zero off diagonals ... but it does exist.

tests/test_models.cc
akleeman: `InducingFeature` is defined earlier in the `.cc` file (but wasn't changed so you probably can't see it.).

```
enum InducingFeatureType { ConstantEverywhereType, ConstantPerIntervalType };

struct InducingFeature {
  InducingFeatureType type;
  long location;
};
enum InducingFeatureType { ConstantEverywhereType, ConstantPerIntervalType };

struct InducingFeature {
  InducingFeatureType type;
  long location;
};
```
And yeah, what it's testing is that we can fit a model with two datasets which are defined for two different types .

include/albatross/src/core/distribution.hpp
seth-swiftnav: Would it be better to assert that there is a "covariance"?

akleeman: That would prevent us from being able to concatenate two marginal distributions if one of them doesn't have a covariance defined which seems a bit restrictive to me.

In general I'd like to just gid rid of the "optional" covariance for `MarginalDistribution` but it is used in a few places (namely `fit`) to allow you do pass targets which are the mean only through a model happily.  I'd be excited to change that behavior ... but that's a larger (less urgent) task.

PR #143, opened 2019-07-08
Add BlockSymmetric

akleeman:
This adds an additional class to the existing block matrix utilities which allows you to invert a block symmetric matrix.

This should prove useful in a follow up PR in which I plan to add the ability to update a model which has already been fit.  Ie, consider the situation where you have a GP which has already been fit.  In that situation you have features `f` and a covariance matrix `A` for which the LDLT has already been computed.  This gives you:

```
 f] ~ N(0, A)
```
Now say you want to update the model to include new features `u`, you'd end up with something like:
```
 f]  ~ N(0, [A    B])
 u]         [B.T  C]
```
One way to do so would be to re-form the entire covariance matrix and compute the full LDLT, but this block matrix utility will let you re-use `A.ldlt()` to effeciently perform the resulting augmented solves required when using a GP to predict.

### DIFF
--- a/include/albatross/src/utils/block_utils.hpp
+++ b/include/albatross/src/utils/block_utils.hpp
@@ -52,6 +52,53 @@ struct BlockDiagonal {
   Eigen::MatrixXd toDense() const;
 };
 
+struct BlockSymmetric {
+
+  /*
+   * Stores a covariance matrix which takes the form:
+   *
+   *   X = |A   B|
+   *       |B.T C|
+   *
+   * It is assumes that both A and C - B.T A^-1 B are invertible and is
+   * designed for the situation where A is larger than C.  The primary
+   * use case is for a situation where you have a pre computed LDLT of
+   * a submatrix (A) and you'd like to perform a solve of the larger
+   * matrix (X)
+   */
+
+  BlockSymmetric(){};
+
+  BlockSymmetric(const Eigen::SerializableLDLT &A_, const Eigen::MatrixXd &B_,
+                 const Eigen::SerializableLDLT &S_)
+      : A(A_), Ai_B(A_.solve(B_)), S(S_) {}
+
+  BlockSymmetric(const Eigen::SerializableLDLT &A_, const Eigen::MatrixXd &B_,
+                 const Eigen::MatrixXd &C)
+      : BlockSymmetric(
+            A_, B_,
+            Eigen::SerializableLDLT(C - B_.transpose() * A_.solve(B_))){};
+
+  /*
+   * https://en.wikipedia.org/wiki/Block_matrix#Block_matrix_inversion
+   */
+  template <class _Scalar, int _Rows, int _Cols>
+  Eigen::Matrix<_Scalar, _Rows, _Cols>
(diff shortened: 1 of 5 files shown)


tests/test_block_utils.cc
```diff
@@ -107,4 +107,30 @@ TEST(test_block_utils, test_matrix_l) {
   EXPECT_LE((block_result - dense_result).norm(), 1e-6);
 }
 
+TEST(test_block_utils, test_block_symmetric) {
+
+  Eigen::Index k = 5;
```
seth-swiftnav: A lot of this setup is duplication of the struct `BlockSymmetricMatrix : public SerializableType<BlockSymmetric>`  made for the serialization test. Can it be reused here? Some sort of example/test_blocksymetric?

include/albatross/src/utils/block_utils.hpp
```diff
@@ -52,6 +52,53 @@ struct BlockDiagonal {
...
+   * It is assumes that both A and C - B.T A^-1 B are invertible and is
+   * designed for the situation where A is larger than C.  The primary
+   * use case is for a situation where you have a pre computed LDLT of
+   * a submatrix (A) and you'd like to perform a solve of the larger
+   * matrix (X)
+   */
+
+  BlockSymmetric(){};
+
+  BlockSymmetric(const Eigen::SerializableLDLT &A_, const Eigen::MatrixXd &B_,
```
seth-swiftnav: It's not clear to me based off the constructor arguments and comment what `S` is. It's only when looking at both the comment and the next constructor that this becomes clear. Should this constructor be private? It seems easy to use incorrectly.

tests/test_block_utils.cc
akleeman: Yeah I'll throw a method in `test_utils.h`

include/albatross/src/utils/block_utils.hpp
akleeman: This is actually the constructor we'll end up using: (https://github.com/swift-nav/albatross/pull/144/files#diff-c7c355d66e5a4b694f01b0f3aaf0ed9eR497) so I don't want to make it private but could make the other one private (though that'll mess with tests) and/or add docs?

akleeman:
@seth-swiftnav What other features do you have in mind?  What would they be used for?

seth-swiftnav:
> @seth-swiftnav What other features do you have in mind? What would they be used for?
> 
> EDIT: Ha! I mis-interpreted "features". I think I understand what you were asking about now, but I'm still not clear what it'd be used for?

I don't have anything in mind. It just seems like a natural extension.

akleeman:
@seth-swiftnav Yeah we could have this templated such that it can take any type that has a solve method, which means you could recursively build

Ie, right now it'd be:
```
BlockSymmetric<Eigen::SerializableLDLT>
```
but could end up:
```
BlockSymmetric<BlockSymmetric<Eigen::SerializableLDLT>>
```
etc ...

I suspect we're going to need that sort of functionality before long actually (but I'd rather not include it in this PR).

PR #144, opened 2019-07-08
Update Method for Gaussian Processes

akleeman:
This let's you fit a model:
```
partial_fit = model.fit(dataset);
```
Then subsequently update that fit:
```
updated_fit = update(partial_fit, other_dataset);
```
This is useful in the situation where you may not have all information available at a certain time, but you want to pre-compute the bulk of the model, then add data (or constraints) later.


include/albatross/src/models/gp.hpp
```diff
@@ -45,6 +45,8 @@ struct Fit<GPFit<CovarianceRepresentation, FeatureType>> {
   static_assert(has_solve<CovarianceRepresentation, Eigen::MatrixXd>::value,
                 "CovarianceRepresentation must have a solve method");
 
+  using Feature = FeatureType;
```
seth-swiftnav: Is this alias used anywhere? I don't see any.

akleeman: Yeah good catch, this isn't used in this change ... but I did find it very useful when debugging.  Sometimes you might call `model.fit(features)` but internally those features will get converted to a different type and it's nice to have some (easyish) visibility into which feature was used.

PR #146, opened 2019-07-11
Expand BlockSymmetric

akleeman:
Currently `BlockSymmetric` only works with an `Eigen::SerializableLDLT` type, but this templates the class to allow for anything with a `.solve()` method.


akleeman:
The only benchmarking that's happened has been me trying it with relatively large models and not noticing any horribly performance hit.  I'd like to hold off on supporting multiple updates for a subsequent PR.  As it stands only two variant types are supported, but I'll soon add the ability to support multiples at which point that test would be a great idea.

PR #147, opened 2019-07-16
Add support for variants with more than two types.

akleeman:
This change allows us to use covariance functions and concatenation routines with `variant` types with more than two types.  In the process I ended up moving a lot of code around into more appropriate headers and added a BUNCH more tests.  Changes include:

- The `Callers` (`DirectCaller`, `SymmetricCaller`, `MeasurementForwarder`, `VariantForwarder`) are all in a new file `callers.hpp`.
- The `Caller` classes are no longer templated on the covariance function in favor have an additional parameter in their `call` methods.
- Several new traits are introduced (`has_valid_variant_cov_caller`, `has_valid_caller_for_all_variants`, `is_variant`, `is_in_variant`) which facilitate the logic of forwarding covariance calls on to the proper `_call_impl`.
- `concatenate(feature_x, features_y)` methods were defined which work when `X` and `Y` are the same types, different types as well as when one or both of them are variants (in which case the variant types are possibly augmented by one).
- Added a lot more tests.


include/albatross/src/covariance_functions/traits.hpp
```diff
@@ -32,14 +32,41 @@ HAS_METHOD(_call_impl);
...
+};
+
+template <typename U, typename Caller, typename A, typename B>
+struct has_valid_cross_cov_caller<U, Caller, A, variant<B>>
+    : public has_valid_cross_cov_caller<U, Caller, A, B> {
+  ;
+};
+
+template <typename U, typename Caller, typename A, typename B, typename... Ts>
+struct has_valid_cross_cov_caller<U, Caller, A, variant<B, Ts...>> {
```
jbangelo: You'd know better than me, but it seems like it doesn't make sense to say if there's valid cross cov caller if not all of the variant types have a caller function.

akleeman: Yeah, great catch.  This forced me to revise some of the tests as well.  Thanks.

include/albatross/src/covariance_functions/measurement.hpp
```diff
@@ -24,6 +24,40 @@ template <typename X> struct Measurement {
...
+
+/*
+ * Takes a vector of features and turns it into a vector of
+ * measurements of those features.
+ */
+template <typename FeatureType>
+auto as_measurements(const std::vector<FeatureType> &features) {
+  std::vector<Measurement<FeatureType>> measurement_features;
+  for (const auto &f : features) {
+    measurement_features.emplace_back(Measurement<FeatureType>(f));
```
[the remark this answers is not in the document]
akleeman: Good call, changed it and everything still works.

PR #149, opened 2019-08-04
VariantCaller can use variants of differing types.

akleeman:
This changes the behavior of the `VariantCaller` to allow for two pairs of variants which have differing types.  This is allowed only when all types involved have a valid covariance method, but they need not have cross covariance.

For example, for a covriance function with the following mtehods:
```
Example : public CovarianceFunction<Example> {
  double _call_impl(const X&, const X&) const;
  double _call_impl(const X&, const Y&) const;
  double _call_impl(const Y&, const Y&) const;
  double _call_impl(const W&, const W&) const;
}
```
The following pairs of argument types would be acceptable:
```
X, X
X, Y
Y, Y
W, W
variant<X, Y>, variant<X, Y>
variant<X, Y>, variant<Y, X>
variant<X, Y>, variant<X, Y, W>
variant<X>, variant<X, Y, W>
variant<X, W, Y>, variant<W, X, Y>
...
```
The key point being that even though the `W` and `Y` cross covariance isn't defined variants containing those pairs will be, but will evaluate to `0`.

### DIFF
--- a/include/albatross/src/covariance_functions/callers.hpp
+++ b/include/albatross/src/covariance_functions/callers.hpp
@@ -217,14 +217,14 @@ template <typename SubCaller> struct VariantForwarder {
     return apply_visitor(visitor, x);
   }
 
-  // Deals with two identical variants.
-  template <typename CovFunc, typename... Ts,
+  // Deals with two variants.
+  template <typename CovFunc, typename... Xs, typename... Ys,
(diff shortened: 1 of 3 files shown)


seth-swiftnav:
lgtm! I'm not familiar with the variant syntax, but the description and test make sense.

PR #152, opened 2019-08-23
Log Transformation of Priors While Tuning

akleeman:
This adds a `LogUniformPrior` which lets you specify the bounds of a uniform prior which is subsequently explored in log space while tuning.

The nugget parameters in the sparse gaussian process were then reverted to non-log scale with the expectation that if a log scale is required (for tuning) the user can manually set the prior.

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -137,9 +137,11 @@ public:
 
   void initialize_params() {
     measurement_nugget_ = {details::DEFAULT_NUGGET,
-                           std::make_shared<NonNegativePrior>()};
+                           std::make_shared<LogScaleUniformPrior>(
+                               PARAMETER_EPSILON, PARAMETER_MAX)};
     inducing_nugget_ = {details::DEFAULT_NUGGET,
-                        std::make_shared<NonNegativePrior>()};
+                        std::make_shared<LogScaleUniformPrior>(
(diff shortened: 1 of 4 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -136,27 +136,27 @@ class SparseGaussianProcessRegression
   };
 
   void initialize_params() {
-    log_measurement_nugget_ = {log10(details::DEFAULT_NUGGET),
-                               std::make_shared<UninformativePrior>()};
-    log_inducing_nugget_ = {log10(details::DEFAULT_NUGGET),
-                            std::make_shared<UninformativePrior>()};
+    measurement_nugget_ = {details::DEFAULT_NUGGET,
```
seth-swiftnav: Should this be the new type,`LogScaleUniformPrior` ?

akleeman: Yeah, so I did that as first ... but then found it hard to assign an upper bound that would apply to all applications.  I suppose I could do something like `[1e-16, 1e16]` and that should be generally applicable (and if not users can always overrwrite the default prior.

akleeman: I've followed up, switching to `LogScaleUniformPrior` with the numerical limits of the double as bounds (ie, as close to zero as possible and as large as possible).

PR #155, opened 2019-08-31
Avoid an assert and add a ForwardDeclarations include

akleeman:
A few minor changes which:
- Avoid hard failures during tuning by removing an assert.  If an assert is still desired that *could* be manually enforced by first getting the tuning parameters, then hard enforcing the bounds externally.
- Allow setting a set of parameters which don't neccesarily overlap.  Putting this inside the `ParameterHandlingMixin` instead of in user code makes it a bit more efficient since there's access to `unchecked_set_param`.
- Add an include target for `ForwardDeclarations`, `CsvUtils` and `MapUtils`

### DIFF
--- a/include/albatross/Common
+++ b/include/albatross/Common
@@ -23,9 +23,7 @@
 
 #include <math.h>
 #include <iomanip>
-#include <map>
 #include <set>
-#include <string>
 #include <iostream>
 #include <unordered_map>
 #include <functional>
@@ -34,7 +32,7 @@
 #include <numeric>
 #include <random>
 
-#include <albatross/src/core/declarations.hpp>
+#include "ForwardDeclarations"
 
 #include <albatross/src/details/traits.hpp>
 #include <albatross/src/details/has_any_macros.hpp>
--- a/include/albatross/src/core/parameter_handling_mixin.hpp
+++ b/include/albatross/src/core/parameter_handling_mixin.hpp
@@ -142,10 +142,29 @@ public:
     }
   }
 
+  void set_params_if_exists(const ParameterStore &params) {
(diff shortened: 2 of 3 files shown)


include/albatross/Common
```diff
@@ -34,7 +32,7 @@
 #include <numeric>
 #include <random>
 
-#include <albatross/src/core/declarations.hpp>
+#include "ForwardDeclarations"
```
benjamin0: no filetype?

include/albatross/src/core/parameter_handling_mixin.hpp
```diff
@@ -142,13 +142,32 @@ class ParameterHandlingMixin {
     }
   }
 
+  void set_params_if_exists(const ParameterStore &params) {
```
seth-swiftnav: What's the use case for these functions? When would the ParameterStore not have the exact same parameters as the model being set?

include/albatross/Common
akleeman: Nope, `albatross` uses a similar include pattern to `Eigen`.  Files like `Eigen/Core` include all the headers you'd typically need.  I think this is something that comes out of the use of CRTP which can end up requiring a specific include order, so it's nice to have that explicitly done in one place.

include/albatross/src/core/parameter_handling_mixin.hpp
akleeman: I end up using this where I have one model with a given covariance function and another with a subset covariance function and I want the smaller model to take on the parameters of the parent.  I don't see this is a very common pattern, but it beats re-writting the `if(contains){set}` pattern in multiple places.

PR #157, opened 2019-09-06
Null Model

akleeman:
This change adds a `NullModel` which is a lot like it sounds.  You can treat it like any other model, ie `fit`, `predict` etc ... but it will always just return zero predictions which can be helpful as a benchmark model to make sure another model is actually doing something useful.

### DIFF
--- a/include/albatross/NullModel
+++ b/include/albatross/NullModel
@@ -1,0 +1,20 @@
+/*
+ * Copyright (C) 2019 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef ALBATROSS_NULL_MODEL_H
+#define ALBATROSS_NULL_MODEL_H
+
+#include "Core"
+
+#include <albatross/src/models/null_model.hpp>
+
+#endif
--- a/include/albatross/src/models/null_model.hpp
+++ b/include/albatross/src/models/null_model.hpp
@@ -64,9 +64,9 @@ public:
   _predict_impl(const std::vector<FeatureType> &features,
                 const Fit<NullModel> &fit,
                 PredictTypeIdentity<JointDistribution> &&) const {
-    const Eigen::Index en = static_cast<Eigen::Index>(features.size());
-    const Eigen::VectorXd mean = Eigen::VectorXd::Zero(en);
-    const Eigen::MatrixXd cov = 1.e4 * Eigen::MatrixXd::Identity(en, en);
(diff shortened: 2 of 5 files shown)


include/albatross/NullModel
```diff
@@ -0,0 +1,20 @@
...
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef ALBATROSS_NULL_MODEL_H
+#define ALBATROSS_NULL_MODEL_H
```
pgrgich: This doesn't have mark up, probably because the file doesn't have an extension - it should be a .h, right?

include/albatross/src/models/null_model.hpp
```diff
@@ -0,0 +1,91 @@
...
+    FitModel<NullModel, Fit<NullModel>> fit_model(m, Fit<NullModel>());
+    return fit_model;
+  }
+
+  template <typename FeatureType>
+  JointDistribution
+  _predict_impl(const std::vector<FeatureType> &features,
+                const Fit<NullModel> &fit,
+                PredictTypeIdentity<JointDistribution> &&) const {
+    const Eigen::Index en = static_cast<Eigen::Index>(features.size());
```
pgrgich: `en`?

tests/test_cross_validation.cc
```diff
@@ -42,7 +42,7 @@ TEST(test_cross_validation, test_fold_creation) {
 
 bool is_monotonic_increasing(const Eigen::VectorXd &x) {
   for (Eigen::Index i = 0; i < x.size() - 1; i++) {
-    if (x[i + 1] - x[i] <= 0.) {
+    if (x[i + 1] - x[i] < 0.) {
```
pgrgich: The definition of monotonic increasing is just that it doesn't decrease, right? Perhaps rename it if we want this change.

include/albatross/NullModel
akleeman: Actually, `albatross` has been using the `Eigen` style of includes, ie individual header files don't include anything and then a centralized include file (`#include <albatross/NullModel>`) deals with knowing what needs to be included and in what order.

include/albatross/src/models/null_model.hpp
akleeman: bad abbreviation on my part, `n` would be more clear huh.

tests/test_cross_validation.cc
akleeman: I think this actually matches the definition now.  Before it was checking that it "always increases" but as you pointed out monotonic increasing just means "doesn't decrease" which is what it now checks.

pgrgich: Oh, my mistake. :)

include/albatross/src/models/null_model.hpp
pgrgich: Probably, though I'm still rather in favor of something more descriptive like `num_features`. If `albatross` convention isn't to do that, we should make consistency the first priority.

include/albatross/NullModel
akleeman: The rest of the centralized includes are actually extension free as well.  This has confused other people too, so could be worth switching, but does match `Eigen`, (`#include <Eigen/Core>` and `#include <albatross/Core>`).

PR #158, opened 2019-09-10
Nearest Neighbor Model

akleeman:


include/albatross/src/evaluation/cross_validation_utils.hpp
```diff
@@ -93,15 +93,25 @@ inline MarginalDistribution concatenate_marginal_predictions(
   Eigen::VectorXd variance(n);
   Eigen::Index number_filled = 0;
   // Put all the predicted means back in order.
+  bool has_covariance = false;
   for (const auto &pair : indexer) {
     assert(preds.at(pair.first).size() == pair.second.size());
     set_subset(preds.at(pair.first).mean, pair.second, &mean);
-    set_subset(preds.at(pair.first).covariance.diagonal(), pair.second,
-               &variance);
+    if (preds.at(pair.first).has_covariance()) {
```
pgrgich: Do we want to do this or just add `1e6` as variances where we currently have none?

include/albatross/src/models/nearest_neighbor.hpp
```diff
@@ -0,0 +1,115 @@
...
+    return Fit<NearestNeighborFit<FeatureType>>(
+        RegressionDataset<FeatureType>(features, targets));
+  }
+
+  template <typename FeatureType>
+  auto fit_from_prediction(const std::vector<FeatureType> &features,
+                           const JointDistribution &prediction) const {
+    const NearestNeighborModel<DistanceMetric> m(*this);
+    MarginalDistribution marginal_pred(
+        prediction.mean, prediction.covariance.diagonal().asDiagonal());
```
pgrgich: Wouldn't `prediction.covariance` work here? Or are you looking to zero the non-diagonal elements?

include/albatross/src/models/nearest_neighbor.hpp
```diff
@@ -0,0 +1,115 @@
...
+  template <typename FeatureType>
+  std::size_t
+  index_with_min_distance(const FeatureType &ref,
+                          const std::vector<FeatureType> &features) const {
+    assert(features.size() > 0);
+
+    std::size_t min_index = 0;
+    double min_distance = distance_metric(ref, features[0]);
+
+    for (std::size_t i = 1; i < features.size(); ++i) {
```
pgrgich: Could we turn `min_distance` into an `optional` so the loop can start at 0? The only difference in the loop would be `!min_distance &&` going at the start of the `if`.

include/albatross/src/evaluation/cross_validation_utils.hpp
akleeman: In this method we're taking a pair of `MarginalDistribution` and concatenating them, so if the both don't have a defined covariance then we want to preserve that in the concatenation.  Somewhere on my list of want to dos is to remove the optional behavior for convariances in favor of a third distribution type, something like:
```
using MeanOnlyDistribution = Distribution<Empty>;
using MarginalDistribution = Distribution<DiagonalMatrixXd>;
using JointDistribution = Distribution<Eigen::MatrixXd>;
```
or something along those lines, but that's out of scope here.

include/albatross/src/models/nearest_neighbor.hpp
akleeman: Yeah exactly, I need to zero the non-diagonal elements since the `NearestNeighbor` model can never actually predict off diagonals.

include/albatross/src/models/nearest_neighbor.hpp
akleeman: I like that pattern better too ... but so far `albatross` doesn't used any optionals!  So we'd have to add a third party lib for it which I've been avoiding (though perhaps the time has come).

pgrgich: An alternative is to initialize the `min_distance` to `DBL_MAX` or some such, so it will always be replaced by the first distance.

PR #159, opened 2019-09-12
Remove const strings.

akleeman:
These `const std::string` lines caused multiple declaration errors, switched to functions since the more standard `extern` pattern doesn't work well for header only libraries.

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -16,9 +16,13 @@
 namespace albatross {
 
 namespace details {
+
 constexpr double DEFAULT_NUGGET = 1e-12;
-const std::string measurement_nugget_name = "measurement_nugget";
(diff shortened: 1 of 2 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -16,9 +16,13 @@
 namespace albatross {
 
 namespace details {
+
 constexpr double DEFAULT_NUGGET = 1e-12;
-const std::string measurement_nugget_name = "measurement_nugget";
-const std::string inducing_nugget_name = "inducing_nugget";
+
+inline std::string measurement_nugget_name() { return "measurement_nugget"; }
```
jbangelo: Potential small optimization for later would be to declare the string as a `static const` variable inside the function and return a `const &` to it, to avoid calling the `std::string` constructor each time the function is called. No need for that now, or probably ever since it's such a small optimization.

PR #160, opened 2019-09-26
GroupBy: Split Apply Combine

akleeman:
A pretty common pattern that has emerged when using `albatross` is the desire to split a dataset into groups, and inspect or modify on a per group basis, then occasionally recombine the results into a new dataset.  This PR adds a `group_by` method to dataset which facilitates these sorts of operations.

For example, what used to look like this:
```
const LeaveOneGroupOut<FeatureType> get_groups(get_group_name);
const auto by_group = by_groups(dataset.features);
for (const auto &pair : by_group) {
  const std::string group_name = pair.first;
  const std::vector<std::size_t> group_indexer = pair.second;
  const auto group_dataset = albatross::subset(dataset, group_indexer);
  do_something(group_dataset);
}
```
Can now be done like this:
```
dataset.group_by(get_group_name).apply(do_something);
```
Similarly you do sort-of-kind-of pandas style split-apply-combine operations:
```
const auto remove_bad_things = [](const auto &group_name, const auto &dataset) {
  return remove(dataset);
}
const auto clean_dataset = dataset.group_by(&get_group_name).apply(remove_bad_things).combine();
```
I stopped short of completely flushing out all desired functionality in the interest of getting some feedback (and keeping an already large PR from growing larger).  A lot of the trait logic (for example) is going to get an overhaul.  And for the moment this PR doesn't actually use the group by logic anywhere (but I have a follow up PR which does).

Currently supported operations:
```
auto groups = dataset.group_by(get_group).groups();
auto same_dataset = groups.combine();
auto modified_groups = dataset.group_by(get_group).apply(do_something_to_each_group);
// same for vectors
auto groups = group_by(vector, get_group).groups();
etc ...
// some simple operations over the groups
auto map_containing_size_of_each_group = dataset.group_by(get_group).counts();
// ability to remove groups (using filter).
auto dataset_without_outliers = dataset.group_by(get_group).filter(is_ok).combine();
```

### DIFF
--- a/include/albatross/src/core/distribution.hpp
+++ b/include/albatross/src/core/distribution.hpp
@@ -117,13 +117,52 @@ concatenate_marginals(const MarginalDistribution &x,
   if (!x.has_covariance() && !y.has_covariance()) {
     return MarginalDistribution(mean);
   } else {
-    Eigen::VectorXd variance = Eigen::VectorXd::Zero(mean.size());
+    Eigen::VectorXd variance(mean.size());
+    variance.fill(NAN);
     if (x.has_covariance()) {
       variance.block(0, 0, x.mean.size(), 1) = x.covariance.diagonal();
     }
     if (y.has_covariance()) {
       variance.block(x.mean.size(), 0, y.mean.size(), 1) =
           y.covariance.diagonal();
+    }
+
+    return MarginalDistribution(mean, variance.asDiagonal());
+  }
+}
+
+inline MarginalDistribution
+concatenate_marginals(const std::vector<MarginalDistribution> &dists) {
+
+  if (dists.size() == 0) {
+    return MarginalDistribution();
+  }
+
+  std::size_t size = 0;
+  bool any_has_covariance = false;
+  for (const auto &dist : dists) {
+    size += dist.size();
+    any_has_covariance = any_has_covariance || dist.has_covariance();
+  }
+
+  Eigen::VectorXd mean(static_cast<Eigen::Index>(size));
+  Eigen::Index i = 0;
+  for (const auto &dist : dists) {
+    mean.block(i, 0, dist.mean.size(), 1) = dist.mean;
+    i += dist.mean.size();
+  }
+
+  if (!any_has_covariance) {
+    return MarginalDistribution(mean);
+  } else {
+    Eigen::VectorXd variance(mean.size());
+    variance.fill(NAN);
+    Eigen::Index i = 0;
+    for (const auto &dist : dists) {
+      if (dist.has_covariance()) {
+        variance.block(i, 0, dist.mean.size(), 1) = dist.covariance.diagonal();
+      }
+      i += dist.mean.size();
     }
 
     return MarginalDistribution(mean, variance.asDiagonal());
--- a/include/albatross/src/core/distribution.hpp
+++ b/include/albatross/src/core/distribution.hpp
@@ -148,7 +148,7 @@ concatenate_marginals(const std::vector<MarginalDistribution> &dists) {
   Eigen::VectorXd mean(static_cast<Eigen::Index>(size));
   Eigen::Index i = 0;
   for (const auto &dist : dists) {
-    mean.block(i, 0, dist.mean.size(), 1) = dist.mean;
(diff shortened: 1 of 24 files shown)


include/albatross/src/indexing/group_by.hpp
```diff
@@ -0,0 +1,443 @@
...
+
+  template <typename ApplyFunction,
+            typename ApplyType = typename details::key_value_apply_return_type<
+                ApplyFunction, KeyType, GroupIndices>::type,
+            typename std::enable_if<
+                details::is_valid_index_apply_function<ApplyFunction, KeyType,
+                                                       GroupIndices>::value &&
+                    std::is_same<void, ApplyType>::value,
+                int>::type = 0>
+  auto index_apply(const ApplyFunction &f) const {
```
jbangelo: This should be a `void` return type?

include/albatross/src/indexing/traits.hpp
```diff
@@ -0,0 +1,195 @@
...
+    : public can_be_called_with_const_ref<
+          ApplyFunction, typename const_ref<KeyType>::type, ArgType> {};
+
+template <typename ApplyFunction, typename ArgType>
+struct is_valid_value_only_apply_function
+    : public can_be_called_with_const_ref<ApplyFunction, ArgType> {};
+
+template <typename ApplyFunction, typename KeyType, typename ArgType>
+struct is_valid_index_apply_function
+    : public can_be_called_with_const_ref<
```
jbangelo: Any reason why you jump between inheriting from other traits and explicitly defining the `value`/`type` members? They should be equivalent, unless there's additional members you don't want to public inherit, but I'm not seeing any cases of that.

include/albatross/src/indexing/traits.hpp
```diff
@@ -0,0 +1,195 @@
...
+/*
+ * The following traits are required in order to allow inspection of
+ * the only partially defined Derived types inside of GroupByBase.
+ *
+ * To get GroupByBase to work with other types you need to add a new
+ * trait struct for the type.
+ *
+ * GrouperFunction is a callable type which should take
+ */
+template <typename T> struct traits {};
```
jbangelo: Would it make sense to call this something like `group_by_traits`?

include/albatross/src/core/distribution.hpp
```diff
@@ -130,6 +131,44 @@ concatenate_marginals(const MarginalDistribution &x,
...
+  bool any_has_covariance = false;
+  for (const auto &dist : dists) {
+    size += dist.size();
+    any_has_covariance = any_has_covariance || dist.has_covariance();
+  }
+
+  Eigen::VectorXd mean(static_cast<Eigen::Index>(size));
+  Eigen::Index i = 0;
+  for (const auto &dist : dists) {
+    mean.block(i, 0, dist.mean.size(), 1) = dist.mean;
```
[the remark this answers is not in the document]
akleeman: Oh cool, I didn't know that existed.

include/albatross/src/indexing/group_by.hpp
```diff
@@ -0,0 +1,443 @@
...
+ * The Grouped class is the output of a call to method such as
+ *
+ *   dataset.groupby(grouper).groups()
+ *
+ * which is basically just a map with additional functionality
+ * that facilitates applying a function to all values, filtering etc ...
+ */
+
+template <typename KeyType, typename ValueType>
+class GroupedBase : public std::map<KeyType, ValueType> {
```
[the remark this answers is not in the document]
akleeman: Ah interesting, I was a bit worried about that.  I think I most use the get/set methods and iterators so it shouldn't be *too* bad to add those all here.

include/albatross/src/indexing/traits.hpp
akleeman: yeah, good call.

include/albatross/src/indexing/traits.hpp
akleeman: No specific reason ... I do plan on overhauling the traits in a follow up PR so I'll leave this for now, then aim for consistency in the follow up.

PR #164, opened 2019-10-04
Revert the measurement change which wasn't actually tested.

akleeman:
Turns out the `Measurement<T>` flags are used in the `SparseGaussianProcess` and are actually quite important there.  Basically, the `SparseGaussianProcess` may be using the same feature type for both the input features and the inducing points.  This is the case in some of the tests where the data points are all `double` and the inducing points are also a uniform grid of `double`.  In this situation we want the covariance function to evaluate the covariance between inputs as being measurements, but do not want the measurement noise when evaluating the covariance between inducing points.

As a result I've put the `as_measurement` step back into `_fit_impl`.  Also, I (re)discovered that the target variance *is* being added inside the `Fit` constructor.

This was all caught after realizing PR #161 actually reduced the tests to only the csv tests :grimacing: .  Overall, not a good streak in the last two PRs!  Ooops!

### DIFF
--- a/include/albatross/src/models/gp.hpp
+++ b/include/albatross/src/models/gp.hpp
@@ -242,10 +242,8 @@ public:
   CholeskyFit<FeatureType>
   _fit_impl(const std::vector<FeatureType> &features,
             const MarginalDistribution &targets) const {
-    Eigen::MatrixXd cov = covariance_function_(features);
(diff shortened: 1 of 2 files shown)


akleeman:
@seth-swiftnav yes, never a stupid question in a review!

Issue #166, opened 2019-10-07
Measurement Noise Edge Case


akleeman:
There is an unfortunate edge case which occurs when using a templated `_call_impl` in a `CovarianceFunction` along with the `Measurement` wrapper used in Gaussian process fits.  For example, consider the following covariance function:

```
struct T {};
struct Foo : public CovarianceFunction<Foo> {
  template <typename X, typename Y>
  double _call_impl(const X&, const Y&) {
    return 0.;
  }

  double _call_impl(const T &, const T &) {
    return 1.;
  }
}
```
When called directly you would expect the following behavior:
```
Foo foo;
T t;
int x;
// 
foo(t, t); // 1.
foo(x, t); // 0.
foo(x, x) // 0.
```
Basically, you would expect that a call with types `T` would get caught by the specialization of `_call_impl` while everything else would get caught by the templated catch all.  This *is* true, but the confusing part comes when used in a Gaussian process.  Consider this example:
```
const auto gp = gp_from_covariance(foo);
RegressionDataset<T> some_dataset = get_dataset();
const auto fit_model = gp.fit(some_dataset);
```
If you inspected the resulting covariance matrix `fit_model.get_fit().train_covariance;
` you'd find that it consists entirely of zeros!!  What happens is that the call to `.fit()` will wrap the input with a `Measurement` wrapper before evaluating the covariance:
```
  template <typename FeatureType,
            typename std::enable_if<
                has_call_operator<CovFunc, FeatureType, FeatureType>::value,
                int>::type = 0>
  CholeskyFit<FeatureType>
  _fit_impl(const std::vector<FeatureType> &features,
            const MarginalDistribution &targets) const {
    const auto measurement_features = as_measurements(features);
    Eigen::MatrixXd cov = covariance_function_(measurement_features);
    return CholeskyFit<FeatureType>(features, cov, targets);
  }
```
so `foo` will actually end up being called with types `Measurement<T>` not `T` explicitly.  Without the catch all `_call_impl` in `Foo` it would actually behave as expected because the covariance function is first be inspected for a valid call with `Measurement<T>` and if one is not found the `Measurement<>` wrapper is stripped of and the covariance is called wit the underlying type.

To fix this behavior you could add another method to `Foo` that looked like:
```
  template <typename X, typename Y>
  double _call_impl(const Measurement<X>&, const Measurement<Y>&) = delete;
```
[...]

PR #167, opened 2019-10-07
Macro and Call Trace improvements.

akleeman:
These changes started from an attempt to fix https://github.com/swift-nav/albatross/issues/166 by inspecting covariance functions for specialized implementations specific to `Measurement` types, but that attempt failed.

In the process however I ended up with these two commits which I think are worth getting in.
- The first is a change to the macros which define the `has_any_*` and `has_any_*_with_return_type` traits which consolidated the trickiest part of the logic into one place.
- The second fixes a bug in the `CallTrace` method in which it was possible for the call trace that got generated to differ from the actual covariance evaluation.

### DIFF
--- a/include/albatross/src/details/has_any_macros.hpp
+++ b/include/albatross/src/details/has_any_macros.hpp
@@ -22,7 +22,7 @@ template <typename First, typename Second> struct TypePair {
 
 /*
  * Defines a couple of traits which help inspect whether a type has a method
- * which is callable whne provided with specific arguments.
+ * which is callable when provided with specific arguments.
  *
  *   template <typename T, typename ... Args>
  *   struct class_method_FNAME_traits {
--- a/include/albatross/src/details/has_any_macros.hpp
+++ b/include/albatross/src/details/has_any_macros.hpp
@@ -15,29 +15,70 @@
 
 namespace albatross {
 
-#define HAS_METHOD(fname)                                                      \
-  template <typename T, typename... Args> class has_##fname {                  \
-    template <typename C, typename = decltype(std::declval<C>().fname(         \
-                              std::declval<Args>()...))>                       \
-    static std::true_type test(C *);                                           \
-    template <typename> static std::false_type test(...);                      \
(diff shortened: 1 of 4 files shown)


include/albatross/src/details/has_any_macros.hpp
```diff
@@ -15,29 +15,70 @@
...
+  using first_type = First;
+  using second_type = Second;
+};
 
-#define HAS_METHOD_WITH_RETURN_TYPE(fname)                                     \
-  template <typename T, typename ReturnType, typename... Args>                 \
-  class has_##fname##_with_return_type {                                       \
+/*
+ * Defines a couple of traits which help inspect whether a type has a method
+ * which is callable whne provided with specific arguments.
```
jbangelo:
```suggestion
 * which is callable when provided with specific arguments.
```

include/albatross/src/details/has_any_macros.hpp
```diff
@@ -15,29 +15,70 @@
...
-  template <typename T, typename... Args> class has_##fname {                  \
-    template <typename C, typename = decltype(std::declval<C>().fname(         \
-                              std::declval<Args>()...))>                       \
-    static std::true_type test(C *);                                           \
-    template <typename> static std::false_type test(...);                      \
-                                                                               \
-  public:                                                                      \
-    static constexpr bool value = decltype(test<T>(0))::value;                 \
-  };
+template <typename First, typename Second> struct TypePair {
```
jbangelo: I think you could use `std::pair` as well. It should have `first_type` and `second_type` member types defined.

https://en.cppreference.com/w/cpp/utility/pair

akleeman: Yeah I can *almost* use it, but I found that it fails when you pass a `void` type as one of the types in a pair because the pair actually tries storing all the types.

jbangelo: Oh! I guess that makes sense.

PR #168, opened 2019-10-07
Eigen Value Debug Utils

akleeman:
Added `linalg_utils.cc` which includes a couple methods which make it easy to print small / large eigen values along with the features that correspond.  This has proven useful when debugging situations that are either unobservable or overly constrained.

To illustrate how this can be useful here's an example in which we build a covariance matrix which represents an ill conditioned problem. Staring with two independent variables of different scales:
```
u ~ N(0, 10)
v ~ N(0, 0.01)
```
We can then build a new system which will contain two very highly correlated variables
```
x = u + v
y = u - v
```
This sets up that situation:
```
Eigen::Index k = 2;
// This represents the variance of u and v 
Eigen::VectorXd diag(k);
diag << 100., 1e-4;
// permute translates from u, v to x, y
Eigen::MatrixXd permute(k, k);
permute.row(0) << 1., 1.;
permute.row(1) << 1., -1.;
// Build the new system
Eigen::MatrixXd covariance = permute * diag.asDiagonal();
covariance = covariance * permute.transpose();
// Use some simple features
std::vector<int> features;
for (Eigen::Index i = 0; i < k; ++i) {
  features.push_back(i);
}
```
After which a call to: `print_small_eigen_directions(covariance, features, k)` would result in:
```
MIN: 0.0002
MAX: 200
eigen value: 0.0002
       -0.707107   0
        0.707107   1
eigen value: 200
        0.707107   0
        0.707107   1
```
From which we can see `x - y` is very observable (variance of `0.0002`) while `x + y` is extremely noisy (variance of `200`).

### DIFF
--- a/include/albatross/src/utils/linalg_utils.hpp
+++ b/include/albatross/src/utils/linalg_utils.hpp
@@ -17,10 +17,13 @@ namespace albatross {
 
 namespace details {
 
+constexpr double DEFAULT_EIGEN_VALUE_PRINT_THRESHOLD = 1e-3;
+
 template <typename FeatureType, typename Comparator>
 inline void _print_eigen_directions(const Eigen::MatrixXd &matrix,
                                     const std::vector<FeatureType> &features,
                                     std::size_t count, Comparator comp,
+                                    double print_if_above,
                                     std::ostream *stream) {
   Eigen::SelfAdjointEigenSolver<Eigen::MatrixXd> ev(matrix);
   const Eigen::VectorXd values = ev.eigenvalues().real();
@@ -47,7 +50,7 @@ inline void _print_eigen_directions(const Eigen::MatrixXd &matrix,
     (*stream) << "eigen value: " << value << std::endl;
     for (Eigen::Index j = 0; j < vector.size(); ++j) {
       double coef = vector[j];
-      if (fabs(coef) > 1e-3) {
+      if (fabs(coef) > print_if_above) {
         (*stream) << "    " << std::setw(12) << coef << "   " << features[j]
                   << std::endl;
       }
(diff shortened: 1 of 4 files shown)


include/albatross/src/utils/linalg_utils.hpp
```diff
@@ -0,0 +1,84 @@
...
+    values_and_vectors.emplace_back(values[i], vectors.col(i));
+  }
+
+  std::sort(values_and_vectors.begin(), values_and_vectors.end(),
+            [&](const ValueAndVector &a, const ValueAndVector &b) -> bool {
+              return comp(std::get<0>(a), std::get<0>(b));
+            });
+
+  (*stream) << "MIN: " << values.minCoeff() << std::endl;
+  (*stream) << "MAX: " << values.maxCoeff() << std::endl;
```
seth-swiftnav: Is there a reason to call these methods when values_and_vectors is already sorted?

include/albatross/src/utils/linalg_utils.hpp
```diff
@@ -0,0 +1,84 @@
...
+  for (std::size_t i = 0; i < count; ++i) {
+    const double value = std::get<0>(values_and_vectors[i]);
+    const auto vector = std::get<1>(values_and_vectors[i]);
+    (*stream) << "eigen value: " << value << std::endl;
+    for (Eigen::Index j = 0; j < vector.size(); ++j) {
+      double coef = vector[j];
+      if (fabs(coef) > 1e-3) {
+        (*stream) << "    " << std::setw(12) << coef << "   " << features[j]
+                  << std::endl;
+      }
```
seth-swiftnav: Can you explain the `1e-3` cutoff?
Do you have an example of what this looks like when printed? It looks like it would print  each feature a number of times equal to the height of the matrix.

If an element of an eigenvector is small, print the feature that corresponds to that dimension. Can you explain how this could be used to see unobservable and over constrained features alike?

akleeman: Yeah, 1e-3 was a pretty arbitrary pick, but mostly just a way of removing smaller influences.  I'll change it to an argument.

include/albatross/src/utils/linalg_utils.hpp
akleeman: Yeah, the sort order can change from ascending to descending so this makes sure min corresponds to min.

include/albatross/src/utils/linalg_utils.hpp
akleeman: There's a test case which when printed looks like this:
```
MIN: 0.0344498
MAX: 8.74038
eigen value: 0.0344498
        0.174948   0
       -0.690045   1
       0.0589917   2
       -0.314007   3
      -0.0752929   4
      -0.0982879   5
         0.24261   6
         0.45257   7
       -0.254815   8
        -0.21728   9
eigen value: 0.292064
        0.332327   0
       0.0426091   1
       -0.267221   2
...
```
Not a super informative example, but shows the printing format, in this case the features are `int`.  I'll include another example in the description here for future reference.

seth-swiftnav: That was an excellent description. Thank you.

PR #169, opened 2019-10-07
Chi Squared CDF

akleeman:
Adds a function to compute the CDF of the chi_squared distribution which can be used in multi-variate normality tests.


include/albatross/src/stats/chi_squared.hpp
```diff
@@ -0,0 +1,62 @@
...
+  if (std::isinf(x)) {
+    return 1.;
+  }
+
+  return chi_squared_cdf_unsafe(x, degrees_of_freedom);
+}
+
+} // namespace details
+
+inline double chi_squared_cdf(double x, std::size_t degrees_of_freedom) {
```
seth-swiftnav: assert x and degrees_of_freedom non-negative?

tests/test_stats.cc
```diff
@@ -0,0 +1,44 @@
...
+namespace albatross {
+
+TEST(test_stats, test_chi_squared) {
+
+  EXPECT_LT(fabs(chi_squared_cdf(9.260, 23) - 0.005), 1e-5);
+  EXPECT_LT(fabs(chi_squared_cdf(38.932, 21) - 0.99), 1e-5);
+  EXPECT_LT(fabs(chi_squared_cdf(96.578, 80) - 0.9), 1e-5);
+  EXPECT_LT(fabs(chi_squared_cdf(70.065, 100) - 0.01), 1e-5);
+
+  EXPECT_LT(chi_squared_cdf(0., 0.), 1e-6);
```
seth-swiftnav: This seems ill defined to me. The pdf of chi2(0) is a delta dirac at 0. This makes all positive numbers greater then the anything sampled from chi2(0), but it makes 0 equal to anything sampled from chi2(0). I would've said this should be 1.

include/albatross/src/stats/incomplete_gamma.hpp
```diff
@@ -0,0 +1,178 @@
...
+               lg_term) *
+           incomplete_gamma_quadrature_weight_vals(lb, ub, counter);
+  }
+}
+
+inline double incomplete_gamma_quadrature_lb(double a, double z) {
+  if (a > 1000) {
+    return std::max(0., std::min(z, a) - 11 * sqrt(a));
+  } else if (a > 800) {
+    return std::max(0., std::min(z, a) - 11 * sqrt(a));
```
seth-swiftnav: There seem to be several instances of conditional with the same execution.

include/albatross/src/stats/incomplete_gamma.hpp
```diff
@@ -0,0 +1,178 @@
...
+  } else if (a > 40) {
+    return std::min(z, a + 5 * sqrt(a));
+  }
+  return std::min(z, a + 4 * sqrt(a));
+}
+
+double incomplete_gamma_quadrature(double a, double z) {
+  return incomplete_gamma_quadrature_recursive(
+      incomplete_gamma_quadrature_lb(a, z),
+      incomplete_gamma_quadrature_ub(a, z), a, lgamma(a), 0);
```
seth-swiftnav: The upper and lower bounds quadrature functions seem heavily related. I might consider merging them into a single function that returns a tuple/pair/struct. I think that'll make the conditionals more intuitive if they are zippered together.

tests/test_stats.cc
akleeman: nice catch

include/albatross/src/stats/chi_squared.hpp
akleeman: going to have it return NAN in that case to avoid putting an assert in the functions.

include/albatross/src/stats/incomplete_gamma.hpp
akleeman: Yeah, kinda weird.  Not sure how those got selected, I cleaned it up a bit.

include/albatross/src/stats/incomplete_gamma.hpp
```diff
@@ -67,58 +67,46 @@ inline double incomplete_gamma_quadrature_recursive(double lb, double ub,
...
-    return std::min(z, a + 10 * sqrt(a));
-  } else if (a > 800) {
-    return std::min(z, a + 10 * sqrt(a));
+  if (a > 800) {
+    return std::make_pair(std::max(0., std::min(z, a) - 11 * sqrt(a)),
+                          std::min(z, a + 10 * sqrt(a)));
   } else if (a > 500) {
-    return std::min(z, a + 9 * sqrt(a));
+    return std::make_pair(std::max(0., std::min(z, a) - 10 * sqrt(a)),
+                          std::min(z, a + 9 * sqrt(a)));
```
seth-swiftnav: Is this necessary if the next conditional will cover this case with the same result?
Is this the desired behavior or is there a typo in the bounds.

akleeman: Nah I just missed it.  all this is taken from https://github.com/kthohr/gcem/blob/master/include/gcem_incl/incomplete_gamma.hpp#L76-L113

include/albatross/src/stats/incomplete_gamma.hpp
```diff
@@ -67,58 +67,43 @@ inline double incomplete_gamma_quadrature_recursive(double lb, double ub,
...
+    return std::make_pair(std::max(0., std::min(z, a) - 11 * sqrt(a)),
+                          std::min(z, a + 10 * sqrt(a)));
   } else if (a > 300) {
-    std::min(z, a + 9 * sqrt(a));
+    return std::make_pair(std::max(0., std::min(z, a) - 10 * sqrt(a)),
+                          std::min(z, a + 9 * sqrt(a)));
   } else if (a > 100) {
-    return std::min(z, a + 8 * sqrt(a));
+    return std::make_pair(std::max(0., std::min(z, a) - 9 * sqrt(a)),
+                          std::min(z, a + 8 * sqrt(a)));
```
seth-swiftnav: One more duplicate conditional. :)

akleeman:
Ugh, I just put a bunch of tests together which were testing that the cdf of a bunch of random samples was approximately uniform and it's exposed an issue with small `a` in the incomplete gamma function.  Going to hold off merging this until that get's fixed, standby.

PR #170, opened 2019-10-09
Multivariate Gaussian Chi Squared Statistics

akleeman:
Adds the ability to compute the cumulative density for the squared error of a multivariate gaussian sample with a known covariance.

Includes the addition of the Kolmogorov Smirnov (KS) test for testing whether a set of samples came from a uniform distribution.

### DIFF
--- a/include/albatross/src/evaluation/prediction_metrics.hpp
+++ b/include/albatross/src/evaluation/prediction_metrics.hpp
@@ -124,7 +124,11 @@ struct NegativeLogLikelihood : public PredictionMetric<PredictType> {
 
 inline double chi_squared_cdf(const JointDistribution &prediction,
                               const MarginalDistribution &truth) {
-  return chi_squared_cdf(prediction.mean - truth.mean, prediction.covariance);
+  Eigen::MatrixXd covariance(prediction.covariance);
+  if (truth.has_covariance()) {
+    covariance += truth.covariance;
+  }
+  return chi_squared_cdf(prediction.mean - truth.mean, covariance);
 }
 
 struct ChiSquaredCdf : public PredictionMetric<JointDistribution> {
--- a/include/albatross/src/evaluation/prediction_metrics.hpp
+++ b/include/albatross/src/evaluation/prediction_metrics.hpp
@@ -121,6 +121,16 @@ struct NegativeLogLikelihood : public PredictionMetric<PredictType> {
   NegativeLogLikelihood()
       : PredictionMetric<PredictType>(negative_log_likelihood) {}
 };
+
+inline double chi_squared_cdf(const JointDistribution &prediction,
+                              const MarginalDistribution &truth) {
+  return chi_squared_cdf(prediction.mean - truth.mean, prediction.covariance);
+}
+
+struct ChiSquaredCdf : public PredictionMetric<JointDistribution> {
+  ChiSquaredCdf() : PredictionMetric<JointDistribution>(chi_squared_cdf){};
+};
+
 } // namespace albatross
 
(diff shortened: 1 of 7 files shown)


include/albatross/src/evaluation/prediction_metrics.hpp
```diff
@@ -121,6 +121,16 @@ struct NegativeLogLikelihood : public PredictionMetric<PredictType> {
   NegativeLogLikelihood()
       : PredictionMetric<PredictType>(negative_log_likelihood) {}
 };
+
+inline double chi_squared_cdf(const JointDistribution &prediction,
+                              const MarginalDistribution &truth) {
+  return chi_squared_cdf(prediction.mean - truth.mean, prediction.covariance);
```
seth-swiftnav: What are the merits/negatives to adding the `truth.diagonal` to the prediction.covariance to the second argument? If the truth is low certainty, any errors would be low certainty aswell.

include/albatross/src/stats/ks_test.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * if the samples came from a uniform distribution.
+ */
+inline double uniform_ks_test(const std::vector<double> &samples) {
+  double largest_difference_between_sample_and_expected_cdf = 0.;
+  std::vector<double> sorted(samples);
+  std::sort(sorted.begin(), sorted.end());
+  double n = static_cast<double>(sorted.size());
+  for (std::size_t i = 0; i < sorted.size(); ++i) {
+
+    double difference = std::max((i + 1) / n - sorted[i], sorted[i] - i / n);
```
seth-swiftnav: I would've expected something symmetric here like `fabs(  sorted[i] -  i/(n-1)  )`, but doing n=1 and n=2 in my head, this works out better.

include/albatross/src/evaluation/prediction_metrics.hpp
akleeman: Yeah I went back and forth on this.  Though you're swaying me back again.  On one hand the prediction might contain measurement noise already, but yeah the truth may also have a different type of measurement noise.  It's probably worth capturing that here.  I'll change it.

include/albatross/src/stats/ks_test.hpp
akleeman: Yeah I originally an abs in there ... then switched after looking at the python implementation:

https://github.com/scipy/scipy/blob/bf3c6471aba06cf6a96843d62f5198ec90e15cf1/scipy/stats/stats.py#L4868

which does what I'm doing here (for the "two_sided") mode.  I'm not totally convinced it's better, but sorta assume enough people use that implementation that it must be somewhat weather proven.

seth-swiftnav: I worded my first comment poorly. The python/PR implementation is definitely better for small N examples, and equivalent for large N.

tests/test_stats.cc
```diff
@@ -55,4 +56,54 @@ TEST(test_stats, test_chi_squared) {
...
+  std::default_random_engine gen;
+  std::uniform_real_distribution<double> uniform(0.0, 1.0);
+
+  std::vector<double> samples;
+  std::size_t n = 1000;
+  for (std::size_t i = 0; i < n; ++i) {
+    samples.emplace_back(uniform(gen));
+  }
+
+  EXPECT_LT(uniform_ks_test(samples), 0.05);
```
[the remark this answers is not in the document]
akleeman: ah yeah, should definitely set a seed.  thanks.

PR #173, opened 2019-10-14
Ransac: Output and is_valid_candidate.

akleeman:
This PR improves ransac in two different ways.

1. A `RansacOutput` struct is now returned which contains information about the inliers, outliers a return code and the final consensus metric all of which are intended to provide more information upstream when a model fit using RANAC has failed.
2. An `is_valid_candidate` tie in which lets you reject a candidate set.  I added this after I encountered an edge case in which a candidate set which contained an outlier was able to reasonably predict other outliers which in turn ended up in the final consensus set.  The default behavior for generic models is to accept all candidate sets (so no change from master) but for Gaussian process RANSAC the chi squared statistic of the candidate set is checked to do a first scan for outliers.

### DIFF
--- a/include/albatross/src/models/ransac.hpp
+++ b/include/albatross/src/models/ransac.hpp
@@ -78,6 +78,26 @@ struct RansacOutput {
   double consensus_metric;
 };
 
+struct RansacConfig {
+
+  RansacConfig(){};
+
+  RansacConfig(double inlier_threshold_, std::size_t random_sample_size_,
+               std::size_t min_consensus_size_, std::size_t max_iterations_,
+               std::size_t max_failed_candidates_)
+      : inlier_threshold(inlier_threshold_),
+        random_sample_size(random_sample_size_),
+        min_consensus_size(min_consensus_size_),
+        max_iterations(max_iterations_),
+        max_failed_candidates(max_failed_candidates_){};
+
+  double inlier_threshold;
+  std::size_t random_sample_size;
+  std::size_t min_consensus_size;
+  std::size_t max_iterations;
+  std::size_t max_failed_candidates;
+};
+
 /*
  * This RANdom SAmple Consensus (RANSAC) algorithm works as follows.
  *
@@ -187,7 +207,16 @@ auto ransac(const RansacFunctions<FitType> &ransac_functions,
             std::size_t random_sample_size, std::size_t min_consensus_size,
             std::size_t max_iterations) {
   return ransac(ransac_functions, map_keys(indexer), inlier_threshold,
-                random_sample_size, min_consensus_size, max_iterations);
+                random_sample_size, min_consensus_size, max_iterations,
+                max_iterations);
+}
+
+template <typename FitType>
+auto ransac(const RansacFunctions<FitType> &ransac_functions,
+            const FoldIndexer &indexer, const RansacConfig &config) {
+  return ransac(ransac_functions, map_keys(indexer), config.inlier_threshold,
+                config.random_sample_size, config.min_consensus_size,
+                config.max_iterations, config.max_failed_candidates);
 }
 
 template <typename ModelType, typename FeatureType, typename InlierMetric,
@@ -334,26 +363,6 @@ struct Fit<RansacFit<ModelType, StrategyType, FeatureType>> {
   RansacOutput ransac_output;
 };
 
-struct RansacConfig {
-
-  RansacConfig(){};
-
-  RansacConfig(double inlier_threshold_, std::size_t random_sample_size_,
-               std::size_t min_consensus_size_, std::size_t max_iterations_,
-               std::size_t max_failed_candidates_)
-      : inlier_threshold(inlier_threshold_),
-        random_sample_size(random_sample_size_),
-        min_consensus_size(min_consensus_size_),
-        max_iterations(max_iterations_),
-        max_failed_candidates(max_failed_candidates_){};
-
-  double inlier_threshold;
-  std::size_t random_sample_size;
-  std::size_t min_consensus_size;
-  std::size_t max_iterations;
-  std::size_t max_failed_candidates;
-};
-
 /*
  * This wraps any other model and performs ransac each time fit is called.
  */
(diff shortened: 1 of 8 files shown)


include/albatross/src/models/ransac_gp.hpp
```diff
@@ -36,11 +36,30 @@ inline
...
+  return [&, indexer, cov, dataset](const std::vector<FoldName> &groups) {
+    auto inds = indices_from_names(indexer, groups);
+    const auto train_dataset = subset(dataset, inds);
+    const auto train_cov = symmetric_subset(cov, inds);
+
+    const JointDistribution prior(Eigen::VectorXd::Zero(train_cov.rows()),
+                                  train_cov);
+    const double chi2 = chi_squared_cdf(prior, train_dataset.targets);
+    const double skip_every_1000_candidates = 0.999;
+    return (chi2 < skip_every_1000_candidates);
```
seth-swiftnav: The var name is confusing to me. Does it skip every 1000 candidates or skip one on every 1000 candidates? (the later). But it only skips 1/1000 candidates if the candidates are sampled from the prior, so this is really something designed to remove things that don't look like the prior at all. `inclusion_ratio` or `probabilty_threshold` seem like good terms here.
`probability_targets_not_from_prior` or similar could replace `chi2`.
There's also an asymmetry with low chi2 values not being removed.

include/albatross/src/models/ransac.hpp
```diff
@@ -100,21 +153,26 @@ ransac(const RansacFunctions<FitType> &ransac_functions,
...
+        output.consensus_metric = consensus_metric_value;
+        output.outliers = outliers;
       }
     }
+
+    ++i;
   }
-  return best_consensus;
+
+  output.return_code = RANSAC_RETURN_CODE_SUCCESS;
```
seth-swiftnav: Logically, it looks like if no valid candidate set forms a valid consensus set, then we'll result in setting the return code to `rc_SUCCESS` despite having no `output.inliers`, which downstream will result in an empty fit/crash.

include/albatross/src/models/ransac.hpp
```diff
@@ -289,38 +395,37 @@ class Ransac : public ModelBase<Ransac<ModelType, StrategyType>> {
...
-        sub_model_.fit(consensus_set), inliers, outliers);
   }
 
   template <typename PredictFeatureType, typename FitType, typename PredictType>
   PredictType _predict_impl(const std::vector<PredictFeatureType> &features,
                             const FitType &ransac_fit_,
                             PredictTypeIdentity<PredictType> &&) const {
-    return ransac_fit_.fit_model.predict(features).template get<PredictType>();
+    // If RANSAC failed it's up to the user to determine that from the output of
+    // fit and deal with it appropriately.
```
seth-swiftnav: This comment says "If RANSAC failed" ... but it seems like this still includes "rc_Success" but with an empty_fit(). Is empty_fit a success or a failure? I'd say failure, since it can't be used to make predications.

include/albatross/src/models/ransac_gp.hpp
akleeman: I changed it to `skip_every_1000th_candidate`, that make it more clear?  and I'll change `chi2` to `probability_prior_exceeded`.  I think the asymmetry is OK.  For example, we may want to use a somewhat relaxed prior to perform a cursory check for outliers before proceeding with RANSAC, so I feel like we only want o discard large values.

include/albatross/src/models/ransac.hpp
```diff
@@ -29,21 +33,49 @@ template <typename FitType> struct RansacFunctions {
...
+} ransac_return_code_t;
+
+inline bool ransac_success(const ransac_return_code_t &rc) {
+  return rc == RANSAC_RETURN_CODE_SUCCESS;
+}
+
+struct RansacOutput {
+
+  RansacOutput()
+      : return_code(RANSAC_RETURN_CODE_FAILURE), inliers(), outliers(),
```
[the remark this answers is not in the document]
akleeman: changed to `RANSAC_RETURN_CODE_INVALID`.

include/albatross/src/models/ransac.hpp
akleeman: yep good catch.

include/albatross/src/models/ransac.hpp
akleeman: assuming this is the same issue you pointed out on line 166?  If it's something else let me know.

include/albatross/src/models/ransac.hpp
```diff
@@ -408,17 +417,15 @@ class Ransac : public ModelBase<Ransac<ModelType, StrategyType>> {
...
-               config_.max_iterations);
+    const auto ransac_output = ransac(ransac_functions, indexer, config_);
 
     if (ransac_success(ransac_output.return_code)) {
       const auto good_inds = indices_from_names(indexer, ransac_output.inliers);
       const auto consensus_set = subset(dataset, good_inds);
       return Fit<RansacFit<ModelType, StrategyType, FeatureType>>(
           sub_model_.fit(consensus_set), ransac_output);
     } else {
+      std::cout << ransac_output.return_code << std::endl;
```
seth-swiftnav: Debug print statement inside of primary albatross code?

include/albatross/src/models/ransac.hpp
seth-swiftnav: This is the same issue I mentioned before. Thankyou for addressing it :)

include/albatross/src/models/ransac_gp.hpp
seth-swiftnav: I can understand the code right now, but I still think `skip_every_1000th_candididate`, while more clear, isn't what it's actually doing. It skips every 1000th candidate sampled from the prior, but it removes things that don't look like the prior outright.

tests/test_ransac.cc
```diff
@@ -137,4 +137,61 @@ TEST(test_outlier, test_ransac_groups) {
   EXPECT_LE(result.inliers.size(), indexer.size());
 }
 
+inline bool never_accept_candidates(const std::vector<FoldName> &) {
+  return false;
+}
+
+RansacConfig get_reasonable_ransac_config() {
```
seth-swiftnav: nit: I love the joke, but I'm unsure how it balances with professionalism.

akleeman: Heh yeah there was no joke intended. changed the NAN to a 0. .

PR #182, opened 2019-10-25
Actually Use Group By

akleeman:
In #160 a `group_by` method was added for use with `vector` and `RegressionDataset` but until this change it wasn't actually used anywhere.  Switching to `group_by` means that a grouping function can return anything which is a valid `std::map` key (instead of being restricted to a `std::string`).  As a result:

- Cross validation was refactored to remove `LeaveOneOut` and `LeaveOneGroupOut` in favor of using grouper functions (this refactor is arguably the biggest "actual" change in this PR).
- Anything which works with groups (`Ransac`, `SparseGaussianProcess`, etc ...) needed to also keep track of the `GroupKey` now (the majority of changed lines come from this).
- `RegressionFolds` is now a map from `GroupKey` to `RegressionFold` (instead of a vector of folds).  As a result the `RegressionFold` no longer contains the `FoldName` member (pretty minor change).
- Some trait logic had to be updated to consider the `GroupKey`.

### DIFF
--- a/include/albatross/src/core/declarations.hpp
+++ b/include/albatross/src/core/declarations.hpp
@@ -52,7 +52,7 @@ template <typename GroupKey, typename ValueType> class Grouped;
 template <typename GroupKey>
 using GroupIndexer = Grouped<GroupKey, GroupIndices>;
 
-struct LeaveOneOut;
+struct LeaveOneOutGrouper;
 
 template <typename ValueType, typename GrouperFunction> class GroupBy;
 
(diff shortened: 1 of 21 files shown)


include/albatross/src/core/declarations.hpp
```diff
@@ -94,24 +94,16 @@ template <typename ImplType = NullLeastSquaresImpl> class LeastSquares;
...
-using FoldIndices = std::vector<std::size_t>;
-using FoldName = std::string;
-using FoldIndexer = std::map<FoldName, FoldIndices>;
-
 template <typename FeatureType> struct RegressionFold;
 
-template <typename FeatureType>
-using GroupFunction = std::string (*)(const FeatureType &);
+template <typename GroupKey, typename FeatureType>
+using RegressionFolds = Grouped<GroupKey, RegressionFold<FeatureType>>;
```
jbangelo: You've gotta be kidding me with this diff github... 🤦‍♂

PR #184, opened 2019-11-06
GroupBy helper methods.

akleeman:
Adds a variety of methods such as `sum`, `mean`, `min_vale`, `max_value` for dealing with grouped types which are numeric. 

Also contains helper methods for accessing groups such as `first_group` and `get_group` for when you only want one group.

### DIFF
--- a/include/albatross/src/indexing/group_by.hpp
+++ b/include/albatross/src/indexing/group_by.hpp
@@ -71,6 +71,22 @@ public:
 
   std::map<KeyType, ValueType> get_map() const { return map_; }
 
+  std::pair<KeyType, ValueType> first_group() const { return *map_.begin(); }
+
+  auto min_value() const {
+    const auto value_compare = [](const auto &x, const auto &y) {
+      return x.second < y.second;
+    };
+    return std::min_element(begin(), end(), value_compare)->second;
+  }
+
+  auto max_value() const {
+    const auto value_compare = [](const auto &x, const auto &y) {
+      return x.second < y.second;
+    };
+    return std::max_element(begin(), end(), value_compare)->second;
+  }
+
   /*
    * Filtering a Grouped object consists of deciding which of the
    * groups you would like to keep.  This is done by providing a function which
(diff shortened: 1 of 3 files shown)


include/albatross/src/indexing/group_by.hpp
```diff
@@ -71,6 +71,22 @@ template <typename KeyType, typename ValueType> class GroupedBase {
 
   std::map<KeyType, ValueType> get_map() const { return map_; }
 
+  std::pair<KeyType, ValueType> first_group() const { return *map_.begin(); }
+
+  auto min_value() const {
+    const auto value_compare = [](const auto &x, const auto &y) {
```
jbangelo: You don't need to change it, but I don't think the `const` is strictly necessary here. Lambda captures are `const` by default in C++, so the `const` here isn't applied to the captures. And the fact that they have a unique and anonymous type makes it difficult for me to see how you would reassign the variable holding the lambda;

include/albatross/src/indexing/group_by.hpp
```diff
@@ -174,12 +190,40 @@ template <typename KeyType, typename ValueType> class GroupedBase {
...
+template <typename KeyType, typename ValueType>
+class Grouped<KeyType, ValueType,
+              std::enable_if_t<std::is_arithmetic<ValueType>::value>>
+    : public GroupedBase<KeyType, ValueType> {
+  using Base = GroupedBase<KeyType, ValueType>;
+  using Base::Base;
+
+public:
+  auto sum() const {
+    double output = 0.;
```
jbangelo: Wouldn't you want this type to be `ValueType`? In the case of large-ish integer values you'd end up losing precision by internally using a `double` to contain the sum.

include/albatross/src/indexing/group_by.hpp
akleeman: Oh I see, so just `auto value_compare`?  I *have* been able to modify variables captured by a lambda, I can actually do so even if the function is marked `const`.  This for example outputs `1`:
```
    int i = 0;
    const auto increment = [&]() {
        ++i;
    };
    increment();
    std::cout << i << std::endl;
```

include/albatross/src/indexing/group_by.hpp
akleeman: yeah, good call.

include/albatross/src/indexing/group_by.hpp
jbangelo: **_Somewhat moot tangent incoming_**

That's expected behavior. In that case the captured variable is a reference which is `const`, not a reference to a `const`. Put another way the type of the captured `i` variable is `int& const` not `const int&`. It's super confusing. The end result is that lambdas can't modify their internal state by default, but can modify external state if that's how you capture it. Capturing by value wouldn't compile in that case:
```
int i = 0;
auto increment = [=]() {
    ++i; // error: increment of read-only variable 'i'
};
increment();

auto mut_increment = [=]() mutable {
    ++i; // compiles
};
mut_increment();
```

include/albatross/src/indexing/group_by.hpp
```diff
@@ -174,12 +190,40 @@ template <typename KeyType, typename ValueType> class GroupedBase {
...
+  auto sum() const {
+    ValueType output = 0.;
+    for (const auto &pair : this->map_) {
+      output += pair.second;
+    }
+    return output;
+  }
+
+  auto mean() const {
+    double output = 0.;
```
jbangelo: I guess for this function it's a more nuanced problem and we wouldn't really want to use `ValueType` here. Or at least we'd want to calculate the sum differently in the case of it being an integer...

I guess a `double` is the best and easiest choice.

include/albatross/src/indexing/group_by.hpp
akleeman: I see, thanks for the explanation!

PR #185, opened 2019-11-06
Add free functions for using apply directly on vectors.

akleeman:
This change lets you do things like:

```
std::vector<T> foo;
const auto bar = apply(foo, convert_foo_to_bar);
```

It's not unlike `std::transform` but supports situations where the apply function returns `void` and doesn't require knowing the type of the transformed type ahead of time.

### DIFF
--- a/include/albatross/src/indexing/apply.hpp
+++ b/include/albatross/src/indexing/apply.hpp
@@ -30,6 +30,34 @@ inline auto filter(const std::vector<ValueType> &values,
   return output;
 }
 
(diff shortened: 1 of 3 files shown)


akleeman:
@jbangelo yeah ... they pretty much are.  I just dislike the syntax for `transform` etc.  (Though maybe I should use it internally).

jbangelo:
I don't see a reason to suddenly start. I agree that they are clunky to use, requiring the pairs of iterators. The "range" versions in C++20 are nicer to use.

PR #186, opened 2019-11-06
Any/All method

akleeman:
Adds methods for checking if any/all `bool` values in a vector of `bool` are true.

### DIFF
--- a/include/albatross/src/utils/vector_utils.hpp
+++ b/include/albatross/src/utils/vector_utils.hpp
@@ -32,10 +32,14 @@ inline std::vector<double> linspace(double a, double b, std::size_t n) {
 }
 
 inline bool all(const std::vector<bool> &xs) {
+  // The thinking on having all({}) return true comes from interpreting "all" as
+  // "is there anything that isn't true".
   if (xs.size() == 0) {
-    return false;
+    return true;
   }
-  for (const auto &x : xs) {
+  // Due to optimizations of vector<bool> you can't get actual references so we
+  // copy here to avoid future issues.
+  for (const auto x : xs) {
     if (!x) {
       return false;
     }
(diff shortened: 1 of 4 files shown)


include/albatross/src/utils/vector_utils.hpp
```diff
@@ -31,6 +31,30 @@ inline std::vector<double> linspace(double a, double b, std::size_t n) {
   return xs;
 }
 
+inline bool all(const std::vector<bool> &xs) {
```
jbangelo: Uh oh, you've stumbled into the dreaded `std::vector<bool>`! There's a specialization in the standard for `std::vector<bool>` which attempts to save space by storing each `bool` in a single bit. The consequence of this is that when you iterate over a `std::vector<bool>` you don't get actual references you get reference like classes.

TL;DR: I think the `for` loops below probably shouldn't make `x` be a reference.

include/albatross/src/utils/vector_utils.hpp
```diff
@@ -31,6 +31,30 @@ inline std::vector<double> linspace(double a, double b, std::size_t n) {
...
+  if (xs.size() == 0) {
+    return false;
+  }
+  for (const auto &x : xs) {
```
jbangelo:
```suggestion
  for (const auto x : xs) {
```

include/albatross/src/utils/vector_utils.hpp
```diff
@@ -31,6 +31,30 @@ inline std::vector<double> linspace(double a, double b, std::size_t n) {
   return xs;
 }
 
+inline bool all(const std::vector<bool> &xs) {
+  if (xs.size() == 0) {
+    return false;
```
jbangelo: I've usually seen this base case for this all type function default to `true` when empty and the `any` function default to `false` when empty. I guess in this situation you're logic would be simpler with both returning `false`?

akleeman: Yeah I wasn't totally sure about the empty cases.  I think I agree that `all` should probably return `true`.  I was tempted to put an assert in there ... but that didn't feel right.

PR #188, opened 2019-11-06
Improve consistency of GrouperFunction use.

akleeman:
The ransac strategies were some what inconsistent in the order of arguments, template parameters and constructor arguments.  This resolves several of those instances.

### DIFF
--- a/include/albatross/src/models/ransac_gp.hpp
+++ b/include/albatross/src/models/ransac_gp.hpp
@@ -187,15 +187,18 @@ get_gp_ransac_functions(
 };
 
 template <typename InlierMetric, typename ConsensusMetric,
-          typename GrouperFunction, typename IsValidCandidateMetric>
+          typename IsValidCandidateMetric, typename GrouperFunction>
 struct GaussianProcessRansacStrategy {
 
   GaussianProcessRansacStrategy() = default;
 
-  GaussianProcessRansacStrategy(const InlierMetric &inlier_metric,
-                                const ConsensusMetric &consensus_metric,
-                                const GrouperFunction &grouper_function)
+  GaussianProcessRansacStrategy(
+      const InlierMetric &inlier_metric,
+      const ConsensusMetric &consensus_metric,
+      const IsValidCandidateMetric &is_valid_candidate,
+      GrouperFunction grouper_function)
       : inlier_metric_(inlier_metric), consensus_metric_(consensus_metric),
+        is_valid_candidate_(is_valid_candidate),
         grouper_function_(grouper_function){};
 
   template <typename ModelType, typename FeatureType>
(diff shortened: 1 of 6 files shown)


include/albatross/src/models/ransac_gp.hpp
```diff
@@ -214,34 +217,35 @@ struct GaussianProcessRansacStrategy {
...
+          typename IsValidCandidateMetric, typename GrouperFunction>
 auto gp_ransac_strategy(const InlierMetric &inlier_metric,
-                        const IndexingFunction &indexing_function,
                         const ConsensusMetric &consensus_metric,
-                        const IsValidCandidateMetric &is_valid_candidate) {
-  return GaussianProcessRansacStrategy<
-      InlierMetric, ConsensusMetric, IndexingFunction, IsValidCandidateMetric>(
-      inlier_metric, consensus_metric, indexing_function);
+                        const IsValidCandidateMetric &is_valid_candidate,
+                        GrouperFunction grouper_function) {
```
seth-swiftnav: Should these all be const reference?

include/albatross/src/models/ransac_gp.hpp
```diff
@@ -187,15 +187,18 @@ get_gp_ransac_functions(
...
   GaussianProcessRansacStrategy() = default;
 
-  GaussianProcessRansacStrategy(const InlierMetric &inlier_metric,
-                                const ConsensusMetric &consensus_metric,
-                                const GrouperFunction &grouper_function)
+  GaussianProcessRansacStrategy(
+      const InlierMetric &inlier_metric,
+      const ConsensusMetric &consensus_metric,
+      const IsValidCandidateMetric &is_valid_candidate,
+      GrouperFunction grouper_function)
```
seth-swiftnav: Should these all be const reference?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -137,8 +137,7 @@ class SparseGaussianProcessRegression
     initialize_params();
   };
   SparseGaussianProcessRegression(
-      CovFunc &covariance_function,
-      GrouperFunction &independent_group_function_,
+      CovFunc &covariance_function, GrouperFunction independent_group_function_,
```
seth-swiftnav: Can you explain the change of the GrouperFunction from an address to an instance? I don't think its wrong, I just want to understand.

akleeman: Yeah it get's tricky.  Basically the `GrouperFunction` needs to be non-const reference in order to support both passing a function directly as well as passing a function pointer.  It has to do with the way the compiler resolves the templated parameter.  I think it tries to     Here's an example that might illustrate what happens:
```
template <typename Func>
auto get_foo_const_ref(const Func &func) {
    std::cout << __PRETTY_FUNCTION__ << std::endl;
    return func();
}

template <typename Func>
auto get_foo(Func func) {
    std::cout << __PRETTY_FUNCTION__ << std::endl;
    return func();
}

void something() {};

int main() {

    get_foo(something);
    get_foo(&something);
    get_foo([](){something();});    

    get_foo_const_ref(something);
    get_foo_const_ref(&something);
    get_foo_const_ref([](){something();});   

    return 0;
}
```
Which outputs:
```
auto get_foo(Func) [with Func = void (*)()]
auto get_foo(Func) [with Func = void (*)()]
auto get_foo(Func) [with Func = main()::<lambda()>]
auto get_foo_const_ref(const Func&) [with Func = void()]
auto get_foo_const_ref(const Func&) [with Func = void (*)()]
auto get_foo_const_ref(const Func&) [with Func = main()::<lambda()>]
```
Notice that `get_foo` converts `something` into a function pointer where `get_foo_const_ref` does not. @jbangelo might be able to shed light on why it happens.  But in our case we end up storing the function inside `SparseGaussianProcessRegression` and in order to store it we need to make sure it's a pointer so only the non const ref version works.

PR #190, opened 2019-11-14
Variant Utils

akleeman:
Add VariantUtils which adds tools for applying trait logic to variant types as well as ways to easily convert between possibly different variant types.

Specifically it deals with situations where you may have compatible but not equivalent variant types such as,
```
variant<X. Y> foo;
variant<Y, X> bar;
```
where assignment operators don't work (ie, `foo = bar` won't compile).  For those situations you can now do:
```
set_variant(foo, bar);
```
similarly you can convert vectors of variants,
```
std::vector<variant<X, Y>> foos;

std::vector<variant<Y, X>> bars = to_variant_vector<variant<Y, X>>(foos);
```

### DIFF
--- a/include/albatross/src/utils/variant_utils.hpp
+++ b/include/albatross/src/utils/variant_utils.hpp
@@ -25,8 +25,8 @@ template <typename X> struct ToVariantIdentity {};
  */
 template <typename... Ts, typename X>
 inline std::enable_if_t<is_sub_variant<X, variant<Ts...>>::value, void>
-set_variant(variant<Ts...> &to_set, const X &x) {
-  x.match([&to_set](const auto &v) { to_set = v; });
+set_variant(const X &x, variant<Ts...> *to_set) {
+  x.match([&to_set](const auto &v) { *to_set = v; });
 }
 
 /*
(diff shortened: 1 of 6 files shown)


include/albatross/src/utils/variant_utils.hpp
```diff
@@ -0,0 +1,66 @@
...
+template <typename X> struct ToVariantIdentity {};
+} // namespace details
+
+/*
+ * In this case X is a variant which is compatible so we need to
+ * convert it.
+ */
+template <typename... Ts, typename X>
+inline std::enable_if_t<is_sub_variant<X, variant<Ts...>>::value, void>
+set_variant(variant<Ts...> &to_set, const X &x) {
```
akleeman: I'm not totally sure what to do here.  On one hand I like the current syntax `set_variant(to, from)` since it preserves the order of `to = from`, but it does conflict with code style which would probably prefer `set_variant(from, &to)`.  Maybe renaming it `assign_to(from, &to)` would be a good option?  Thoughts?

include/albatross/src/utils/variant_utils.hpp
```diff
@@ -0,0 +1,66 @@
...
+  for (std::size_t i = 0; i < xs.size(); ++i) {
+    set_variant(to_set[i], xs[i]);
+  }
+}
+
+template <typename OutputType, typename X,
+          std::enable_if_t<is_variant<OutputType>::value, int> = 0>
+inline std::vector<OutputType>
+to_variant_vector(const std::vector<X> &xs,
+                  details::ToVariantIdentity<OutputType> && = {}) {
```
jbangelo: This second argument is only here to assist in the type deduction?

include/albatross/src/utils/variant_utils.hpp
jbangelo: Have you typically followed one style over the other in albatross? I'd suggest to keep the code inside of a project/repo consistent over starting to follow the suggested style half way through and having inconsistent style.

include/albatross/src/utils/variant_utils.hpp
akleeman: I put it there so we don't end up with overloads with different return types.  I had to do this for the `predict` methods, though it occurs to me the behavior may be slightly different for free functions so maybe this isn't required.

include/albatross/src/utils/variant_utils.hpp
akleeman: Ah yeah actually a quick grep reminded me that I do this `set_subset` which passes the `to` as a pointer.  I'll switch to that.

include/albatross/src/utils/variant_utils.hpp
jbangelo: I think those rules are the same for free functions and member functions.

PR #195, opened 2019-11-22
WIP: Conditionally Independent Merge

akleeman:
This work in progress adds a method for merging two Gaussian process fits under the assumption that the corresponding sets of observations are conditionally independent of each give some state space representation.


akleeman:
A few experiments indicated this method may "work" but the independence assumption was too strong for our use case.  Closing under the assumption that I won't get around to finishing it.

PR #196, opened 2019-11-22
Random Utils

akleeman:
This creates an include target `#include <albatross/utils/RandomUtils>` and adds a bunch of utility functions dealing with producing random `Eigen` vectors and matrices.  `Eigen` includes some methods for this already, but they mostly deal with uniform distributions.  Here we deal mostly with mutlivariate gaussian distributions.

A lot of these methods were already used in tests, but have now been moved to a more accessible location (and generalized a bit).

### DIFF
--- a/include/albatross/src/utils/eigen_utils.hpp
+++ b/include/albatross/src/utils/eigen_utils.hpp
@@ -142,31 +142,6 @@ inline auto truncated_psd_solve(
   return output;
 }
 
-template <typename _Scalar, int _Rows, int _Cols,
-          typename RandomNumberGenerator>
-void gaussian_fill(Eigen::Matrix<_Scalar, _Rows, _Cols> &matrix, double mean,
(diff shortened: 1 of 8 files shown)


include/albatross/src/evaluation/prediction_metrics.hpp
```diff
@@ -71,13 +71,29 @@ struct RootMeanSquareError : public PredictionMetric<Eigen::VectorXd> {
...
+  if (x.size() == 1) {
+    return 0.;
+  }
+
+  double output = 0.;
+  double mean = x.mean();
+  const auto n_elements = static_cast<double>(x.size());
+
+  for (Eigen::Index i = 0; i < x.size(); ++i) {
+    output += pow(x[i] - mean, 2.) / (n_elements - 1);
```
pgrgich: The division can be done outside the loop.

PR #197, opened 2019-12-05
Random Covariance

akleeman:
Some utilities for producing random matrices, covariance matrices in particular, already exist in `albatross` but here we put these tools in the same place and add a target `RandomUtils` to make them easier to use.


akleeman:
Ooops, duplicated with:
https://github.com/swift-nav/albatross/pull/197

PR #198, opened 2019-12-05
Grouped Eigen Type Safety

akleeman:
This change puts in a safety check to make sure you don't accidentally put an `Eigen` expression into a map.  Doing so can lead to unexpected behavior as things lose scope.

For example, we don't want to do something like:
```
Eigen::MatrixXd lhs = Eigen::MatrixXd::Random(3, 4);
auto product_eval = [&](const Eigen::MatrixXd &rhs) {
    return lhs * rhs;
};

auto product = group_of_matrices.apply(product_eval);
```
The reason is that you may expect the result (`product`) would also contain matrices ... but they'll actually hold something like `Eigen::Product<Eigen::MatrixXd>` objects which hold reference to the objects involved.  As a result if (for example) `lhs` were to lose scope you may get memory access crashes.

To play it safe in these situations you'll need to add `.eval()` to the end of the expression.

### DIFF
--- a/include/albatross/src/details/traits.hpp
+++ b/include/albatross/src/details/traits.hpp
@@ -162,6 +162,35 @@ struct variant_size<variant<Ts...>>
     : public std::tuple_size<std::tuple<Ts...>> {};
 
 /*
+ * Eigen helpers
+ */
+template <typename T> class is_eigen_plain_object {
+  template <typename C>
+  static typename std::enable_if<
+      std::is_base_of<Eigen::PlainObjectBase<C>, C>::value,
+      std::true_type>::type
+  test(int);
+
+  template <typename C> static std::false_type test(...);
+
+public:
+  static constexpr bool value = decltype(test<T>(0))::value;
+};
+
+template <typename T> class is_eigen_xpr {
+  template <
+      typename C,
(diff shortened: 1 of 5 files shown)


include/albatross/src/details/traits.hpp
```diff
@@ -161,6 +161,35 @@ template <typename... Ts>
...
+  template <typename C> static std::false_type test(...);
+
+public:
+  static constexpr bool value = decltype(test<T>(0))::value;
+};
+
+template <typename T> class is_eigen_xpr {
+  template <
+      typename C,
+      typename BaseType = typename Eigen::internal::dense_xpr_base<C>::type,
```
jbangelo: Is the `BaseType` parameter needed? I'm not seeing how it's used.

jbangelo:
I guess there is a compiler error.

include/albatross/src/details/traits.hpp
akleeman: Yeah it actually is used, but I should make it more clear how.  Basically `Eigen::internal::dense_xpr_base<C>::type` will fail if `C` isn't an eigen matrix like type, so that line checks that it's possibly an eigen type, and the next line checks if it isn't a plain object.

PR #199, opened 2019-12-05
Add CRTP safety to ModelMetrics and convenience method to

akleeman:
This minor change puts some safety checks and extra utility methods to the metrics.

### DIFF
--- a/include/albatross/src/evaluation/prediction_metrics.hpp
+++ b/include/albatross/src/evaluation/prediction_metrics.hpp
@@ -71,13 +71,29 @@ struct RootMeanSquareError : public PredictionMetric<Eigen::VectorXd> {
       : PredictionMetric<Eigen::VectorXd>(root_mean_square_error) {}
 };
 
+static inline double standard_deviation(const Eigen::VectorXd &x) {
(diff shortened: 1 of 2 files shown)


include/albatross/src/evaluation/prediction_metrics.hpp
```diff
@@ -71,13 +71,29 @@ struct RootMeanSquareError : public PredictionMetric<Eigen::VectorXd> {
...
+
+  double output = 0.;
+  double mean = x.mean();
+  const auto n_elements = static_cast<double>(x.size());
+
+  for (Eigen::Index i = 0; i < x.size(); ++i) {
+    output += pow(x[i] - mean, 2.) / (n_elements - 1);
+  }
+
+  return std::sqrt(output);
```
seth-swiftnav: What's the reason behind this change from the prior implementation with `dot`?

akleeman: I didn't actually run into this issue ... but it's possible that summing all the squared values before averaging could lead to a loss in numerical precision.  By dividing iteratively you avoid this.

PR #201, opened 2019-12-20
Patchwork Refactor

akleeman:
This PR builds off #200 with more tests and with the majority of the work pushed into the `fit` method.


akleeman:
See #200

PR #202, opened 2019-12-23
Serialization Improvements

akleeman:
This change includes a few modifications which enable:

- Serialization of the model name in Gaussian process models.
- The ability to write variant types to csv.
- Serialization of the variant typeid along with the variant to make inspection easier.

### DIFF
--- a/include/albatross/src/cereal/variant.hpp
+++ b/include/albatross/src/cereal/variant.hpp
@@ -63,6 +63,10 @@ inline void save(Archive &archive, const variant<VariantTypes...> &f,
                  const std::uint32_t) {
 
   archive(cereal::make_nvp("which", f.which()));
+  f.match([&](const auto &x) {
+    std::string which_typeid = typeid(x).name();
+    archive(cereal::make_nvp("which_typeid", which_typeid));
+  });
   f.match([&archive](const auto &x) { archive(cereal::make_nvp("data", x)); });
 }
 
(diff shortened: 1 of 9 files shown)


seth-swiftnav:
I haven't reviewed the code yet, but question - is this compatible with current serialization?

akleeman:
@seth-swiftnav I haven't tested it yet ... but it should be backward compatible (though it may need a serialization version bump).  Basically extra fields have been added to, but not removed from, the existing serialization.

include/albatross/src/cereal/variant.hpp
```diff
@@ -63,6 +63,10 @@ inline void save(Archive &archive, const variant<VariantTypes...> &f,
                  const std::uint32_t) {
 
   archive(cereal::make_nvp("which", f.which()));
+  f.match([&](const auto &x) {
+    std::string which_typeid = typeid(x).name();
```
jbangelo: This will give an implementation defined name for the type. It's not guaranteed to be the same across compilers, or ever versions of the same compiler.  How critical is it that these names are always the same?

jbangelo: I also think this tends to give the mangled name, in case that's a problem.

akleeman: Yeah that's totally fine.  I mostly added this so that when you look at serialized output you can tell which of the variants was used (since the `which` element is just an integer).

jbangelo: Makes sense. I guess the `which` field is just has fragile since the order of the template parameters.

PR #207, opened 2020-01-13
Sparse GP Likelihood

akleeman:
Add a log likelihood method for sparse Gaussian processes in order to allow us to perform Bayesian inference on the parameters of sparse models.

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -117,6 +117,8 @@ struct StateSpaceInducingPointStrategy {
  * get those deails straight:
  *
  *     - https://bwengals.github.io/pymc3-fitcvfe-implementation-notes.html
+ *        (beware, the final formulae are correct but there are typos in
+ *         their derivations).
  *     - https://github.com/SheffieldML/GPy see fitc.py
  */
 template <typename CovFunc, typename MeanFunc, typename GrouperFunction,
@@ -200,22 +202,23 @@ public:
     }
   }
 
-  template <typename FeatureType>
-  auto _fit_impl(const std::vector<FeatureType> &out_of_order_features,
-                 const MarginalDistribution &out_of_order_targets) const {
-    static_assert(is_invocable<GrouperFunction, FeatureType>::value,
-                  "GrouperFunction is not defined for the required types");
-    static_assert(
-        is_invocable<InducingPointStrategy, CovFunc,
-                     std::vector<FeatureType>>::value,
-        "InducingPointStrategy is not defined for the required types");
+  struct SparseGPComponents {
+    Eigen::LLT<Eigen::MatrixXd> K_uu_llt;
+    Eigen::MatrixXd P;
+    BlockDiagonalLLT A_llt;
+    Eigen::MatrixXd RtR;
+    Eigen::SerializableLDLT B_ldlt;
+    Eigen::VectorXd y;
+  };
+
+  template <typename FeatureType, typename InducingFeatureType>
+  SparseGPComponents
+  sparse_components(const std::vector<FeatureType> &out_of_order_features,
+                    const std::vector<InducingFeatureType> &inducing_features,
+                    const MarginalDistribution &out_of_order_targets) const {
 
     const auto indexer =
         group_by(out_of_order_features, independent_group_function_).indexers();
-
-    // Determine the set of inducing points, u.
-    const auto u = inducing_point_strategy_(this->covariance_function_,
-                                            out_of_order_features);
 
     const auto out_of_order_measurement_features =
         as_measurements(out_of_order_features);
@@ -236,12 +239,14 @@ public:
         subset(out_of_order_measurement_features, reordered_inds);
     auto targets = subset(out_of_order_targets, reordered_inds);
 
-    this->mean_function_.remove_from(features, &targets.mean);
+    this->mean_function_.remove_from(
+        subset(out_of_order_features, reordered_inds), &targets.mean);
 
-    Eigen::Index m = static_cast<Eigen::Index>(u.size());
+    Eigen::Index m = static_cast<Eigen::Index>(inducing_features.size());
 
-    const Eigen::MatrixXd K_fu = this->covariance_function_(features, u);
-    Eigen::MatrixXd K_uu = this->covariance_function_(u);
+    const Eigen::MatrixXd K_fu =
+        this->covariance_function_(features, inducing_features);
+    Eigen::MatrixXd K_uu = this->covariance_function_(inducing_features);
 
     K_uu.diagonal() +=
         inducing_nugget_.value * Eigen::VectorXd::Ones(K_uu.rows());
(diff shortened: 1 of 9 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -315,23 +321,100 @@ class SparseGaussianProcessRegression
...
+    //   L = 1/2 (n log(2 pi) + log(|K|) + y^T K^-1 y)
+    //
+    // where in our case we have K = A + Q_ff
+
+    // First we get the determinant, |K|:
+    //
+    //   |K| = |A + Q_ff|
+    //       = |K_uu + K_uf A^-1 K_fu| |K_uu^-1| |A|
+    //       = |LL^T + K_uf A^-1 K_fu| |L^-T L^-1| |A|
+    //       = |L^-1| |LL^T + K_uf A^-1 K_fu| |L^-T| |A|
```
pgrgich: We can move the `L^-1` out the front because `|L^-T L^-1|=|L^-T| |L^-1|` and the determinants are just scalars, is that right?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -315,23 +321,100 @@ class SparseGaussianProcessRegression
...
+    //       = |I + L^-1 K_uf A^-1 K_fu L^-T| |A|
+    //       = |I + P A^-1 P^T| |A|
+    //       = |B| |A|
+    //
+    // which is derived here (though there are errors in their derivation):
+    //   https://bwengals.github.io/pymc3-fitcvfe-implementation-notes.html
+
+    const double log_det_a = sc.A_llt.log_determinant();
+    const double log_det_b = sc.B_ldlt.vectorD().array().log().sum();
```
pgrgich: Why does `B`s determinant not involve the `L` part of the decomposition?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -315,23 +321,100 @@ class SparseGaussianProcessRegression
...
+    const double log_det = log_det_a + log_det_b;
+
+    // Then we need the Mahalanobis distance. To do so we'll draw heavily
+    // on some of the definitions used early in this module.
+    //
+    //   d = y^T K^-1 y
+    //     = y^T (A + Q_ff)^-1 y
+    //     = y^T (A^-1 + A^-1 K_fu S^-1 K_uf A^-1) y
```
pgrgich: Can you point to the definitions used to get this and the next line?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -315,23 +321,100 @@ class SparseGaussianProcessRegression
...
+    //     = y^T (A^-1 + A^-1 P^T B^-1 P A^-1) y
+    //     = y^T A^-1 y + y^T A^-1 P^T B^-1 P A^-1 y
+    //     = y^T (A^-1 y) + (B^{-1/2} P A^-1 y)^T (B^{-1/2} P A^-1 y)
```
pgrgich: `A` and `B` are symmetric, I presume.

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -315,23 +321,100 @@ class SparseGaussianProcessRegression
...
+    Eigen::VectorXd c = sc.B_ldlt.sqrt_solve((sc.P * y_a).eval());
+
+    double log_quadratic = sc.y.transpose() * y_a;
+    log_quadratic -= c.transpose() * c;
+
+    const double rank = static_cast<double>(sc.y.size());
+    const double log_dimension = rank * log(2 * M_PI);
+
+    return -0.5 * (log_det + log_quadratic + log_dimension) +
+           this->prior_log_likelihood();
```
pgrgich: Can you point to the formula used here?

tests/test_samplers.cc
```diff
@@ -133,7 +134,54 @@ TEST(test_samplers, test_samplers_gp) {
   const auto ensemble_samples =
       ensemble_sampler(model, dataset, walkers, max_iterations, gen, callback);
 
-  assert(oss->str().size() > 1);
+  EXPECT_GT(oss->str().size(), 1);
+}
+
+inline long int get_group(const double &f) { return lround(floor(f / 5.)); }
```
seth-swiftnav: Is `lround` being used to cast from `double` to `long int`? Is there a way that doesn't use two rounding math functions?
This also appears in `test_sparse_gp.cc`

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -117,6 +117,8 @@ struct StateSpaceInducingPointStrategy {
  * get those deails straight:
  *
  *     - https://bwengals.github.io/pymc3-fitcvfe-implementation-notes.html
+ *        (beware, the final formulae are correct but there are typos in
+ *         their derivations).
```
seth-swiftnav: Should this comment be dated? The referenced post could be corrected in the future.

include/albatross/src/models/sparse_gp.hpp
akleeman: Yeah exactly.

include/albatross/src/models/sparse_gp.hpp
akleeman: That's because the `LDLT` imposes that the `L` has a unit diagonal.  Maybe I'll break this out into a little helper function.

include/albatross/src/models/sparse_gp.hpp
akleeman: Will do.

tests/test_samplers.cc
akleeman: we could `static_cast<long int>(floor(f / 5.))` ... but casting from doubles to ints always scares me.

include/albatross/src/models/sparse_gp.hpp
akleeman: It's actually given in a comment at the top of the method.

include/albatross/src/models/sparse_gp.hpp
akleeman: yep

include/albatross/src/models/sparse_gp.hpp
akleeman: done

tests/test_samplers.cc
pgrgich: I'm pretty sure an integer is guaranteed to have no inaccuracy in `double` representation, because it can be represented exactly, therefore you can safely cast. @jbangelo could you confirm?

jbangelo: Casting an `int` to a `double` should be safe on most common platforms since the factional part of a `double` is 52-bits, but casting a `double` to an `int` obviously risks losing precision/data.

`long` is a little more complicated because it tends to be 32 bits on 32 bit platforms, and 64 bits on 64 bit platforms. But casting a `double` to a `long` runs the same risks as with `int`.

It should also be noted that `lround()` won't necessarily clamp the return value if the `double` value is outside the range of `long`.

PR #211, opened 2020-02-04
Tune Using Free Functions

akleeman:
This change refactors tuning a bit to allow easier tuning using free functions (and lambdas).

For example, you can now do something like this:

```
  std::vector<double> initial_x = create_initial_guess();
  GenericTuner tuner(initial_x);
  const std::vector<double> tuned_params = tuner.tune([](const std::vector<double> &xs) {return foo(xs);});
```
or
```
  ParameterStore initial_x = create_initial_guess();
  GenericTuner tuner(initial_x);
  const ParameterStore tuned_params = tuner.tune([](const ParameterStore &xs) {return foo(xs);});
```
or
```
  Eigen::VectorXd initial_x = create_initial_guess();
  GenericTuner tuner(initial_x);
  const Eigen::VectorXd tuned_params = tuner.tune([](const Eigen::VectorXd &xs) {return foo(xs);});
```

### DIFF
--- a/include/albatross/src/tune/tune.hpp
+++ b/include/albatross/src/tune/tune.hpp
@@ -34,50 +34,194 @@ inline std::string to_string(const nlopt::result result) {
   return result_strings[result];
 }
 
-/*
- * This function API is defined by nlopt, when an optimization algorithm
- * requires the gradient nlopt expects that the grad argument gets
- * modified inside this function.  The TuneModelConfig which holds
- * any information about which functions to call etc, needs to be passed
- * in through a void pointer.
- */
-template <typename ModelType, typename MetricType, typename FeatureType>
-double objective_function(const std::vector<double> &x,
-                          std::vector<double> &grad, void *void_tune_config) {
-  if (!grad.empty()) {
-    throw std::invalid_argument("The algorithm being used by nlopt requires"
-                                "a gradient but one isn't available.");
-  }
-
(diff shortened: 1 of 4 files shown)


include/albatross/src/tune/tune.hpp
```diff
@@ -34,50 +34,194 @@ inline std::string to_string(const nlopt::result result) {
...
+  optimizer.set_upper_bounds(tunable_params.upper_bounds);
+  // the sensitivity to parameters varies greatly between parameters so
+  // terminating based on change in x isn't a great criteria, we only
+  // terminate based on xtol if the change is super small.
+  optimizer.set_xtol_abs(1e-18);
+  optimizer.set_xtol_rel(1e-18);
+  return optimizer;
+}
+
+inline ParameterStore uniformative_params(const std::vector<double> &values) {
```
seth-swiftnav: `uniformative` should be a word meaning it forms one thing...

akleeman:
@seth-swiftnav there actually is a way to change those by accessing the `optimizer` object directly, ie

```
GenericTuner tuner(initial_x);
tuner.optimizer.set_ftol(1e-12);
tuner.tune(...)
```

include/albatross/src/tune/tune.hpp
akleeman: Oppps, typo it should be `uninformative`

PR #212, opened 2020-02-06
Async Utils

akleeman:
This provides some tools for using `std::async` in conjunction with `apply()`.  It's still a bit experimental, I imagine it'd be easy to use this in a way that isn't thread safe ... but it's a start.

Depends on #217

### DIFF
--- a/include/albatross/src/utils/async_utils.hpp
+++ b/include/albatross/src/utils/async_utils.hpp
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2020 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef INCLUDE_ALBATROSS_SRC_UTILS_ASYNC_UTILS_HPP_
+#define INCLUDE_ALBATROSS_SRC_UTILS_ASYNC_UTILS_HPP_
+
+namespace albatross {
+
+// This method makes sure we don't accidentally call async with the
+// default mode which has some flaws:
+//
+// https://eli.thegreenplace.net/2016/the-promises-and-challenges-of-stdasync-task-based-parallelism-in-c11/
+template <typename F, typename... Ts>
+inline auto async_safe(F &&f, Ts &&... params) {
+  return std::async(std::launch::async, std::forward<F>(f),
+                    std::forward<Ts>(params)...);
+}
+
+template <typename ValueType, typename ApplyFunction,
+          typename ApplyType = typename details::value_only_apply_result<
+              ApplyFunction, ValueType>::type,
+          typename std::enable_if<details::is_valid_value_only_apply_function<
+                                      ApplyFunction, ValueType>::value &&
+                                      std::is_same<void, ApplyType>::value,
+                                  int>::type = 0>
+void async_apply(const std::vector<ValueType> &xs, const ApplyFunction &f) {
+  std::vector<std::future<void>> futures;
+  for (const auto &x : xs) {
+    futures.emplace_back(async_safe(f, x));
+  }
+  for (auto &f : futures) {
+    f.wait();
+    f.get();
+  }
+}
+
+template <typename ValueType, typename ApplyFunction,
+          typename ApplyType = typename details::value_only_apply_result<
+              ApplyFunction, ValueType>::type,
+          typename std::enable_if<details::is_valid_value_only_apply_function<
+                                      ApplyFunction, ValueType>::value &&
+                                      !std::is_same<void, ApplyType>::value,
+                                  int>::type = 0>
+auto async_apply(const std::vector<ValueType> &xs, const ApplyFunction &f) {
+  std::vector<std::future<ApplyType>> futures;
+  for (const auto &x : xs) {
+    futures.emplace_back(async_safe(f, x));
+  }
+
+  std::vector<ApplyType> output;
+  for (auto &f : futures) {
+    f.wait();
+    output.emplace_back(f.get());
+  }
+  return output;
+}
+
+} // namespace albatross
+
+#endif /* INCLUDE_ALBATROSS_SRC_UTILS_ASYNC_UTILS_HPP_ */
--- a/include/albatross/src/utils/async_utils.hpp
+++ b/include/albatross/src/utils/async_utils.hpp
@@ -38,7 +38,6 @@ void async_apply(const std::vector<ValueType> &xs, const ApplyFunction &f) {
     futures.emplace_back(async_safe(f, x));
   }
   for (auto &f : futures) {
-    f.wait();
     f.get();
   }
 }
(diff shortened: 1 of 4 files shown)


tests/test_async_utils.cc
```diff
@@ -0,0 +1,101 @@
...
+  std::vector<int> xs = {0, 1, 2, 3, 4, 5};
+
+  std::mutex mu;
+  int sum = 0.;
+
+  std::vector<int> order_processed;
+
+  auto add_to_sum = [&](const int x) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(abs(x - 2)));
+    mu.lock();
```
jbangelo: It's preferred to use `std::lock_guard<std::mutex>` for locking a mutex.

tests/test_async_utils.cc
```diff
@@ -0,0 +1,101 @@
...
+    sum += x;
+    order_processed.push_back(x);
+    mu.unlock();
+  };
+
+  async_apply(xs, add_to_sum);
+
+  EXPECT_EQ(sum, std::accumulate(xs.begin(), xs.end(), 0));
+  // Make sure the async apply was indeed processed out of order.
+  EXPECT_NE(order_processed, xs);
```
jbangelo: This could randomly fail. `std::this_thread::sleep_for` isn't guaranteed to sleep for exactly as long as you request, it could sleep longer.

include/albatross/src/utils/async_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+                                      ApplyFunction, ValueType>::value &&
+                                      std::is_same<void, ApplyType>::value,
+                                  int>::type = 0>
+void async_apply(const std::vector<ValueType> &xs, const ApplyFunction &f) {
+  std::vector<std::future<void>> futures;
+  for (const auto &x : xs) {
+    futures.emplace_back(async_safe(f, x));
+  }
+  for (auto &f : futures) {
+    f.wait();
```
jbangelo: I don't think this call to `wait()` is needed. `get()` should implicitly wait for the result.

tests/test_async_utils.cc
akleeman: cool yeah that seems much nicer.

tests/test_async_utils.cc
akleeman: Hmmm, any other thoughts on how to test this?

include/albatross/src/utils/async_utils.hpp
akleeman: yep, works without the `wait`

akleeman:
Makes sense about whether this actually speeds things up or not.  The use case I'm aiming for is inside the `SparseGaussianProcess` in which we serially compute large matrix decompositions which could be done in parallel so I feel confident we can see benefits.  I'm not familiar enough with how this works in C++ so I have lots of questions :)  What are the costs of firing up a new thread, I'm guessing it captures via references?  Ie, are there ways to store/compute the matrix decompositions that will lend itself better to `async`?  If we only have 4 cores and call `async` with 100 "calls" will it effectively run at 4x the speed of doing it in serial or are there hidden costs to having too many threads for your CPUs?

tests/test_async_utils.cc
jbangelo: The hacky approach would be to wait until this fails and then increase the order of magnitude of sleep duration. This will also increase how long it takes to run this particular test.

I'd question if you really need to test that they executed out of order. If that's truly something that needs to be tested for it will need to be enforced in someway with synchronization primitives. I don't think the intention of this PR is to enable any kind of prioritization or order of the parallel calculations, so testing for that doesn't make a whole lot of sense.

akleeman: Yeah the out of order comparison was my attempt to test that execution was indeed happening asynchronously.  I suppose the subsequent test (where I test that execution is faster than using a single thread) accomplishes that to some degree.  I'll just remove this test for now.

jbangelo:
Fair warning I'm not a particular expert on concurrent programming. I'll give the best answers I have, but they may be vague in places.  😃 

> What are the costs of firing up a new thread, I'm guessing it captures via references? 

The C++ threading stuff are just abstractions over the platform specific threading libraries. So the exact details vary from OS to OS, and to some extent hardware to hardware. The primary "costs" of spawning new threads are in the system calls to start up and tear down the threads. The OS has to allocate a stack for the new thread and add it to the scheduler and when the thread ends it needs to clean up the allocated resources and notify the other threads that hold a handle the the thread. I don't have a quantifiable minimum amount of "work" that represents a break even point for this.

> If we only have 4 cores and call async with 100 "calls" will it effectively run at 4x the speed of doing it in serial or are there hidden costs to having too many threads for your CPUs?

Roughly speaking yes, though there is also overhead in switching between threads. If your threads are all CPU bound then you'll in theory get a nice linear speed up for every additional thread you make until you have a thread for each CPU core and then you'll see a very small decrease in performance for additional threads. Again I don't have hard numbers (benchmarking is the usual approach I've taken), but I'd say you don't see much impact until the number of threads greatly outnumbers the CPU cores available. 100 threads to 4 cores would probably meet this threshold, but I'm sure it would still be faster than not parallelizing the calculation.

> Ie, are there ways to store/compute the matrix decompositions that will lend itself better to async?

I'm not familiar with the math involved, but if there's any computation that could be reused across matrices it would be preferable to share that with other decompositions. Cross thread communications can be done, but they also have performance implications (if there's a central cache of data that all of the threads are constantly writing to then you've effectively serialized your parallel threads as only one thread can write to shared memory at a time). Ideally there'd be some way to identify ahead of time matrices which would have parts of the decomposition calculation that could be reused and process those similar matrices serially in a thread together, and have little or now cross thread communication.

martin-swift:
Fascinating - so I found this: https://stackoverflow.com/questions/3929774/how-much-overhead-is-there-when-creating-a-thread
I copied that code (second answer) and ran it on my laptop:
```
stiaszny@t480s:~/testcode/thread-creation$ time ./a.out

real    0m4.196s
user    0m1.782s
sys     0m7.598s
```
So the short answer is that it takes about 20 microseconds to create a thread on a `Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz`  
@akleeman and I were talking yesterday about whether it makes to try to use this to paralellize the computations in the SSRG.  My preliminary profiling suggest these computations are about .2 seconds each, so the overhead of thread creation here is trivial (1/10000).  

This could be a game-changer for the SSRG design as we try to cover CONUS.  @benjamin0 @ljbade

jbangelo:
@martin-swift I'd still suggest trying to bound the number of threads created to some reasonable guess at the maximum number of cores you might have. 1000 threads are almost certainly going to be slower than 16 due to the overhead of the OS context switching between them.

We might also want to consider using a more powerful concurrency framework than what the STL provides. `std::async` is a pretty basic abstraction, and we may outgrow it if we need to do much beyond the most simple parallelization discussed here.

martin-swift:
Right now we have 100 grid points; while it remains to be seen, it is likely that the number of points per grid in a full CONUS situation will be closer to 36.  So if we just naively create 36 threads it will be a little bit of a drag on the scheduler, but not too much.  This post implies that a context switch is about half of a thread launch:
https://eli.thegreenplace.net/2018/measuring-context-switching-and-memory-overheads-for-linux-threads/
So let's say 10 microseconds for my machine.  The default scheduler time slice is 100 milli-sec:
```
stiaszny@t480s:~$ cat /proc/sys/kernel/sched_rr_timeslice_ms 
100
```
Again, a factor of 10,000 difference.  So popping off an extra 20 threads should only waste about 200 microseconds every time slice.  Meh.  I think we should only worry about it if we see that it is actually a problem :)  (like if each thread has a huge stack, or something)

martin-swift:
Also 100% agreed that we should not be creating 1000 threads...

benjamin0:
@martin-swift Are you saying that there would be one thread per grid point, not per tile?

martin-swift:
correct - that is the way I currently envision it; experiments will be forthcoming.

martin-swift:
FTR: https://snav.slack.com/archives/CDBCP43BM/p1585247400054200

PR #217, opened 2020-03-23
Apply and Filter Refactor

akleeman:
This moves the `apply` and `filter` methods to be free functions which work with any map like objects.

Note 99% of this is just moved code, the largest change is the need to add a few more template parameters to the free function versions of the formerly class level `apply` and `filter` functions.


include/albatross/src/indexing/apply.hpp
```diff
@@ -54,6 +40,70 @@ auto apply(const std::vector<ValueType> &xs, const ApplyFunction &f) {
...
+
+template <template <typename...> class Map, typename KeyType,
+          typename ValueType, typename ApplyFunction,
+          typename ApplyType = typename details::value_only_apply_result<
+              ApplyFunction, ValueType>::type,
+          typename std::enable_if<details::is_valid_value_only_apply_function<
+                                      ApplyFunction, ValueType>::value &&
+                                      std::is_same<void, ApplyType>::value,
+                                  int>::type = 0>
+auto apply(const Map<KeyType, ValueType> &map, const ApplyFunction &f) {
```
jbangelo: Return type should probably be explicitly `void`

include/albatross/src/indexing/apply.hpp
```diff
@@ -54,6 +40,70 @@ auto apply(const std::vector<ValueType> &xs, const ApplyFunction &f) {
...
+template <
+    template <typename...> class Map, typename KeyType, typename ValueType,
+    typename ApplyFunction,
+    typename ApplyType = typename details::key_value_apply_result<
+        ApplyFunction, KeyType, ValueType>::type,
+    typename std::enable_if<details::is_valid_key_value_apply_function<
+                                ApplyFunction, KeyType, ValueType>::value &&
+                                std::is_same<void, ApplyType>::value,
+                            int>::type = 0>
+void apply(const Map<KeyType, ValueType> &map, const ApplyFunction &f) {
```
jbangelo: Shouldn't these functions be declared `inline`? Probably true for all of them in this file.

include/albatross/src/indexing/filter.hpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Map
+
+template <template <typename...> class Map, typename KeyType,
+          typename ValueType, typename ToKeepFunction,
+          typename std::enable_if<details::is_valid_value_only_filter_function<
+                                      ToKeepFunction, ValueType>::value,
+                                  int>::type = 0>
+auto filter(const Map<KeyType, ValueType> &map, ToKeepFunction to_keep) {
```
jbangelo: `inline`?

include/albatross/src/indexing/apply.hpp
akleeman: Yes!  Nice catch.

include/albatross/src/indexing/apply.hpp
akleeman: good call

PR #220, opened 2020-03-27
Group by async_apply

akleeman:
This adds the `async_apply` methods which work on map-like objects which makes it compatible with `group_by`.


martin-swift:
from slack - I applied this on top of the no-copy branch (rebased on top of the async utils) and there was a little problem; the lambda I was using for the mean won't compile anymore:
```
    auto predict_single_mean = [&](const auto &f) {
      return iono_model_->predict_mean({f})[0];
    };
```
fails with this error:
```
/home/stiaszny/tmp/orion/third_party/orion-engine/src/ssr_generator/ssr_generator.cc:307:44: error: converting to 'const std::vector<orion::SignalArc, std::allocator<orion::SignalArc> >' from initializer list would use explicit constructor 'std::vector<_Tp, _Alloc>::vector(const allocator_type&) [with _Tp = orion::SignalArc; _Alloc = std::allocator<orion::SignalArc>; std::vector<_Tp, _Alloc>::allocator_type = std::allocator<orion::SignalArc>]'
       return iono_model_->predict_mean({f})[0];
```
reverting this change (and my one include) makes the lambda compile again.

akleeman:
Try again, I think I might have fixed the issue.  Basically if you do:
```
std::vector<ValueType> values;
for (const auto &v : values) { ... }
```
`auto` might have been resolving to `std::allocator<ValueType>`, I explicitly made it `ValueType` in the async for loops.

PR #225, opened 2020-04-12
Square Root Solve

akleeman:
Uses `sqrt_solve` and the `SerializableLDLT` instead of `Eigen::LLT` throughout the sparse Gaussian process to improve numerical stability.


akleeman:
@seth-swiftnav I'm going to push this through so I can run an experiment over the weekend, but when you get a chance to review I can follow up with a PR to make any changes.

akleeman:
The tests were all fine, I was running into strange behavior upstream and tracked it down to numerical stability issues which I believe originated here, in particular when the inducing point prior is nearly singular.  Not 100% sure if it's all resolved, but it is definitely improved.

PR #227, opened 2020-04-14
First pass async sparse GP solve

seth-swiftnav:


examples/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ target_link_libraries(sinc_example
   )
 add_custom_target(run_sinc_example
   DEPENDS sinc_example
-  COMMAND sinc_example -input ${PROJECT_SOURCE_DIR}/examples/sinc_input.csv -output ./sinc_predictions.csv -tune
+  COMMAND sinc_example -input ${PROJECT_SOURCE_DIR}/examples/sinc_input.csv -output ./sinc_predictions.csv -tune > sinc_example_output.txt
```
akleeman, line 12: nice

include/albatross/src/core/model.hpp
```diff
@@ -17,6 +17,8 @@ namespace albatross {
 
 using Insights = std::map<std::string, std::string>;
 
+constexpr bool DEFAULT_USE_ASYNC = false;
```
akleeman, line 20: In previous conversation we decided not to put an `async` in model base, I think I'd still prefer to keep it isolated to the sparse GP (which is the only thing that uses it at the moment) and include it in all models once we get a critical mass using it.  Was there an issue that came up that changed that decision?

include/albatross/src/utils/block_utils.hpp
```diff
@@ -302,6 +312,51 @@ inline Eigen::Matrix<_Scalar, _Rows, _Cols> BlockDiagonalLDLT::sqrt_solve(
...
+  return block_to_row;
+}
+
+template <class _Scalar, int _Rows, int _Cols>
+inline Eigen::Matrix<_Scalar, _Rows, _Cols> BlockDiagonalLDLT::async_solve(
+    const Eigen::Matrix<_Scalar, _Rows, _Cols> &rhs) const {
+  assert(cols() == rhs.rows());
+  Eigen::Matrix<_Scalar, _Rows, _Cols> output(rows(), rhs.cols());
+  auto solve_and_fill_one_block = [&](const size_t i, const Eigen::Index row) {
+    const auto rhs_chunk = rhs.block(row, 0, blocks[i].rows(), rhs.cols());
```
akleeman, line 335: nit: I'm curious what the difference between this and directly doing
```
output.block(row, 0, blocks[i].rows(), rhs.cols()) = blocks[i].solve(rhs.block(row, 0, blocks[i].rows(), rhs.cols()));
```
would be.  On one hand I think the type for `rhs_chunk` will be some sort of `Eigen` block view, so it might be pretty lightweight, but there *could* be an advantage to passing it directly into `solve` as an rvalue.  I really don't know if that'll even make a difference though, feel free to skip.

include/albatross/src/core/model.hpp
seth-swiftnav: It makes it much more difficult to add upstream configurability. We perform model_interface lookup by name, without configuration options. Changing configuration afterwards requires compatibility with the NullModel, which is really only possible if more than the SparseGP changed.
An alternative is adding toggling async by adding it to the model name string. A second alternative is passing a config to interface creation, but that will require decloning functionality shared between our 3 main model types in OE.
Really, my choice here was all about the OE interaction.

akleeman: I think I remember, doing it for only the sparse GP makes it so you subsequently need to special case `NullModel` etc if you want to set the async flag for a generic model.  I'm fine with this as is for now and we can consider alternatives down the line.

PR #228, opened 2020-04-15
Switch Sparse GP to the QR Decomposition

akleeman:
This PR addresses a numerical stability issue we noticed with the previous approach to the `SparseGaussianProcess`.  In particular the posterior covariance used to be computed as:
```
  C = L^-T R^T R B^-1 L^-1
```
Notice that there is not way to break that formula into a square root form and as a result what was happening is that as the system becomes ill-conditioned not only does the mean computation become unstable (at some point that's unavoidable regardless of approach) but the predicted covariance also because asymmetric.  This would cause steps downstream to fail since subsequent computations using the covariance might fail hard.

To fix this we switched to the QR based approach proposed in,

[Stable and Efficient Gaussian Process Calculations](http://www.jmlr.org/papers/volume10/foster09a/foster09a.pdf)

An outline of the specifics to our approach was included in code comments, but here is a (much) more detailed description:

[Sparse_Gaussian_Process.pdf](https://github.com/swift-nav/albatross/files/4494112/Sparse_Gaussian_Process.pdf)

@pgrgich

### DIFF
--- a/include/albatross/src/eigen/serializable_ldlt.hpp
+++ b/include/albatross/src/eigen/serializable_ldlt.hpp
@@ -15,6 +15,7 @@
 
 namespace Eigen {
 
+// See LDLT.h in Eigen for a detailed description of the decomposition
 class SerializableLDLT : public LDLT<MatrixXd, Lower> {
 public:
   SerializableLDLT() : LDLT<MatrixXd, Lower>(){};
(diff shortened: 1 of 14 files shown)


include/albatross/src/eigen/serializable_ldlt.hpp
```diff
@@ -52,20 +53,66 @@ class SerializableLDLT : public LDLT<MatrixXd, Lower> {
...
+  /*
+   * Computes the square root of the diagonal, D^{1/2}
+   */
+  Eigen::DiagonalMatrix<double, Eigen::Dynamic> diagonal_sqrt() const {
+    Eigen::VectorXd thresholded_diag = this->vectorD();
+    for (Eigen::Index i = 0; i < thresholded_diag.size(); ++i) {
+      if (thresholded_diag[i] > 0.) {
+        thresholded_diag[i] = std::sqrt(thresholded_diag[i]);
+      } else {
+        thresholded_diag[i] = 0.;
```
seth-swiftnav, line 81: Should these functions be asserting the diagonal is nonnegative? Can we get a zero diagonal? It's valid for this function, but a problem everywhere else in LDLT.

include/albatross/src/utils/linalg_utils.hpp
```diff
@@ -15,6 +15,37 @@
 
 namespace albatross {
 
+inline Eigen::MatrixXd
+get_R(const Eigen::ColPivHouseholderQR<Eigen::MatrixXd> &qr) {
+  // Unfortunately the matrixR() method in Eigen's QR decomposition isn't
+  // actually the R matrix, it's tall skinny matrix whose lower trapezoid
+  // contains internal data, only the upper triangular portion is useful
```
seth-swiftnav, line 22: I guess Eigen never read Clean Code.

tests/test_async_utils.cc
```diff
@@ -83,7 +83,6 @@ TEST(test_async_utils, test_async_apply_map_key_value_function) {
     std::lock_guard<std::mutex> lock(mu);
     sum += x;
     order_processed.push_back(x);
-    std::cout << key << std::endl;
```
seth-swiftnav, line 86: Sorry.

include/albatross/src/eigen/serializable_ldlt.hpp
akleeman: While in theory we should never get a negative diagonal for a PSD matrix numerically we can and this logic amounts to zeroing out any of the numerically negative definitely directions and mimics what Eigen does internally:

https://github.com/libigl/eigen/blob/master/Eigen/src/Cholesky/LDLT.h#L578-L584

We actually assert that the diagonal is positive here:

https://github.com/swift-nav/albatross/blob/master/include/albatross/src/eigen/serializable_ldlt.hpp#L39

but I'm not convinced that's the correct thing to do, we should probably just be warning and continuing but unfortunately we don't have logging setup in albatross yet.

tests/test_async_utils.cc
akleeman: np, looks like you removed it too.

pgrgich:
@akleeman it doesn't look like `Sparse_Gaussian_Process.pdf` is referenced in the comments in the code anywhere. Can that be put in?

akleeman:
@pgrgich I figured I'd update the documentation with a version of the pdf, otherwise (besides adding the pdf directly to the repo) I'm not sure how we'd host a public version of it.  That make sense?

pgrgich:
@akleeman Oh, because `albatross` is publicly available. Hmm, that makes it tricky. I was thinking og putting it in Confluence, but I guess that doesn't work. Ok, sounds good.

akleeman:
https://github.com/swift-nav/albatross/pull/229

PR #232, opened 2020-05-11
Remove LDLT assertion

akleeman:
This removes the assertion that the LDLT is positive definite.  While checking that the minimum diagonal value is positive is certainly a good thing to do we don't want a hard failure when you encounter that situation.  Instead checking `is_valid` and deciding how to deal with it appropriately is left to the user.

### DIFF
--- a/include/albatross/src/eigen/serializable_ldlt.hpp
+++ b/include/albatross/src/eigen/serializable_ldlt.hpp
@@ -20,26 +20,16 @@ class SerializableLDLT : public LDLT<MatrixXd, Lower> {
 public:
   SerializableLDLT() : LDLT<MatrixXd, Lower>(){};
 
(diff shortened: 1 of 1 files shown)


include/albatross/src/eigen/serializable_ldlt.hpp
```diff
@@ -20,26 +20,16 @@ class SerializableLDLT : public LDLT<MatrixXd, Lower> {
...
-    assert_valid();
-  };
+      : LDLT<MatrixXd, Lower>(std::move(ldlt)){};
 
-  void assert_valid() const {
-    if (this->vectorD().minCoeff() <= 0.) {
-      assert(false && "Attempt to compute LDLT of a non PSD matrix");
-    }
-  }
+  void is_valid() const { return this->vectorD().minCoeff() > 0.; }
```
seth-swiftnav, line 32: bool is_valid()

PR #233, opened 2020-05-22
Miscellaneous Helpers

akleeman:
This adds a few changes that have been helpful upstream:

- min/max key and value lookups: `group_by(get_key).apply(get_value).max_key()` will return the key corresponding to the maximum value returned by `get_value`.
- the eigen value printing utilities now do so sorted by contribution to that eigen vector.
- you can extract subsets from datasets which hold variants.
- an equality operator for linear combinations so they can be compared.
- a `deduplicate` method for datasets.

### DIFF
--- a/include/albatross/src/core/dataset.hpp
+++ b/include/albatross/src/core/dataset.hpp
@@ -73,6 +73,28 @@ subset(const RegressionDataset<FeatureType> &dataset,
                                         subset(dataset.targets, indices));
 }
 
+template <typename FeatureType>
+RegressionDataset<FeatureType>
+deduplicate(const RegressionDataset<FeatureType> &dataset) {
+  auto is_unique = [&](std::size_t index) -> bool {
+    for (std::size_t j = index + 1; j < dataset.features.size(); ++j) {
+      if (dataset.features[index] == dataset.features[j]) {
+        return false;
+      }
+    }
+    return true;
+  };
+
+  std::vector<std::size_t> unique_inds;
+  for (std::size_t i = 0; i < dataset.size(); ++i) {
+    if (is_unique(i)) {
+      unique_inds.push_back(i);
+    }
+  }
+
+  return albatross::subset(dataset, unique_inds);
+}
+
 template <typename X>
 inline auto concatenate_datasets(const RegressionDataset<X> &x,
                                  const RegressionDataset<X> &y) {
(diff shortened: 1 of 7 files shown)


include/albatross/src/indexing/group_by.hpp
```diff
@@ -82,20 +82,32 @@ template <typename KeyType, typename ValueType> class GroupedBase {
...
   ValueType first_value() const { return map_.begin()->first; }
 
-  auto min_value() const {
+  // The min entry based on the values
+  auto min() const {
     const auto value_compare = [](const auto &x, const auto &y) {
       return x.second < y.second;
     };
-    return std::min_element(begin(), end(), value_compare)->second;
+    return *std::min_element(begin(), end(), value_compare);
```
seth-swiftnav, line 90: Do we need the `value_compare` method here? It looks like `min_element` might use `<` as a default operator.

include/albatross/src/core/dataset.hpp
```diff
@@ -56,6 +56,28 @@ template <typename FeatureType> struct RegressionDataset {
   group_by(GrouperFunc grouper) const;
 };
 
+template <typename FeatureType>
+RegressionDataset<FeatureType>
+deduplicate(const RegressionDataset<FeatureType> &dataset) {
+  auto is_unique = [&](std::size_t index) -> bool {
```
seth-swiftnav, line 62: Slightly confusing function name for "appears_later".

include/albatross/src/utils/variant_utils.hpp
```diff
@@ -85,6 +85,27 @@ extract_from_variants(const std::vector<variant<VariantTypes...>> &xs,
...
+  std::vector<OutputType> features;
+  for (std::size_t i = 0; i < dataset.size(); ++i) {
+    dataset.features[i].match(
+        [&](const OutputType &f) {
+          indices.emplace_back(i);
+          features.emplace_back(f);
+        },
+        [](const auto &) {});
+  }
+  return RegressionDataset<OutputType>(features,
```
seth-swiftnav, line 105: Interesting. Subset can't be used on the Dataset as a whole, as that would keep the features as variants. Cool function. Took me several reads to understand it.

include/albatross/src/indexing/group_by.hpp
akleeman: Yeah, without `value_compare` it'll try to compare the `std::pair<KeyType, ValueType>` but we want to only compare based on the value.

include/albatross/src/core/dataset.hpp
akleeman: I see, yeah I'll change that.

PR #235, opened 2020-07-01
Sparse Gaussian Process Update

akleeman:
Here we add the ability to update a `SparseGaussianProcessRegression` class with new groups in an online manner.  For example, if given a stream of datasets:
```
data_0
data_1
...
```
you could fit a model to the first:
```
auto fit_model = sparse_gp.fit(data_0);
```
Then incrementally update the model as new groups come in,
```
fit_model.update_in_place(data_i);
```
If a new dataset consists of `n` observations and the model contains `m` inducing points then each update will require approximately `O(n + m)m^2` operations.  (with a little work I think we could get that to `O(nm^2)`).

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -230,6 +230,83 @@ public:
     }
   }
 
+  struct GroupedComponents {
+    BlockDiagonalLDLT A_ldlt;
+    Eigen::MatrixXd K_uu;
+    Eigen::MatrixXd K_fu;
+    Eigen::VectorXd y;
+  };
+
+  template <typename InducingFeatureType, typename FeatureType>
+  GroupedComponents
+  grouped_components(const std::vector<InducingFeatureType> &inducing_features,
+                     const std::vector<FeatureType> &out_of_order_features,
+                     const MarginalDistribution &out_of_order_targets) const {
+
+    const auto indexer =
+        group_by(out_of_order_features, independent_group_function_).indexers();
+
+    const auto out_of_order_measurement_features =
+        as_measurements(out_of_order_features);
+
+    std::vector<std::size_t> reordered_inds;
+    BlockDiagonal K_ff;
+    for (const auto &pair : indexer) {
+      reordered_inds.insert(reordered_inds.end(), pair.second.begin(),
+                            pair.second.end());
+      auto subset_features =
+          subset(out_of_order_measurement_features, pair.second);
+      K_ff.blocks.emplace_back(this->covariance_function_(subset_features));
+      K_ff.blocks.back().diagonal() +=
+          subset(out_of_order_targets.covariance.diagonal(), pair.second);
+    }
+
+    const auto features =
+        subset(out_of_order_measurement_features, reordered_inds);
+    auto targets = subset(out_of_order_targets, reordered_inds);
+
+    this->mean_function_.remove_from(
+        subset(out_of_order_features, reordered_inds), &targets.mean);
+
+    const Eigen::MatrixXd K_fu =
+        this->covariance_function_(features, inducing_features);
+
+    Eigen::MatrixXd K_uu = this->covariance_function_(inducing_features);
+
+    K_uu.diagonal() +=
+        inducing_nugget_.value * Eigen::VectorXd::Ones(K_uu.rows());
+
+    const Eigen::SerializableLDLT K_uu_ldlt = K_uu.ldlt();
+    // P is such that:
+    //     Q_ff = K_fu K_uu^-1 K_uf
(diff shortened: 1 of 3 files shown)


tests/test_sparse_gp.cc
```diff
@@ -204,4 +204,108 @@ TYPED_TEST(SparseGaussianProcessTest, test_likelihood) {
...
+  sparse.set_param(details::inducing_nugget_name(), 1e-3);
+  sparse.set_param(details::measurement_nugget_name(), 1e-12);
+
+  auto groups = dataset.group_by(grouper).groups();
+  const std::size_t original_size = groups.size();
+
+  const auto held_out_pair = groups.first_group();
+  groups.erase(held_out_pair.first);
+
+  ASSERT_LT(groups.size(), original_size);
```
seth-swiftnav, line 262: Any reason to not be more explicit with `ASSERT_EQ(groups.size(), original_size - 1);` instead?

tests/test_sparse_gp.cc
```diff
@@ -204,4 +204,108 @@ TYPED_TEST(SparseGaussianProcessTest, test_likelihood) {
...
+
+  const auto partial_dataset = groups.combine();
+
+  const auto full_dataset =
```
seth-swiftnav, line 266: Isn't this equal to `dataset` from `make_toy_linear_data`? Do we need the second copy?

tests/test_sparse_gp.cc
```diff
@@ -204,4 +204,108 @@ TYPED_TEST(SparseGaussianProcessTest, test_likelihood) {
...
+      albatross::concatenate_datasets(partial_dataset, held_out_pair.second);
+
+  auto full_fit = sparse.fit(full_dataset);
+  auto partial_fit = sparse.fit(partial_dataset);
+
+  // Copy the fit so we can update it in place
+  decltype(partial_fit) updated_fit(partial_fit);
```
seth-swiftnav, line 273: What happens when we try `partial_fit.update_in_place()`? `partial_fit` isn't const.

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -531,22 +517,103 @@ class SparseGaussianProcessRegression
...
 
 private:
+  // This method takes care of a lot of the common book keeping required to
+  // setup the Sparse Gaussian Process problem.  Namely, we want to get from
+  // possibly unordered features to a structured representation
+  // in the form K_ff = A_ff + Q_ff where Q_ff = K_fu K_uu^-1 K_uf and
+  // A_ff is block diagonal and is formed by subtracting Q_ff from K_ff.
+  //
+  template <typename InducingFeatureType, typename FeatureType>
+  void compute_internal_components(
```
seth-swiftnav, line 543: Both uses of this function are immediately followed with
` const auto B_qr = compute_sigma_qr(K_uu_ldlt, A_ldlt, K_fu);`
Does it make sense to combine `compute_sigma_qr` and `compute_internal_components` in some way?

tests/test_sparse_gp.cc
```diff
@@ -204,4 +204,108 @@ TYPED_TEST(SparseGaussianProcessTest, test_likelihood) {
...
+      *std::max_element(dataset.features.begin(), dataset.features.end());
+
+  FixedInducingPoints strategy(min, max, 8);
+  auto sparse =
+      sparse_gp_from_covariance(covariance, grouper, strategy, "sparse");
+  sparse.set_param(details::inducing_nugget_name(), 1e-3);
+  sparse.set_param(details::measurement_nugget_name(), 1e-12);
+
+  auto groups = dataset.group_by(grouper).groups();
+  const std::size_t original_size = groups.size();
```
seth-swiftnav, line 257: `original_size` is clear in context, but could be confused with `dataset.size()` instead of `num_groups`.

tests/test_sparse_gp.cc
akleeman: That'd work fine (and would be the actual work flow) here I just wanted to leave it partially fit so we can make sure it's different than the full prediction.

tests/test_sparse_gp.cc
akleeman: done

tests/test_sparse_gp.cc
akleeman: done

include/albatross/src/models/sparse_gp.hpp
akleeman: There's actually a third use in `_update_impl` which doesn't follow it with `B_qr` which is why that's split out.

tests/test_sparse_gp.cc
akleeman: done

PR #236, opened 2020-07-02
Add update() to FitModel

akleeman:
The primary change here is the addition of `update` methods to `FitModel` types.  If the model type supports an update (through an `_update_impl` method) the resulting fit model will be capable of,
```
auto partial_fit_model = model.fit(dataset_a);
auto fit_model = partial_fit_model.update(dataset_b);
```
which should be equivalent to,
```
auto fit_model = model.fit(concatenate_datasets(dataset_a, dataset_b));
```

Includes some miscellaneous touchups.

### DIFF
--- a/include/albatross/src/indexing/group_by.hpp
+++ b/include/albatross/src/indexing/group_by.hpp
@@ -80,7 +80,7 @@ public:
 
   std::pair<KeyType, ValueType> first_group() const { return *map_.begin(); }
 
-  ValueType first_value() const { return map_.begin()->first; }
+  ValueType first_value() const { return map_.begin()->second; }
 
   // The min entry based on the values
   auto min() const {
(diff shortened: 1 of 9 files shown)


include/albatross/src/indexing/group_by.hpp
```diff
@@ -80,7 +80,7 @@ template <typename KeyType, typename ValueType> class GroupedBase {
 
   std::pair<KeyType, ValueType> first_group() const { return *map_.begin(); }
 
-  ValueType first_value() const { return map_.begin()->first; }
+  ValueType first_value() const { return map_.begin()->second; }
```
seth-swiftnav, line 83: Is there a test that would've caught this sooner?

akleeman: good call.

PR #237, opened 2020-07-22
Rebase Inducing Points

akleeman:
This adds some methods for manipulating sparse Gaussian processes.  In particular it allows you "rebase" the inducing points, this can be used to do something like,
```
fit_model = sparse_gp.fit(dataset_a);
rebased_model = rebase_inducing_points(fit_model, new_inducing_points);
rebased_model.update(dataset_b);
```
For some applications it might make more sense to simply fit everything all at once, `sparse_g.fit(datasets_a_and_b)` or simply perform and update `fit_model.update(dataset_b)` but the rebase and update process may prove very useful when working with time series. 

For example, consider the situation where you have data arriving in real time.  In this situation you'd get `dataset_a` before `dataset_b` and you might want to make a prediction using the data you have so far.
```
fit_model = sparse_gp.fit(dataset_a);
fit_model.predict(something_at_time_a);
```
You then get new data, `dataset_b`, and want to update the model with it but calling `fit_model.update(dataset_b)` might not be ideal because the inducing points used may only span up to time `a` so and using them to predict after `a` might incur loses.  This is where you may then want to rebase before updating.  The entire process might then consist of having a sliding window of inducing points such that you have,
```
// develop estimates of quantities in the time range: [a - T, a] given data up to a
fit_model = sparse_gp.fit(dataset_a)
// shift the model to contain estimates in the time range: [b - T, b] given data up to a
rebased_model = rebase_inducing_points(fit_model, new_inducing_points);
// update the model to contain estimates given data up to b
rebased_model.update(dataset_b);
```

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -649,9 +649,46 @@ auto rebase_inducing_points(
       fit_model.predict(new_inducing_points).joint();
   new_fit.information = new_fit.train_covariance.solve(new_prediction.mean);
 
-  const Eigen::SerializableLDLT P_ldlt(new_prediction.covariance);
+  // Here P is the posterior covariance at the new inducing points.  If
+  // we consider the case where we rebase and then use the resulting fit
+  // to predict the new inducing points then we see that the predictive
+  // covariance (see documentation above) would be,:
+  //
+  //    C = K_zz - Q_zz + K_zz Sigma K_*zz
+  //
+  // We can use this, knowing that at the inducing points K_zz = Q_zz, to
+  // derive our updated Sigma,
+  //
+  //    C = K_zz - K_zz + K_zz Sigma K_**
+  //    C  = K_zz Sigma K_**
+  //    Sigma = K_zz^-1 C K_zz
+  //
+  // And since we need to store Sigma in sqrt form we get,
+  //
+  //    Sigma = (B_z^T B_z)^-1
+  //          = K_zz^-1 C K_zz^-1
+  //
+  //    B_z^T B_z = Sigma^-1
+  //              = K_zz C^-1 K_zz
+  //              = P_z R_z^T R_z P_z^T
+  //
+  // Where the last line above is a reminder that we only need to store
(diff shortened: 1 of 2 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -620,6 +625,82 @@ class SparseGaussianProcessRegression
...
+  // we consider the case where we rebase and then use the resulting fit
+  // to predict the new inducing points then we see that the predictive
+  // covariance (see documentation above) would be,:
+  //
+  //    C = K_zz - Q_zz + K_zz Sigma K_*zz
+  //
+  // We can use this, knowing that at the inducing points K_zz = Q_zz, to
+  // derive our updated Sigma,
+  //
+  //    C = K_zz - K_zz + K_zz Sigma K_**
```
pgrgich, line 662: Why does `K_*zz` become `K_**`? Something to do with being at the induction points?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -620,6 +625,82 @@ class SparseGaussianProcessRegression
...
+  //    C  = K_zz Sigma K_**
+  //    Sigma = K_zz^-1 C K_zz
```
pgrgich, line 664: Is `K_**^-1=K_zz` for some reason?

include/albatross/src/models/sparse_gp.hpp
akleeman: Ah the last `K_**` is a typo.  I use `*` to indicate something you're predicting and in this case I'm predicting at the new inducing points `z`.

include/albatross/src/models/sparse_gp.hpp
akleeman: Ah, nope but I'm missing a `^-1` at the end.  Nice catch.

tao-swift:
LGTM. One comment is that since we know the estimated/predicted orbits, clocks, and biases are stable for a certain amount of time, maybe it is good to wait until all one epoch's from all stations are arrived, and then do the estimation/prediction?

PR #238, opened 2020-07-30
Add/Improve Dataset Utilities

akleeman:
Adds:
-  assertions to the `subset` methods if they're about to index out of range.
-  the ability to do `std::cout << dataset`
-  a method which aligns datasets so they can more easily be compared `align_datasets(&a, &b)`

### DIFF
--- a/include/albatross/src/core/dataset.hpp
+++ b/include/albatross/src/core/dataset.hpp
@@ -96,6 +96,32 @@ deduplicate(const RegressionDataset<FeatureType> &dataset) {
 }
 
 template <typename X>
+inline auto align_datasets(RegressionDataset<X> *x, RegressionDataset<X> *y) {
+
+  std::vector<std::size_t> x_inds;
+  std::vector<std::size_t> y_inds;
+  for (std::size_t i = 0; i < x->size(); ++i) {
+    for (std::size_t j = 0; j < y->size(); ++j) {
+      if (x->features[i] == y->features[j]) {
+        x_inds.push_back(i);
+        y_inds.push_back(j);
+        continue;
+      }
+    }
+  }
+
+  assert(x_inds.size() == y_inds.size());
+
+  if (x_inds.size() == 0) {
+    *x = RegressionDataset<X>();
+    *y = RegressionDataset<X>();
+  } else {
+    *x = subset(*x, x_inds);
+    *y = subset(*y, y_inds);
+  }
+}
+
+template <typename X>
 inline auto concatenate_datasets(const RegressionDataset<X> &x,
                                  const RegressionDataset<X> &y) {
   const auto targets = concatenate_marginals(x.targets, y.targets);
--- a/include/albatross/src/core/dataset.hpp
+++ b/include/albatross/src/core/dataset.hpp
@@ -129,4 +129,26 @@ inline auto concatenate_datasets(const RegressionDataset<X> &x,
 
 } // namespace albatross
 
+template <typename X, typename std::enable_if_t<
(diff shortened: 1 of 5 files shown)


include/albatross/src/core/dataset.hpp
```diff
@@ -95,6 +95,32 @@ deduplicate(const RegressionDataset<FeatureType> &dataset) {
   return albatross::subset(dataset, unique_inds);
 }
 
+template <typename X>
+inline auto align_datasets(RegressionDataset<X> *x, RegressionDataset<X> *y) {
```
seth-swiftnav, line 99: Name of the function is a bit strange to me. "Align" on its own doesn't mean much to me, and the loss of data is important here. "Intersect" would be good.
`aligned_dataset_intersection`? `intersect_datasets`?

include/albatross/src/core/dataset.hpp
```diff
@@ -129,4 +155,26 @@ inline auto concatenate_datasets(const RegressionDataset<X> &x,
 
 } // namespace albatross
 
+template <typename X, typename std::enable_if_t<
+                          albatross::is_streamable<X>::value, int> = 0>
+std::ostream &operator<<(std::ostream &os,
+                         const albatross::RegressionDataset<X> &dataset) {
+  for (std::size_t i = 0; i < dataset.size(); ++i) {
+    os << dataset.features[i] << "    " << dataset.targets.mean[i] << "   +/- "
```
seth-swiftnav, line 163: Is there a separate cout operator definition for the target distribution? Sometimes I print the full covariance, but often I print the mean-diagonal pairing w/o features. It's unclear what would be default. Not important. Writing customized print loops isn't a pain point.

tests/test_core_dataset.cc
```diff
@@ -43,6 +43,64 @@ TEST(test_dataset, test_deduplicate) {
...
+  std::vector<int> features_a = {0, 1, 2};
+  Eigen::VectorXd targets_a = Eigen::VectorXd::Random(features_a.size());
+  RegressionDataset<int> dataset_a(features_a, targets_a);
+
+  std::vector<int> features_b = {3, 4, 5};
+  Eigen::VectorXd targets_b = Eigen::VectorXd::Random(features_b.size());
+  RegressionDataset<int> dataset_b(features_b, targets_b);
+
+  align_datasets(&dataset_a, &dataset_b);
+  EXPECT_EQ(dataset_a.size(), 0);
```
seth-swiftnav, line 101: Should check that `dataset_b` is empty too, as this test doesn't do direct feature comparison between a and b like the others do

include/albatross/src/core/dataset.hpp
akleeman: I was mirroring the vocab used in pandas: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.align.html

I think unless you're very opposed to `align` I'd like to stick to it.  I'd be OK with `align_datasets_on_intersection` though.

include/albatross/src/core/dataset.hpp
akleeman: Ah I see, yeah I could add one for distributions too.  I was thinking of adding tools like `.top()` and `.bottom()` along the lines of the Eigen methods.  It'd be pretty nice to be able to do:
```
std::cout << dataset.top(5) << std::endl;
```
or
```
std::cout << dataset.targets.top(5) << std::endl;
```
I'll add that in a followup

tests/test_core_dataset.cc
akleeman: good catch.

PR #240, opened 2020-08-04
Fix Sparse GP Stability

akleeman:
A handful of fixes encountered while trying to make sure of the new Sparse GP methods downstream.

- Ran into some numerical instability in the `rebase_inducing_points` methods which was fixed by switching from an LDLT to a QR decomposition.
- Fixed a bug in the `.marginal()` prediction method of the sparse gaussian processes.
- Add some print helpers for distributions.

### DIFF
--- a/include/albatross/src/core/dataset.hpp
+++ b/include/albatross/src/core/dataset.hpp
@@ -157,8 +157,8 @@ inline auto concatenate_datasets(const RegressionDataset<X> &x,
 
 template <typename X, typename std::enable_if_t<
                           albatross::is_streamable<X>::value, int> = 0>
-std::ostream &operator<<(std::ostream &os,
-                         const albatross::RegressionDataset<X> &dataset) {
+inline std::ostream &
+operator<<(std::ostream &os, const albatross::RegressionDataset<X> &dataset) {
   for (std::size_t i = 0; i < dataset.size(); ++i) {
     os << dataset.features[i] << "    " << dataset.targets.mean[i] << "   +/- "
        << std::sqrt(dataset.targets.get_diagonal(i)) << std::endl;
@@ -168,8 +168,8 @@ std::ostream &operator<<(std::ostream &os,
 
 template <typename X, typename std::enable_if_t<
                           !albatross::is_streamable<X>::value, int> = 0>
-std::ostream &operator<<(std::ostream &os,
-                         const albatross::RegressionDataset<X> &dataset) {
+inline std::ostream &
+operator<<(std::ostream &os, const albatross::RegressionDataset<X> &dataset) {
   for (std::size_t i = 0; i < dataset.size(); ++i) {
     os << i << "    " << dataset.targets.mean[i] << "   +/- "
        << std::sqrt(dataset.targets.get_diagonal(i)) << std::endl;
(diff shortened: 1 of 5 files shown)


include/albatross/src/core/dataset.hpp
```diff
@@ -157,8 +157,8 @@ inline auto concatenate_datasets(const RegressionDataset<X> &x,
 
 template <typename X, typename std::enable_if_t<
                           albatross::is_streamable<X>::value, int> = 0>
-std::ostream &operator<<(std::ostream &os,
-                         const albatross::RegressionDataset<X> &dataset) {
+inline std::ostream &
```
seth-swiftnav, line 160: What's the motivation behind using `inline` here? I don't know actual guidance, but its used when the time of cumulative function calls costs more than the added compile size... so its best for small functions called from a few place many times when CPU bound.

tests/test_sparse_gp.cc
```diff
@@ -348,37 +348,77 @@ TYPED_TEST(SparseGaussianProcessTest, test_rebase_inducing_points) {
...
+  const auto direct_pred = direct_fit_model.predict(test_features).joint();
 
-  const auto test_shift = Eigen::VectorXd::Ones(shifted_pred.mean.size());
-  EXPECT_LT((shifted_pred.mean - test_shift - full_pred.mean).norm(), 1e-4);
-  EXPECT_LT((shifted_pred.covariance - full_pred.covariance).norm(), 1e-8);
+  EXPECT_LT((direct_pred.mean - iter_pred.mean).norm(), 1e-5);
+  EXPECT_LT((direct_pred.covariance - iter_pred.covariance).norm(), 1e-5);
 }
 
+// TYPED_TEST(SparseGaussianProcessTest, test_shift_inducing_points) {
```
seth-swiftnav, line 391: Delete commented out code if not used.

include/albatross/src/core/dataset.hpp
akleeman: Actually for templated functions like this it's mandatory, without it the linker will find multiple definitions, it worked fine before because I was only ever printing a dataset of a particular type once, but if I had printed it in multiple files I would have hit a multiple definition error in the linker.

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -646,8 +646,13 @@ auto rebase_inducing_points(
       fit_model.get_model().get_covariance()(new_inducing_points);
   new_fit.train_covariance = Eigen::SerializableLDLT(K_zz);
 
-  const JointDistribution new_prediction =
+  JointDistribution new_prediction =
       fit_model.predict(new_inducing_points).joint();
+  // We're going to need to take the sqrt of the new covariance which
+  // could be extremely small, so here we add a small nugget to avoid
+  // numerical instability
+  new_prediction.covariance.diagonal() += Eigen::VectorXd::Constant(
```
tao-swift, line 658: will this have impact on the positive-definite property of the posteriori variance?

akleeman: It's actually helping ensure that the result is positive definite in situations where it may have ended up numerically negative.

PR #243, opened 2020-08-19
Add a Sparse GP specific fit_from_prediction.

akleeman:
Realized that `rebase_inducing_points` was basically just `fit_from_prediction`, so made that official.

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -400,8 +400,65 @@ public:
     return fit;
   }
 
+  template <typename FeatureType>
+  auto fit_from_prediction(const std::vector<FeatureType> &new_inducing_points,
+                           const JointDistribution &prediction_) const {
+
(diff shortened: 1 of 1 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -400,8 +400,65 @@ class SparseGaussianProcessRegression
     return fit;
   }
 
+  template <typename FeatureType>
+  auto fit_from_prediction(const std::vector<FeatureType> &new_inducing_points,
```
seth-swiftnav, line 404: Is there a better variable name than `new_inducing_points` which suggests the application. What do we use for other `fit_from_prediction` methods? `features_`?

akleeman: I actually did change it to features first ... then realized that whatever gets passed in there become the new inducing points ... so switched back to `new_inducing_points` which I think acts as a clarification since it may not be obvious to someone calling `fit_from_prediction` that the inducing points won't come from the inducing point strategy.

PR #244, opened 2020-09-08
Clang Tidy Issue

akleeman:
Avoid `+=` when forming the prediction for sparse GPs which was causing a clang tidy issue.


seth-swiftnav (approved):
looks ok.

PR #255, opened 2020-11-05
Add filter(dataset, to_keep)

akleeman:
This let's us do things like:
```
RegressionDataset<X> dataset = ...;

auto to_keep = [](const X &x) {
  return should_i_keep_x(x);
}

const auto filtered = filter(dataset, to_keep);
```


seth-swiftnav (approved):
lgtm!

PR #256, opened 2020-11-10
Add a helper for getting member variables

akleeman:
I find myself writing a lot of temporary lambda functions with the `group_by`, `apply` and `filter` methods.  This adds a helper macro which, I think, makes the resulting code less verbose.  As illustrated in comments it allows things like:
```
dataset.group_by(GET(foo));
```
instead of
```
auto get_foo = [](const auto &x) { return x.foo; };
dataset.group_by(get_foo);
```

### DIFF
--- a/include/albatross/src/indexing/group_by.hpp
+++ b/include/albatross/src/indexing/group_by.hpp
@@ -22,6 +22,34 @@
 
 namespace albatross {
 
+// This macro can save the verbosity of creating a lambda which
+// is just accessing a member variable.  For example instead of:
+//
+//   auto get_foo = [](const auto &x) { return x.foo; };
+//   dataset.group_by(get_foo);
+//
+// you can do
+//
+//   dataset.group_by(GET(foo));
+//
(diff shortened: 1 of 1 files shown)


jbangelo (approved):
Looks fine. At the risk of overcomplicating things there's a typesafe templating approach as well: https://godbolt.org/z/q4E1Gd

akleeman:
@jbangelo Whoa!  That's pretty awesome.  Though I can't seem to get it to work with embeded types. https://godbolt.org/z/43TEf8

jbangelo:
@akleeman Yeah, that's a limitation of this approach. I think you can only express pointers to members, I don't think there's a concept of a pointer to a member of a member in C++. For nested data like that you would need to do something like
```c++
auto y_getter = make_getter(&Bar::y);
std::cout << y_getter(f.bar) << "\n";
```

Or use the macro approach.

akleeman:
@jbangelo cool thanks, I'll stick with the macro for now and keep the template approach in mind in case we run into issues

PR #258, opened 2020-12-04
Positive Gaussian Prior

akleeman:
This adds an additional hyper parameter prior, the `PositiveGaussianPrior` aka half Gaussian.  The motivation for this is to allow a weakly-informative prior which encourages smaller values without influencing the posterior too much.

The choice of a Guassian distribution rescaled to reflect only positive support comes from [Prior distributions for variance parameters in hierarchical models](http://www.stat.columbia.edu/~gelman/research/published/taumain.pdf
) which suggests:

```
For a noninformative but proper prior distribution, we recommend approximating the
uniform density [...] or a half-normal centered at 0 with standard deviation set to a high
value such as 100. The latter approach is particularly easy to program as a N(0, 100^2)```

### DIFF
--- a/include/albatross/src/core/priors.hpp
+++ b/include/albatross/src/core/priors.hpp
@@ -15,6 +15,7 @@
 
 namespace albatross {
 
+constexpr double LOG_2_ = 0.6931471805599453;
 constexpr double LOG_2PI_ = 1.8378770664093453;
 constexpr double LARGE_VAL = HUGE_VAL;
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: include/albatross/src/core/priors.hpp [hunk 1]
```diff
@@ -15,6 +15,7 @@
 
 namespace albatross {
 
+constexpr double LOG_2_ = 0.6931471805599453;
```
seth-swiftnav, line 18: Is it obvious to you these are the natural base? I'm quite used to LN, or base 2

seth-swiftnav (approved):
lgtm!

include/albatross/src/core/priors.hpp
akleeman: I feel pretty ok about it considering it's the result of `log(2)` (and `log(2 * pi)` below this line).  If you have strong opinions though I could modify the name somehow.

PR #263, opened 2021-02-03
Serialize linear combinations

akleeman:
This adds serializations for the `LinearCombination<X>` helper so if the feature types, `X`, are serializable the `LinearCombination` will be as well.


seth-swiftnav (approved):
lgtm!

PR #267, opened 2021-02-19
Finite Difference Gradients

akleeman:
In this PR we add some methods for computing the finite difference gradient of a function (`compute_gradient`) and then use that method to enable tuning with gradient based optimization methods such as the [Sequential Least Squares](https://nlopt.readthedocs.io/en/latest/NLopt_Algorithms/#slsqp).

This approach isn't ideal.  Finite difference gradients destroy some of the assumptions used by gradient based optimization routines for determining convergence, but in practice this works "pretty well".  The most notable issue I've found happens near optimal solutions.  In these situations a finite difference gradient may return a non zero gradient implying that further optimization can be achieved.  This confuses the algorithms into attempting iterations for which it can't actually find a satisfactory next step, at which point they typically fail.  One work around is to loosen the convergence criteria, which I've done in the `default_gradient_optimizer`, in an attempt to have the algorithm terminate before it gets into a regime in which the imprecise gradients cause problems.

### DIFF
--- a/include/albatross/src/tune/finite_difference.hpp
+++ b/include/albatross/src/tune/finite_difference.hpp
@@ -1,0 +1,95 @@
+/*
+ * Copyright (C) 2021 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+#ifndef INCLUDE_ALBATROSS_SRC_TUNE_FINITE_DIFFERENCE_HPP_
+#define INCLUDE_ALBATROSS_SRC_TUNE_FINITE_DIFFERENCE_HPP_
+
+namespace albatross {
+
+template <typename Function>
+inline std::vector<double>
+compute_gradient(Function f, const std::vector<double> &params, double f_val,
+                 bool use_async = false) {
+
+  std::vector<std::size_t> inds(params.size());
+  std::iota(std::begin(inds), std::end(inds), 0);
+  const double epsilon = 1e-6;
+
+  auto compute_single_gradient = [&](const std::size_t &ind) {
+    std::vector<double> perturbed(params);
+    perturbed[ind] += epsilon;
+    return (f(perturbed) - f_val) / epsilon;
+  };
+
+  if (use_async) {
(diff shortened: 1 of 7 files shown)


include/albatross/src/tune/finite_difference.hpp
```diff
@@ -0,0 +1,95 @@
...
+namespace albatross {
+
+template <typename Function>
+inline std::vector<double>
+compute_gradient(Function f, const std::vector<double> &params, double f_val,
+                 bool use_async = false) {
+
+  std::vector<std::size_t> inds(params.size());
+  std::iota(std::begin(inds), std::end(inds), 0);
+  const double epsilon = 1e-6;
```
seth-swiftnav, line 24: Are there concerns over the same eps being used for parameters of all magnitudes? (ie. some parameters may be large like length scales, while others are small like angular length scales or nuggets)

include/albatross/src/tune/tune.hpp
```diff
@@ -165,11 +193,19 @@ struct GenericTuner {
 
     auto grad_free_objective = [&](const std::vector<double> &x,
                                    std::vector<double> &grad) {
-      assert(grad.size() == 0 &&
-             "Gradient based algorithm used with a gradient free objective");
-
       double metric = objective(x);
 
+      if (grad.size() > 0) {
```
seth-swiftnav, line 198: This same block of code appears on ln154 above.

include/albatross/src/tune/finite_difference.hpp
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  std::vector<double> output(inds.size());
+
+  auto compute_single_sub_gradient = [&](std::size_t i) {
+    double epsilon = 1e-6;
+    const double range =
+        tunable_params.upper_bounds[i] - tunable_params.lower_bounds[i];
+    if (std::isfinite(range)) {
+      epsilon = 1e-8 * range;
```
seth-swiftnav, line 62: How does this interact with logscale parameters where the range may be large, but the value could be small and the search space handled non-uniformly.

include/albatross/src/tune/finite_difference.hpp
akleeman: Yeah this is one of the downfalls of finite differencing.  In a future iteration I could see us using the prior to derive reasonable steps (kind of doing that already with the `1e-8 * range` logic below).  But basically there's not magic number that will work for all problems.

include/albatross/src/tune/finite_difference.hpp
akleeman: This is working in the log transformed space, so an epsilon here translates to an `exp(x + eps)` change in log transformed parameters.  The numerical optimization is also dealing with log transformed parameters though so it should work out.

include/albatross/src/tune/tune.hpp
akleeman: it's not *exactly* the same, they both are filling in a gradient, but the actual gradient computation is different (one calls `compute_gradient` with a `ParameterStore` the other with a `std::vector<double>`.  Similarly one prints out the parameter name along side the debug output, the other only has access to the index.

seth-swiftnav (approved):

PR #271, opened 2021-03-11
Add nlopt submodule.

akleeman:
Swift Navigation has a fork of nlopt which is guaranteed to work with `albatross` instead of relying on compatibility with whichever version is installed locally.


seth-swiftnav (approved):
lgtm!

PR #275, opened 2021-04-06
Detailed Ransac Output

akleeman:
In a recent PR we added the ability to print debugging information out via a compile time flag: https://github.com/swift-nav/albatross/pull/249.  While helpful for debugging this doesn't allow programmatic comparisons of RANSAC output.  For example, if we wanted to prove that RANSAC could be deterministically repeated we'd want to be able to compare the details from one execution of RANSAC to those from another to demonstrate that each step taken was equivalent.

In this PR we add a `RansacIteration<>` class which holds details of results from a single iteration of RANSAC.  The `RansacOutput` struct was then modified to consist of a `best` iteration and a vector of all the iterations which will allow users to inspect details of what went on inside the algorithm.

### DIFF
--- a/include/albatross/src/models/ransac.hpp
+++ b/include/albatross/src/models/ransac.hpp
@@ -15,8 +15,6 @@
 
 namespace albatross {
 
-const bool DEBUG_RANSAC = false;
-
 template <typename GroupKey>
 inline bool accept_all_candidates(const std::vector<GroupKey> &) {
   return true;
@@ -95,9 +93,9 @@ template <typename GroupKey> struct RansacIteration {
   std::vector<GroupKey> candidates;
   std::map<GroupKey, double> inliers;
   std::map<GroupKey, double> outliers;
-  double consensus_metric;
+  double consensus_metric_value;
 
-  RansacIteration() : consensus_metric(NAN){};
+  RansacIteration() : consensus_metric_value(NAN){};
 
   std::vector<GroupKey> consensus() const {
     std::vector<GroupKey> output(candidates);
@@ -109,8 +107,9 @@ template <typename GroupKey> struct RansacIteration {
 
   bool operator==(const RansacIteration &other) const {
     const bool is_same_consensus_metric =
-        ((std::isnan(consensus_metric) && std::isnan(other.consensus_metric)) ||
-         consensus_metric == other.consensus_metric);
+        ((std::isnan(consensus_metric_value) &&
+          std::isnan(other.consensus_metric_value)) ||
+         consensus_metric_value == other.consensus_metric_value);
     return (candidates == other.candidates && inliers == other.inliers &&
             outliers == other.outliers && is_same_consensus_metric);
   }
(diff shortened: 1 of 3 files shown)


include/albatross/src/models/ransac.hpp
```diff
@@ -91,29 +91,52 @@ inline bool ransac_success(const ransac_return_code_t &rc) {
   return rc == RANSAC_RETURN_CODE_SUCCESS;
 }
 
+template <typename GroupKey> struct RansacIteration {
+  std::vector<GroupKey> candidates;
+  std::map<GroupKey, double> inliers;
+  std::map<GroupKey, double> outliers;
+  double consensus_metric;
```
seth-swiftnav, line 98: Would "consensus_score" or "consensus_metric_score" be more intuitive, so that it doesn't sound like a function.

include/albatross/src/models/ransac.hpp
```diff
@@ -186,91 +195,66 @@ ransac(const RansacFunctions<FitType, GroupKey> &ransac_functions,
...
+    const auto fit = ransac_functions.fitter(iteration.candidates);
 
     // Any group that's part of the candidate set is automatically an inlier.
-    std::vector<GroupKey> candidate_consensus = candidate_groups;
-    std::vector<GroupKey> outliers;
-
-    if (DEBUG_RANSAC) {
-      std::cout << "Inliers:" << std::endl;
-    }
+    std::vector<GroupKey> candidate_consensus = iteration.candidates;
```
seth-swiftnav, line 225: This is unused now? The consensus is built up later with `iteration.consensus()` ?

include/albatross/src/models/ransac.hpp
```diff
@@ -186,91 +195,66 @@ ransac(const RansacFunctions<FitType, GroupKey> &ransac_functions,
...
     // which gives us a consensus (set of inliers).
     for (const auto &possible_inlier : groups) {
-      if (!contains_group(candidate_groups, possible_inlier)) {
+      if (!contains_group(iteration.candidates, possible_inlier)) {
         double metric_value =
             ransac_functions.inlier_metric(possible_inlier, fit);
         const bool accepted = metric_value < inlier_threshold;
         if (accepted) {
-          candidate_consensus.emplace_back(possible_inlier);
+          iteration.inliers[possible_inlier] = metric_value;
```
seth-swiftnav, line 234: I think it's a bit better to use `iteration.inliers.insert({possible_inlier, metric_value});` here.
That skips searching for prior existence of the key in the map (at the cost of doing nothing if it exists)

seth-swiftnav (commented):
Overall looks good.
Slightly more memory, but some perks of having iteration information available for direct comparisons between them later.

We do orphan `DEBUG_RANSAC`, unless we want to add a `cout` after the Ransac operation.

include/albatross/src/models/ransac.hpp
akleeman: I changed it to `consensus_metric_value`.

akleeman:
@seth-swiftnav, yeah I removed the `DEBUG_RANSAC` compile time flag in favor of now being able to directly print the output and get the same information.  Ie, you can now do:
```
const auto result = ransac(...)
std::cout << result << std::endl;
```
which should be pretty close to the same info you'd get with `DEBUG_RANSAC`.  But now you could alternatively direct it to something other than `cout`.

seth-swiftnav:
> @seth-swiftnav, yeah I removed the `DEBUG_RANSAC` compile time flag

It looks like the flag wasn't removed. https://github.com/swift-nav/albatross/blob/228a82f91ecd4ab12f78d228de7f28272c9c9fbb/include/albatross/src/models/ransac.hpp#L18

PR #281, opened 2021-04-29
ConditionalGaussian Model

akleeman:


seth-swiftnav (commented):
Code looks good and clean.

A couple of questions from me. Why is this called a Conditional Gaussian?

Current Ransac precomputes the prior distribution for subsetting. Is the major change that this allows for Ransac to be performed on a posterior distribution as well? If so, is the member variable `JointDistribution prior_;` a bit misleading? I don't have a better suggestion.

akleeman:
@seth-swiftnav I named it `ConditionalGaussian` since effectively what it's doing is storing a large multivariate Gaussian and allowing you to ask for `test_inds` [conditional](https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Conditional_distributions) on `train_inds`.

akleeman:
Also agreed about `prior` possibly being misleading if it's actually the posterior from a different model ... but even in that situation I think you could stretch the interpretation to mean "prior to any subsequent conditioning"

PR #282, opened 2021-04-29
Add dataset.subset()

akleeman:
Adds a convenience method to `RegressionDataset<T>` which lets you do:
```
dataset.subset(indices);
```
instead of
```
albatross::subset(dataset, indices);
```
which aligns with the distributions which also support,
```
joint_distribution.subset(indices);
```


seth-swiftnav (approved):
lgtm!

PR #283, opened 2021-04-29
convert exit() to assert()

akleeman:
The use of `exit()` actually exits the program normally which can mask errors, here I switch to `assert()`.

### DIFF
--- a/include/albatross/src/core/parameter_handling_mixin.hpp
+++ b/include/albatross/src/core/parameter_handling_mixin.hpp
@@ -220,7 +220,7 @@ public:
     if (!map_contains(current_params, key)) {
       std::cerr << "Error: Key `" << key << "` not found in parameters: "
                 << pretty_params(current_params);
-      exit(EXIT_FAILURE);
+      assert(false);
     }
   }
 


include/albatross/src/core/parameter_handling_mixin.hpp
```diff
@@ -220,7 +220,7 @@ class ParameterHandlingMixin {
     if (!map_contains(current_params, key)) {
       std::cerr << "Error: Key `" << key << "` not found in parameters: "
                 << pretty_params(current_params);
-      exit(EXIT_FAILURE);
+      assert(false);
```
jbangelo, line 223: By default asserts will become no-ops in release mode, so this won't reliably kill the process. An ugly hack we've used elsewhere is to use the cmake `target_compile_definitions` command to undefine the `NDEBUG` preprocessor macro for a particular target. That's difficult to do with a header only library like albatross, so you should double check the downstream uses of albatross for that hack.

seth-swiftnav (approved):
lgtm.
Out of curiosity, did a grep. This is the only use of `exit()` outside of cmake and ci.

jbangelo (commented):
I'm not sure that that will do what you want without some other hacky changes. Longer term I think we should consider using other ways to signal errors. Asserts in C++ aren't reliable, and are really meant for sanity checks in tests or debugging situations. Returning error codes is the classic approach, though isn't very ergonomic and makes mishandling errors easy. Exceptions are the other primary alternative, and would kind of allow similar cases like this where the process is killed if an exception isn't handled (threads make this a little more difficult). We've been re-examining our ban on exceptions elsewhere in our code, so it might be a good time to reconsider them for albatross as well.

akleeman:
@jbangelo thanks!  Yeah I'm not sure what the right move is here then.  At the moment there are a LOT of asserts in albatross (and  downstream).  I think I'll go ahead with this PR, just to keep everything consistent, but it does sound like we should consider some other options.  I wonder if an approach [along the lines of `Eigen`](https://github.com/swift-nav/eigen/blob/master/Eigen/src/Core/util/Macros.h#L529-L582) would be the way to go in which we would introduce an `albatross_assert` macro which users could override to either assert, exit, or throw an exception?

jbangelo:
@akleeman That sounds good to me. We haven't made any decisions on what we should do in our other code bases, a similar approach to Eigen has been suggested (maybe Eigen was the original inspiration? I'm not entirely sure), so continuing to do the same thing here for now is reasonable. At the end of the day `assert()` is just a preprocessor macro that prints a string to stderr and calls `std::abort()`, so it's reasonable to re-implement it like Eigen does.

PR #284, opened 2021-04-29
Add method to get prior from GPs

akleeman:
This adds a method which makes it easy to get the prior distribution for a set of features from a Gaussian process.  For example,
```
JointDistribution prior = model.prior(features);
```
I anticipate using this in a helper model I'm putting together to improve the GP ransac flow.


seth-swiftnav (approved):
lgtm!

PR #286, opened 2021-04-30
Ransac Uses Conditional Gaussian

akleeman:
This PR replaces a bunch of the logic in the `ransac_gp.hpp` methods which require indexing into matrices / datasets and producing fits / predictions in an efficient manner.  With this change most of that logic is passed off to the `ConditionalGaussian` model which also lets us remove a large number of template parameters and makes it easier to call these ransac methods with intermediate predictions (instead of requiring an unfit model).

### DIFF
--- a/include/albatross/src/models/ransac_gp.hpp
+++ b/include/albatross/src/models/ransac_gp.hpp
@@ -71,9 +71,7 @@ get_gp_ransac_consensus_metric(const ConditionalGaussian &model,
 struct DifferentialEntropyConsensusMetric {
   double operator()(const JointDistribution &pred,
                     const MarginalDistribution &truth) const {
-    Eigen::MatrixXd cov = pred.covariance;
-    cov += truth.covariance;
-    return differential_entropy(cov);
+    return differential_entropy(pred.covariance);
   }
 };
 
--- a/include/albatross/src/models/ransac_gp.hpp
+++ b/include/albatross/src/models/ransac_gp.hpp
@@ -15,147 +15,105 @@
 
 namespace albatross {
 
-template <typename ModelType, typename FeatureType> struct FitAndIndices {
-  using FitType = typename fit_type<ModelType, FeatureType>::type;
+template <typename GroupKey>
+inline typename RansacFunctions<ConditionalFit, GroupKey>::FitterFunc
+get_gp_ransac_fitter(const ConditionalGaussian &model,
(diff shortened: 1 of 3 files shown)


include/albatross/src/models/ransac_gp.hpp
```diff
@@ -15,176 +15,128 @@
...
-          typename ConsensusMetric, typename IsValidCandidateMetric,
-          typename GroupKey>
-inline RansacFunctions<FitAndIndices<ModelType, FeatureType>, GroupKey>
-get_gp_ransac_functions(
-    const ModelType &model,
-    const RegressionDataset<FeatureType> &dataset_with_mean,
+template <typename InlierMetric, typename ConsensusMetric,
+          typename IsValidCandidateMetric, typename GroupKey>
+inline RansacFunctions<ConditionalFit, GroupKey> get_gp_ransac_functions(
+    const JointDistribution &prior, const MarginalDistribution &truth,
```
seth-swiftnav, line 116: Any pros/cons for having  `const ConditionalGaussian &model` instead of `const JointDistribution &prior, const MarginalDistribution &truth,` in this functions arguments?

include/albatross/src/models/ransac_gp.hpp
```diff
@@ -15,176 +15,128 @@
...
 
 struct DifferentialEntropyConsensusMetric {
-  double operator()(const GroupIndices &indices, const MarginalDistribution &,
-                    const Eigen::MatrixXd &cov) const {
-    const auto consensus_cov = symmetric_subset(cov, indices);
-    return differential_entropy(consensus_cov);
+  double operator()(const JointDistribution &pred,
+                    const MarginalDistribution &truth) const {
+    Eigen::MatrixXd cov = pred.covariance;
+    cov += truth.covariance;
```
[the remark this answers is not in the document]
akleeman: good catch!

include/albatross/src/models/ransac_gp.hpp
akleeman: I don't see any critical pros/cons.  My thought was that it's probably best to keep the `ConditionalGaussian` hidden from the user.  In otherwords, I think that for someone using `albatross` the should be familiar with `JointDistribution` and `MarginalDistribution` and should understand how to produce these things, but they may not know about the `ConditionalGaussian` model.

seth-swiftnav (approved):

PR #287, opened 2021-05-25
Fix randint_without_replacement indices

seth-swiftnav:
Fixes a set of three independent bugs in the `randint_without_replacement` function.
```
inline std::vector<std::size_t>
randint_without_replacement(std::size_t n, std::size_t low, std::size_t high,
                            std::default_random_engine &gen) {
```

1) When `low > 0`,  when `n == n_choices`, the `std::iota` started at `0`, returning the indices`[0, ... ,n-1]` and not `[low, ... ,high]`

2) When `low > 0` and  `n > n_choices/2` , the reciprocal problem of omitting n indices is solved, with the result sent to `indices_complement` https://github.com/swift-nav/albatross/blob/cbe68c412725f7fdb9e83a6e80e553bf3cdc207d/include/albatross/src/indexing/subset.hpp#L190
`indices_complement` assumes the input indices are between `[0,n-1]`, but they were called in the range `[low,high]`. When a set difference is performed between [0,n-1] and subset([low,high] many issues can occur, such as the range [0,min(low,n)-1] being always returned. This was then being "corrected" by adding `low` to each element, restoring to the called range albiet wrong behavior.

3) When  `n > n_choices/2`, an off-by-one in the argument `size` passed to `indices_complement` made the largest valid index excluded from the set_difference. If the largest valid index would've been randomly selected by `randint_without_replacement`, it would be dropped from the returned vector, making the vector one smaller than requested.  This off by one returned vector size when requesting a large majority of indices is the error I encountered during development. When expanding the unit test, I encountered (1) and (2) as independent errors.

Test print statements are temporary to show prior and new behavior in CI.

### DIFF
--- a/include/albatross/src/utils/random_utils.hpp
+++ b/include/albatross/src/utils/random_utils.hpp
@@ -27,9 +27,9 @@ randint_without_replacement(std::size_t n, std::size_t low, std::size_t high,
     assert(false);
   }
 
-  if (n == (high - low + 1)) {
+  if (n == n_choices) {
     std::vector<std::size_t> all_inds(n);
-    std::iota(all_inds.begin(), all_inds.end(), 0);
+    std::iota(all_inds.begin(), all_inds.end(), low);
     return all_inds;
   }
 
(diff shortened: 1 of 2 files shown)


akleeman (approved):
nice catch!

PR #291, opened 2021-05-28
Add greedy_tune

akleeman:
This PR adds a new tuning utility `greedy_tune` which uses multi-threading and a very simple optimization technique to get a rough guess at optimal parameters.  It isn't really designed to provide a full replacement for other tuning methods in `albatross` but is instead a way to efficiently get a good "decent" set of parameters.

The algorithm works as follows:

Given a set of `n` parameters and an objective function `f`:

1. pick the first parameter
2. propose `k` guesses for the parameter value based on lower/upper bounds from the prior.
3. asynchronously evaluate the objective at those `k` guesses.
4. choose the value which minimized the objective
5. pick the next parameter and repeat (from step 2).

I've found this method is particularly useful when the objective function takes a very long time (hours) to evaluate.

### DIFF
--- a/include/albatross/src/tune/greedy_tuner.hpp
+++ b/include/albatross/src/tune/greedy_tuner.hpp
@@ -1,0 +1,208 @@
+/*
+ * Copyright (C) 2021 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+#ifndef ALBATROSS_SRC_TUNE_GREEDY_TUNER_HPP_
+#define ALBATROSS_SRC_TUNE_GREEDY_TUNER_HPP_
+
+namespace albatross {
+
+/*
+ * The greedy_tune method is an alternative to the other tuning approaches
+ * in albatross.  It was designed for coarse tuning of computationally
+ * complex objective functions.  The method uses multi-threading to drive
+ * a relatively simple tuning approach which works by:
+ *
+ *   - Picking one of the parameters.
+ *   - Using the range defined by the prior to produce new guesses.
+ *   - Evaluating the objective at all the guesses (one guess per thread).
+ *   - Picking the best parameter value.
+ *   - Repeating for the next parameter.
+ *
+ * There are almost certainly no global convergence guarantees with
+ * this algorithm, but in practice it seems to do a good job of coarsely
+ * tuning the objective.  One strategy would be to start with an iteration
+ * of this greedy approach and follow up with one of the methods in tune.hpp
+ */
+
+namespace details {
+
+inline std::vector<double> query_ratios(std::size_t n) {
+  const double log_min = -5.;
+  const double log_max = -0.1;
+  if (n == 1) {
+    return {log_min};
+  }
+
+  const double step = (log_max - log_min) / (n - 1);
+  std::vector<double> output;
+  for (std::size_t i = 0; i < n; ++i) {
+    output.push_back(std::pow(10, log_min + i * step));
+  }
+
+  return output;
+}
+
+inline std::vector<double> get_queries(double value, double low, double high,
+                                       std::size_t n) {
+
+  if (std::isinf(high)) {
+    high = 1e8;
+  }
+  if (std::isinf(low)) {
+    low = -1e8;
+  }
+
+  const double low_range = value - low;
+  const double high_range = high - value;
+
+  std::vector<double> queries;
+
+  if (value == low) {
+    // only search higher since we're at the lower bound
+    const auto ratios = query_ratios(2 * n);
+    for (std::size_t i = 0; i < ratios.size(); ++i) {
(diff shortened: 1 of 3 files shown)


include/albatross/src/tune/greedy_tuner.hpp
```diff
@@ -0,0 +1,208 @@
...
+ * There are almost certainly no global convergence guarantees with
+ * this algorithm, but in practice it seems to do a good job of coarsely
+ * tuning the objective.  One strategy would be to start with an iteration
+ * of this greedy approach and follow up with one of the methods in tune.hpp
+ */
+
+namespace details {
+
+inline std::vector<double> query_ratios(std::size_t n) {
+  const double log_min = -5.;
```
seth-swiftnav, line 38: Where do these numbers come from? 1e-5 to 1e-.1

tests/test_tune.cc
```diff
@@ -227,6 +227,41 @@ TEST_F(TestTuneQuadratic, test_generic) {
...
+    }
+    const auto params_result =
+        greedy_tune(mahalanobis_distance_params, params, n_threads,
+                    n_iterations, use_async, &output_stream);
+    auto param_vector = get_tunable_parameters(params_result).values;
+    const Eigen::Map<Eigen::VectorXd> eigen_param_output(
+        &param_vector[0], static_cast<Eigen::Index>(param_vector.size()));
+
+    std::cout << output_stream.str() << std::endl;
+    EXPECT_LE((eigen_param_output - truth).array().abs().maxCoeff(), 0.1);
```
seth-swiftnav, line 259: Before tuning, having a check that "fails" the zerod parameters could be good to show the tuning improves results.
ie. `EXPECT_GE((params/initial_x_map - truth).array().abs().maxCoeff(),  1);`

include/albatross/src/tune/greedy_tuner.hpp
```diff
@@ -0,0 +1,208 @@
...
+inline ParameterStore
+greedy_tune(Function evaluate_function, const ParameterStore &params,
+            std::size_t n_threads = 8, std::size_t n_iterations = 10,
+            bool use_async = true, std::ostream *os = &std::cout) {
+
+  static_assert(
+      has_call_operator<Function, ParameterStore>::value,
+      "evaluate_function must have a single ParameterStore argument.");
+
+  assert(n_threads % 2 == 0); // n_threads must be even;
```
seth-swiftnav, line 112: why does n_threads have to be even?

include/albatross/src/tune/greedy_tuner.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+  for (std::size_t iter = 0; iter < n_iterations; ++iter) {
+    for (std::size_t i = 0; i < tunable.names.size(); ++i) {
+
+      tunable = get_tunable_parameters(best_params);
+      auto values =
+          details::get_queries(tunable.values[i], tunable.lower_bounds[i],
+                               tunable.upper_bounds[i], n_queries);
+      if (iter == 0 && i == 0) {
+        // on the very first iteration we need to include the initial params
```
seth-swiftnav, line 146: What makes iteration 1 unique here? As `best_params` and `tunable` change in value iteration to iterations, wouldn't we always want to keep the surviving best value. As is, would this allow an initial hardcoded value to survive for one round, but (assuming the value won iteration 1), only to be excluded from the search if `get_queries` doesn't include the value?

include/albatross/src/tune/greedy_tuner.hpp
```diff
@@ -0,0 +1,208 @@
...
+        values.push_back(tunable.values[i]);
+      }
+
+      const auto get_params = [&](double v) {
+        return details::set_tunable_param(best_params, i, v);
+      };
+
+      const auto params = apply(values, get_params);
```
seth-swiftnav, line 154: Doesn't this collide with the input argument of the initial paramStore?

include/albatross/src/tune/greedy_tuner.hpp
akleeman: This is a rudimentary way of picking which values should get attempted depending on how many threads you have at your disposal.  Perhaps this helps, here is a diagram of the perturbations which will get attempted as a function of `n`:

image

basically, if `n` is one we only try a small perturbation from the current values.  When it's two, we try a small perturbation and then something near the upper bound, and as n increases we fill in with more attempts in the middle.  The log scale is to favor being close to the current best params.

include/albatross/src/tune/greedy_tuner.hpp
akleeman: The algorithm splits the domain to be searched into "below" and "above" the current value and I allocate an equal number of threads to each.

include/albatross/src/tune/greedy_tuner.hpp
akleeman: The algorithm starts without knowledge of the current function value, so on the first iteration we evaluate at all the proposed values AND at the initial one.  The following logic which picks the best params then keeps track of the current best which may mean it just selects the initial parameters (if all the proposed perturbations were worse).  By iteration two we're guaranteed to have a candidate for `best_params` so nothing get's excluded.

An alternative approach would be to just evaluate the function with the initial params at the beginning of the algorithm ... but if the function takes an hour to evaluate that means optimization won't start for at least an hour, this is meant to get the processes going sooner.

include/albatross/src/tune/greedy_tuner.hpp
seth-swiftnav: The code functionality was clear... I was asking why the specific values of e5 and e-.1 ?

akleeman: Ah, yeah the specific choice is kind of arbitrary, but the idea was to get a range that went from a relatively small perturbation (`1e-5`) to a perturbation that moves towards the upper bound (`0.9`).

seth-swiftnav (approved):
lgtm!

PR #298, opened 2021-08-02
Ransac Threshold Changes

akleeman:
This PR fixes what (I think) is a bug in the determination of inliers.  On master we consider a candidate group an inlier if `inlier_metric < threshold` but I think it should be `inlier_metric <= threshold`.  This came up in some experiments in which I would set the threshold for the chi squared percentile metric to `1` thinking that would continue to RANSAC but would never flag any outliers, but I would still occasionally remove outliers.

Similarly the chi squared valid candidate metric was using less than (and had a hardcoded threshold).  I made the threshold configurable.

### DIFF
--- a/include/albatross/src/stats/chi_squared.hpp
+++ b/include/albatross/src/stats/chi_squared.hpp
@@ -63,7 +63,7 @@ inline double chi_squared_cdf(double x, double degrees_of_freedom) {
 
 inline double chi_squared_cdf(const Eigen::VectorXd &deviation,
                               const Eigen::MatrixXd &covariance) {
-  double distance_squared =
+  const double distance_squared =
       covariance.llt().matrixL().solve(deviation).squaredNorm();
   std::size_t n = static_cast<std::size_t>(deviation.size());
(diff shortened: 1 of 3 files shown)


include/albatross/src/stats/chi_squared.hpp
```diff
@@ -63,7 +63,7 @@ inline double chi_squared_cdf(double x, double degrees_of_freedom) {
 
 inline double chi_squared_cdf(const Eigen::VectorXd &deviation,
                               const Eigen::MatrixXd &covariance) {
-  double distance_squared =
+  const double distance_squared =
       covariance.llt().matrixL().solve(deviation).squaredNorm();
   std::size_t n = static_cast<std::size_t>(deviation.size());
```
tao-swift, line 68: nit, this one can also be defined as `const`?

tao-swift (approved):
lgtm! Looks like this change will have few impact on the normal case, but will handle one special case, right?

akleeman:
@tao-swift right I don't expect this to change any behavior except if you  were to set the threshold to the max value.

PR #299, opened 2021-08-02
Allow for Empty Predictions.

akleeman:
This PR allows you do call:
```
std::vector<Foo> foo = {};
model.fit(dataset).predict(foo).joint();
```
without a hard failure.  The result will be an empty version of whichever predict type you've designated.


tao-swift (approved):
lgtm!

PR #300, opened 2021-08-03
Bug Fix: Update Method Updated In Place

akleeman:
@tao-swift discovered a bug in which calling `fit_model.update()` was updating in place instead of copying.  This change adds a missing copy operation and corresponding tests to make sure the original model was unchanged.


tao-swift (approved):
lgtm! One quick question, which you possibly have already answered in the past, what is the difference between `update` and `update_in_place`? From line 311 and line 315, it seems the `updated_pred.mean` is very close to `updated_in_place_pred.mean`, which means the `mean` from `update` and `update_in_place` are the same?

akleeman:
@tao-swift  `update` returns a copy and `update_in_place` returns void (and performs the update on the current model).

PR #307, opened 2021-09-13
Minimum Spanning Tree

akleeman:
This PR adds some basic graphy theory structures (`Graph` and `Edge`) and adds an implementation of [Prim's algorithm](https://en.wikipedia.org/wiki/Prim%27s_algorithm) for solving for the [min/max]imum spanning tree for a given graph.

The implementation is based largely off these two references: [princeton](https://algs4.cs.princeton.edu/43mst/) and [algotree](https://www.algotree.org/algorithms/minimum_spanning_tree/prims_c++/).

For example, consider this small graph:

image

the minimum spanning tree can be found in `albatross` like this,

```
  std::vector<Edge<std::size_t>> edges;
  edges.emplace_back(0, 1, 1.);
  edges.emplace_back(0, 2, 2.);
  edges.emplace_back(1, 2, 3.);

  const auto tree = minimum_spanning_tree(create_graph(edges));
  std::cout << tree << std::endl;
```
Which results in:
```
0 <--> 1 [1]
0 <--> 2 [2]
```
or visually:
image


include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -0,0 +1,174 @@
...
+  friend std::ostream &operator<<(std::ostream &os,
+                                  const Edge<VertexType> &edge) {
+    os << edge.a << " <--> " << edge.b << " [" << edge.cost << "]";
+    return os;
+  }
+
+  VertexType a;
+  VertexType b;
+  double cost;
+  std::size_t id;
```
lloydmaza, line 42: Is this member useful/necessary? Don't see it utilized anywhere. Ostensibly it might be used as a tie-breaker in the equality comparison, but that doesn't seem to be the case.

include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -0,0 +1,174 @@
...
+  double total_cost = 0;
+  while (!queue.empty()) {
+    const Edge<VertexType> next_edge = queue.top();
+    queue.pop();
+
+    const VertexType &next_vertex = next_edge.b;
+
+    if (!contains_vertex(output, next_vertex)) {
+      total_cost += next_edge.cost;
+      add_edge(next_edge, &output);
```
lloydmaza, line 146: I must be missing something - I don't see how this approach maximizes the span cost (unless maybe the edges are implicitly ordered by decreasing cost).

Not necessarily questioning the implementation, however, just wouldn't mind if you could point out what I might not be seeing.

lloydmaza: As a loose bit of related feedback, I find algorithmic/mathematical implementations in C++ _do_ benefit from regular use of comments to explain steps in the procedure. If this isn't the convention you follow in Albatross, then no harm done leaving it as-is.

lloydmaza (approved):
Nice, compact, and clean! Just a couple of comments but nothing to block on - merge away if you like!

Admittedly I did not review the test implementations thoroughly, but I can see where you sourced them from and feel it's safe to trust you on their accuracy.

include/albatross/src/graph/minimum_spanning_tree.hpp
akleeman: Yep, you got it.  The edges are placed in the `priority_queue` which orders them, so `const Edge<VertexType> next_edge = queue.top();` is the next largest edge.

In terms of comments I think I agree that this could benefit from some high level description of what is happening.  Mostly I try to stick to Swift's code style ... but there are other examples of complex parts of albatross which have comments.

include/albatross/src/graph/minimum_spanning_tree.hpp
akleeman: Yeah I left this in here to make it easy to inverse a conversion from some problem to a spanning tree problem.  For example, say you have some graph defined by some more complicated types:
```
std::vector<Foo> foos = get_application_specific_data();
```
You can build an `albatross::Graph` from them:
```
std::vector<albatross::Edge> edges;
for (std::size_t i = 0; i < foos.size(); ++i) {
  edges.push_back(Edge(foo.a, foo.b, cost(foo), i);
};
```
Find the minimum spanning tree,
```
const auto solution = minimum_spanning_tree(create_graph(edges));
```
And then convert back to the problem specific types
```
std::vector<Foo> foo_solution;
for (const auto &edge : solution.edges) {
  foo_solution.push_back(foos[edge.id]);
}
```
This is actually how I'm using this code in our parent repos.

PR #312, opened 2021-11-03
Implementation of Kruskal's algorithm

seth-swiftnav:
Implementation of Kruskal's minimum spanning tree algorithm.
The tracking of added nodes as a set of disjoint trees can be improved using pointer logic.
If Vertices are ordered as indices into a Vector at an early stage, repeated searches for the Vertices in a set can be avoided.

### DIFF
--- a/include/albatross/src/graph/minimum_spanning_tree.hpp
+++ b/include/albatross/src/graph/minimum_spanning_tree.hpp
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
   return output;
 }
 
+// A helper struct to track if a closed loop would be formed by a proposed edge
+// during Kruskal's algorithm. This struct associates each node with a tree,
+// initialized to each node in a unique tree. When a new edge is added, the
+// trees associated with the edge's nodes are merged into one tree. If a
+// proposed edge already has both nodes in the same tree, a closed loop would be
+// formed, and the edge can be rejected.
+template <typename VertexType> struct KruskalNodeTracker {
+
+  struct VertexWithTreeID {
+    VertexWithTreeID(const VertexType &v_, const size_t &tree_)
+        : v(v_), tree(tree_){};
+    VertexType v;
+    size_t tree;
+  };
+
+  KruskalNodeTracker(const std::set<VertexType> &vertices) : vertices_() {
+    size_t tree_id = 0;
+    for (const auto &v : vertices) {
+      vertices_.emplace_back(v, tree_id);
+      tree_id++;
+    }
+  };
+
+  bool in_same_tree_else_add_edge(const VertexType &u, const VertexType &v) {
+    size_t index_u, index_v; // can remove this lookup if vertices are replaced
+                             // with consecutive 0-indexed IDs. This function
+                             // serves two roles to avoid repeated lookups
+    bool found_u = false;
+    bool found_v = false;
+    for (size_t i = 0; i < vertices_.size(); ++i) {
+      if (vertices_[i].v == u) {
+        index_u = i;
+        found_u = true;
+      }
+      if (vertices_[i].v == v) {
+        index_v = i;
+        found_v = true;
+      }
+    }
+    assert(found_u && found_v);
+    if (vertices_[index_u].tree == vertices_[index_v].tree) {
+      return true;
+    }
+
+    size_t old_tree = vertices_[index_u].tree;
+    size_t new_tree = vertices_[index_v].tree;
+    for (size_t i = 0; i < vertices_.size(); ++i) {
+      if (vertices_[i].tree == old_tree) {
+        vertices_[i].tree = new_tree;
+      }
+    }
+    return false;
+  }
+
+  std::vector<VertexWithTreeID> vertices_;
+};
+
+/*
+ * This implementation of Kruskal's algorithm was based largely off the
+ * following reference:
+ *
+ *   https://en.wikipedia.org/wiki/Kruskal%27s_algorithm
+ *
+ * The general idea is that we add edges with the minimum weight until each edge
+ * is included, skipping addition of edges that would form a closed loop.
+ */
+template <typename VertexType>
+Graph<VertexType> minimum_spanning_forest(const Graph<VertexType> &graph) {
+  Graph<VertexType> output;
+
+  if (graph.edges.size() == 0) {
+    return output;
+  }
+
+  auto sorted_graph = graph;
+  std::sort(sorted_graph.edges.begin(), sorted_graph.edges.end());
+  KruskalNodeTracker<VertexType> ks(sorted_graph.vertices);
+  for (const auto &e : sorted_graph.edges) {
+    if (!ks.in_same_tree_else_add_edge(e.a, e.b)) {
+      add_edge(e, &output);
+    }
+  }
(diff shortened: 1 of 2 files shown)


include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
...
+      vertices_.emplace_back(v, tree_id);
+      tree_id++;
+    }
+  };
+
+  bool in_same_tree_else_add_edge(const VertexType &u, const VertexType &v) {
+    size_t index_u, index_v; // can remove this lookup if vertices are replaced
+                             // with consecutive 0-indexed IDs. This function
+                             // serves two roles to avoid repeated lookups
+    bool found_u = false;
```
akleeman, line 203: nit: I think if I'd implemented this I would have tried to use `std::find_if`, something like:
```
auto find_or_assert = [&](const VertexType &x) {
  auto is_x = [&x](const auto &p) { return p.v == x; };
  const auto iter = std::find_if(vertices_.begin(), vertices_.end(), is_x);
  assert(iter != vertices_.end());
  return *iter;
}
const VertexWithTreeID u_and_tree = find_or_assert(u);
const VertexWithTreeID v_and_tree = find_or_assert(v);
```
Then subsequent steps don't need to do lookups:
```
if (u_and_tree.tree == v_and_tree.tree)
  ...
```
but while that's probably more c++ style, I'm not really sure if it's more clear/efficient.

include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
...
+        index_v = i;
+        found_v = true;
+      }
+    }
+    assert(found_u && found_v);
+    if (vertices_[index_u].tree == vertices_[index_v].tree) {
+      return true;
+    }
+
+    size_t old_tree = vertices_[index_u].tree;
```
akleeman, line 220: nit: `const`

include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
...
+
+  KruskalNodeTracker(const std::set<VertexType> &vertices) : vertices_() {
+    size_t tree_id = 0;
+    for (const auto &v : vertices) {
+      vertices_.emplace_back(v, tree_id);
+      tree_id++;
+    }
+  };
+
+  bool in_same_tree_else_add_edge(const VertexType &u, const VertexType &v) {
```
akleeman, line 199: What does the `else_add_edge` part of this mean?

include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
...
+Graph<VertexType> minimum_spanning_forest(const Graph<VertexType> &graph) {
+  Graph<VertexType> output;
+
+  if (graph.edges.size() == 0) {
+    return output;
+  }
+
+  auto sorted_graph = graph;
+  std::sort(sorted_graph.edges.begin(), sorted_graph.edges.end());
+  KruskalNodeTracker<VertexType> ks(sorted_graph.vertices);
```
akleeman, line 252: what's the `s` part of `ks`? Maybe rename it `node_tracker`?

include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
   return output;
 }
 
+// A helper struct to track if a closed loop would be formed by a proposed edge
+// during Kruskal's algorithm. This struct associates each node with a tree,
+// initialized to each node in a unique tree. When a new edge is added, the
+// trees associated with the edge's nodes are merged into one tree. If a
+// proposed edge already has both nodes in the same tree, a closed loop would be
+// formed, and the edge can be rejected.
+template <typename VertexType> struct KruskalNodeTracker {
```
akleeman, line 182: nit: I think our code style would use `class` here instead of `struct` since it's non-passive : https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

include/albatross/src/graph/minimum_spanning_tree.hpp
```diff
@@ -173,6 +173,92 @@ Graph<VertexType> maximum_spanning_tree(const Graph<VertexType> &graph) {
...
+    size_t new_tree = vertices_[index_v].tree;
+    for (size_t i = 0; i < vertices_.size(); ++i) {
+      if (vertices_[i].tree == old_tree) {
+        vertices_[i].tree = new_tree;
+      }
+    }
+    return false;
+  }
+
+  std::vector<VertexWithTreeID> vertices_;
```
akleeman, line 230: `private:`

akleeman (commented):
looks good! A few optional suggestions and minor changes but looks pretty much good to go.

include/albatross/src/graph/minimum_spanning_tree.hpp
seth-swiftnav: It's likely more efficient unless the compiler optimizes away the incrementing counter when using the iterator. The downside is that it performs a separate lookup for each vertex when both can be looked for on the same iteration. But I could see a variation where both are found in one pass, but this is a premature optimization if a better implementation would be to replace vertices with IDs.
```
auto find_or_assert = [&](const VertexType &x, const VertexType &x) {
  auto is_x_or_y = [&x](const auto &p) { return p.v == x || p.v == y; };
  const auto iter_1 = std::find_if(vertices_.begin(), vertices_.end(), is_x_or_y);
  const auto iter_2 = std::find_if(iter_1++, vertices_.end(), is_x_or_y);
  return std::pair(*iter_1,*iter_2)
}
```
I'll try your suggestion.

include/albatross/src/graph/minimum_spanning_tree.hpp
seth-swiftnav: There's two main steps of the algorithm.
(1) Check if a proposed edge forms a closed loop.
   - Yes - ignore the edge.
   - No add edge to MST and (2) merge the associated trees.
  Since (1) and (2) have repeated work, I would like to do them in one function call. But I can modify things to make this call more intuitive.

akleeman: I think the part that confused me is that it says `else_add_edge` but this function doesn't add an edge, that part happens outside this function, right? My understanding of this function is closer to: `is_same_tree_update_if_not`, but yeah I guess you could also split it into parts, things like:
```
const auto pair = find_vertex_and_tree(e.a, e.b);
if (!is_same_tree(pair.first, pair.second)) {
  update_trees(pair.first, pair.second);
  add_edge(e);
}
```

akleeman (approved):
Nice, much easier to follow! Looks like it needs a pass with clang format, but otherwise good to go.

PR #315, opened 2022-02-21
Add with() method to group_by

akleeman:
The `group_by` methods provide a way to take a vector-like quantity (`std::vector`, `albatross::RegressionDataset<>`) and split them apart then perform further manipulation (`.apply(f)`, `.filter(f)`).  For example, you may have a dataset consisting of information about specific people and you may want to compute the average by gender.  In albatross that can already be accomplished:
```
auto get_gender = [](const auto &x) {
  return x.gender;
}
auto compute_mean = [](const auto &one_gender) {
  return mean(one_gender);
};
people.group_by(get_gender).apply(compute_mean);
```

One common pattern that has arisen is the need to access other information within the apply function.  For example, you may have auxiliary information such as age which isn't stored in the `people` dataset, but which you'd also like available inside the apply function.  Previous to this change you could instead use `index_apply` but then it's up to you to apply the appropriate subsets:
```
std::vector<double> ages = get_ages();

auto compute_means = [&](const auto &indices) {
  const auto group_ages = subset(ages, indices);
  const auto group_people = subset(people, indices);
  return means_by_age(group_people, group_ages);
}

people.group_by(get_gender).index_apply(compute_means);
```

With this change you can instead of:

```
auto compute_means = [&](const auto &data_and_ages) {
  return means_by_age(data_and_ages.first, data_and_ages.second);
}

people.group_by(get_gender).with(ages).apply(compute_means);
```
Similarly, if you already have an existing grouped quantity that you'd like to join with, you can now do:
```
auto compute_with = [&](const auto &data_and_life_expactancy) {
  return calculate_something(data_and_life_expactancy.first, data_and_life_expactancy.second);
}

const auto life_expectancy_by_gender = some_other_function();
people.group_by(get_gender).with(life_expectancy_by_gender).apply(compute_with);
```

### DIFF
--- a/include/albatross/src/indexing/group_by.hpp
+++ b/include/albatross/src/indexing/group_by.hpp
@@ -466,6 +466,36 @@ public:
     return output;
   }
 
+  template <typename SubsetableType,
+            typename std::enable_if<
+                details::is_subsetable<SubsetableType>::value, int>::type = 0>
+  Grouped<KeyType, std::pair<ValueType, SubsetableType>>
+  with(const SubsetableType &other) const {
+    assert(parent_.size() == other.size());
+    Grouped<KeyType, std::pair<ValueType, SubsetableType>> output;
+    for (const auto &key_indexer_pair : indexers()) {
+      output.emplace(
+          key_indexer_pair.first,
+          std::make_pair(albatross::subset(parent_, key_indexer_pair.second),
+                         albatross::subset(other, key_indexer_pair.second)));
+    }
+    return output;
+  }
+
+  template <template <typename...> class Map, typename AlreadyGroupedType>
+  Grouped<KeyType, std::pair<ValueType, AlreadyGroupedType>>
+  with(const Map<KeyType, AlreadyGroupedType> &other) const {
+    assert(other.size() == indexers().size());
+    Grouped<KeyType, std::pair<ValueType, AlreadyGroupedType>> output;
+    for (const auto &key_indexer_pair : indexers()) {
+      output.emplace(
+          key_indexer_pair.first,
+          std::make_pair(albatross::subset(parent_, key_indexer_pair.second),
+                         other.at(key_indexer_pair.first)));
+    }
+    return output;
+  }
+
   std::vector<KeyType> keys() const { return map_keys(indexers()); }
 
   std::size_t size() const { return indexers().size(); }
--- a/tests/test_group_by.cc
+++ b/tests/test_group_by.cc
@@ -12,6 +12,7 @@
 
 #include <albatross/Indexing>
 #include <gtest/gtest.h>
+#include <string.h>
 
 namespace albatross {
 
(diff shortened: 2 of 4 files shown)


include/albatross/src/indexing/group_by.hpp
```diff
@@ -466,6 +466,36 @@ template <typename Derived> class GroupByBase {
...
+    return output;
+  }
+
+  template <template <typename...> class Map, typename AlreadyGroupedType>
+  Grouped<KeyType, std::pair<ValueType, AlreadyGroupedType>>
+  with(const Map<KeyType, AlreadyGroupedType> &other) const {
+    assert(other.size() == indexers().size());
+    Grouped<KeyType, std::pair<ValueType, AlreadyGroupedType>> output;
+    for (const auto &key_indexer_pair : indexers()) {
+      output.emplace(
```
seth-swiftnav, line 491: Do we want to assert that the size of each group between `this` and `other`? We'll fail if any group names don't align, but sizes within groups aren't checked. What happens if:
```
fib_group = group_by( fib(10) , number_of_digits)
prime_group = group_by( primes(10) , number_of_digits)
ill_defined = fib_group.with(prime_group)
```

seth-swiftnav (approved):
lgtm!
Does make me wonder on the application. It seems saver to "zip" before using grouped functions

tests/test_group_by.cc
```diff
@@ -589,4 +590,52 @@ TEST(test_groupby, test_group_by_any_all) {
...
+    std::vector<std::string> strings;
+    for (const auto &x : xs) {
+      strings.push_back(std::to_string(x));
+    }
+    return strings;
+  };
+
+  const auto grouped = group_by(fib, number_of_digits);
+
+  const auto as_strings = grouped.apply(to_strings);
```
tao-swift, line 628: nit, maybe move the helper `to_strings` just before where it is used, and change `to_strings` to `double_to_strings`?

include/albatross/src/indexing/group_by.hpp
akleeman: I think we specifically want to allow different sized groups.  For example, it could be that you first want to compute a map `std::map<Gender, double>` which contains the life expectancy by gender.  Then you might have a dataset and you want to compute the percentage of people over that life expectancy.  You'd want to do something like:
```
dataset.group_by(get_gender).with(life_expectancy).apply(percentage_over_expectancy);
```
If you look at the types you have `dataset.group_by(get_gender)` which would be a map from gender to a dataset. and you'd be joining it with a map from gender to a double.  In otherwords the with argument may not even have a `size()`.

tests/test_indexing.cc
```diff
@@ -332,4 +332,10 @@ TEST(test_indexing, test_matrix_symmetric_subset) {
   EXPECT_EQ(symmetric_subset(x, idx), expected);
 }
 
+TEST(test_indexing, test_is_subsetable) {
+  EXPECT_TRUE(bool(details::is_subsetable<std::vector<int>>::value));
+  EXPECT_TRUE(bool(details::is_subsetable<RegressionDataset<int>>::value));
```
tao-swift, line 337: is there a test for this one?

akleeman: I'll need more context, which "one"?

tao-swift: PR lgtm, trying to understand the `is_subsetable` class, under what type will `is_subsetable` be true?

akleeman: `is_subsetable<X>` is looking to see if there's a method with the signature: `X = subset(X, std::vector<std::size_t>)`.  For example, you can subset vectors ([here](https://github.com/swift-nav/albatross/blob/master/include/albatross/src/indexing/subset.hpp#L22)) and you can subset datasets ([here](https://github.com/swift-nav/albatross/blob/master/include/albatross/src/core/dataset.hpp#L73)), but there is not way to subset a `map<int, int>`.

PR #316, opened 2022-02-22
Convenience method for leave one out variance.

akleeman:
This PR adds a method for computing only the variance of a leave one out cross validated prediction. There are already methods for starting with a model and computing things like the leave one out mean and marginal distributions. However, these methods require a model and sometimes all you have is a covariance. This method gives you a way of starting with a covariance matrix and determining how much of each variable is unpredictable. One use case for this would be filtering (ie, reducing to only a set of variables which are mutually explainable).

Here we add a method: `leave_one_out_conditional_variance` which when called with a covariance matrix:
```
const auto var = leave_one_out_conditional_variance(cov);
```
is the equivalent of iteratively solving for the variance when you remove one row/col at a time:
```
Eigen::VectorXd var(cov.rows());
for (Eigen::Index i = 0; i < cov.rows(); ++i) {
  const auto not_i = get_all_but_ith_row_and_col(cov, i);
  const auto cross = get_ith_row_excluding_diag(cov, i);
  var[i] = cov(i, i) - cross * not_i.ldlt().solve(cross.transpose());
}
```
The iterative approach takes O(n^4) time while, thanks to a linear algebra trick, `leave_one_out_conditional_variance` will be O(n^3).

### DIFF
--- a/tests/test_cross_validation.cc
+++ b/tests/test_cross_validation.cc
@@ -166,11 +166,6 @@ TEST(test_crossvalidation, test_leave_one_out_conditional_variance) {
   Eigen::MatrixXd cov = model.get_covariance()(meas, meas);
   cov.diagonal() = cov.diagonal() + dataset.targets.covariance.diagonal();
   const Eigen::VectorXd loo_variance = leave_one_out_conditional_variance(cov);
-
-  const Eigen::SerializableLDLT ldlt(cov);
-  const Eigen::Index n = cov.rows();
-  const Eigen::MatrixXd inv = ldlt.solve(Eigen::MatrixXd::Identity(n, n));
-
   EXPECT_LE((loo_marginal.covariance.diagonal() - loo_variance).norm(), 1e-8);
 }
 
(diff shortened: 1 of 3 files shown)


tests/test_cross_validation.cc
```diff
@@ -153,6 +153,27 @@ TEST(test_crossvalidation, test_heteroscedastic) {
...
+      model.cross_validate().predict(dataset, loo).marginal();
+
+  const auto meas = as_measurements(dataset.features);
+  Eigen::MatrixXd cov = model.get_covariance()(meas, meas);
+  cov.diagonal() = cov.diagonal() + dataset.targets.covariance.diagonal();
+  const Eigen::VectorXd loo_variance = leave_one_out_conditional_variance(cov);
+
+  const Eigen::SerializableLDLT ldlt(cov);
+  const Eigen::Index n = cov.rows();
+  const Eigen::MatrixXd inv = ldlt.solve(Eigen::MatrixXd::Identity(n, n));
```
seth-swiftnav, line 172: Unused. Replaced with `leave_one_out_conditional_variance` call.

seth-swiftnav (approved):
lgtm!

tao-swift (approved):
lgtm!

tests/test_cross_validation.cc
akleeman: Nice catch!  That was left in from me debugging an issue.

PR #319, opened 2022-04-28
Add leave_one_out_conditional method

akleeman:
This PR adds a convenience methods for computing the leave one out cross validated predictions given a prior distribution and truth.  The leave one out predictions are already available when starting with a Gaussian process:
```
  auto model = MakeGaussianProcess().get_model();
  LeaveOneOutGrouper loo;
  const auto loo_marginal = model.cross_validate().predict(dataset, loo).marginal();
```
But sometimes it's useful to be able to do the same sort of computations using a posterior distribution.  For example, you might fit a model to some set of data, make a posterior prediction of a held out set and then want the leave one out predictions starting with the posterior,
```
const JointDistribution posterior = model.fit(large_dataset).predict(held_out.features).joint();
const MarginalDistribiont loo_posterior = leave_one_out_conditional(posterior, held_out.targets);
```

### DIFF
--- a/include/albatross/src/evaluation/cross_validation_utils.hpp
+++ b/include/albatross/src/evaluation/cross_validation_utils.hpp
@@ -131,15 +131,38 @@ cross_validated_scores(const PredictionMetric<Eigen::VectorXd> &metric,
   return cross_validated_scores(metric, folds, predictions.apply(get_mean));
 }
 
-inline Eigen::VectorXd
-leave_one_out_conditional_variance(const Eigen::MatrixXd &covariance) {
+inline Eigen::VectorXd leave_one_out_conditional_variance(
+    const Eigen::SerializableLDLT &covariance_ldlt) {
   // The leave one out variance will be the inverse of the diagonal of the
   // inverse of covariance (that's a mouthful!) For details see Equation 5.12 of
   // Gaussian Processes for Machine Learning
-  return Eigen::SerializableLDLT(covariance)
-      .inverse_diagonal()
-      .array()
-      .inverse();
+  return covariance_ldlt.inverse_diagonal().array().inverse();
+}
+
+inline Eigen::VectorXd
+leave_one_out_conditional_variance(const Eigen::MatrixXd &covariance) {
+  return leave_one_out_conditional_variance(
+      Eigen::SerializableLDLT(covariance));
+}
+
+inline MarginalDistribution
+leave_one_out_conditional(const JointDistribution &prior,
+                          const MarginalDistribution &truth) {
+  // Computes the conditional distribution of each variable conditional on
+  // all others.
+  //
+  // For details see Equation 5.12 of Gaussian Processes for Machine Learning
+
+  Eigen::MatrixXd covariance = prior.covariance;
+  covariance += truth.covariance;
+
+  Eigen::SerializableLDLT ldlt(covariance);
+  const Eigen::VectorXd loo_variance = leave_one_out_conditional_variance(ldlt);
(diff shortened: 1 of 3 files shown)


include/albatross/src/evaluation/cross_validation_utils.hpp
```diff
@@ -131,15 +131,38 @@ cross_validated_scores(const PredictionMetric<Eigen::VectorXd> &metric,
...
+
+inline MarginalDistribution
+leave_one_out_conditional(const JointDistribution &prior,
+                          const MarginalDistribution &truth) {
+  // Computes the conditional distribution of each variable conditional on
+  // all others.
+  //
+  // For details see Equation 5.12 of Gaussian Processes for Machine Learning
+
+  Eigen::MatrixXd covariance = prior.covariance;
```
peddie, line 156: You should be able to avoid an extra copy of the whole matrix by doing

```
Eigen::SerializableLDLT ldlt(prior.covariance + truth.covariance);
```

tests/test_cross_validation.cc
```diff
@@ -169,6 +169,26 @@ TEST(test_crossvalidation, test_leave_one_out_conditional_variance) {
   EXPECT_LE((loo_marginal.covariance.diagonal() - loo_variance).norm(), 1e-8);
 }
 
+TEST(test_crossvalidation, test_leave_one_out_conditional) {
```
peddie, line 172: Do you want to test that you see an outlier assigned appropriate low likelihood when left out?

peddie (approved):

include/albatross/src/evaluation/cross_validation_utils.hpp
peddie: You pointed out that these operators do not exist, but what about [decomposing in-place](https://eigen.tuxfamily.org/dox/group__InplaceDecomposition.html)?
```
Eigen::MatrixXd covariance = prior.covariance;
covariance += truth.covariance;

Eigen::SerializableLDLT<Eigen::Ref<Eigen::MatrixXd>> ldlt(covariance);
```

akleeman: At the moment `Eigen::SerializableLDLT` doesn't have a template parameter (we could change that), but also I think I found a solution, pass in an rvalue to the `ldlt`:
```
Eigen::MatrixXd covariance = prior.covariance;
covariance += truth.covariance;
Eigen::SerializableLDLT<Eigen::Ref<Eigen::MatrixXd>> ldlt(covariance.ldlt());
```

peddie: I'm assuming that template parameter should have been deleted for it to work.

I don't think this allows avoiding the copy, since `.ldlt()` does not operate in-place.  But maybe at this point we should forget about it and just use your original code :)

peddie: It looks like the way to do this is actually
```
Eigen::SerializableLDLT ldlt(prior.covariance + truth.covariance.diagonal());
//                                                              ^^^^^^^^^^^
```
This convinces Eigen to emit a `DiagonalWrapper`, which unlike a `DiagonalMatrix` (which stores its own data), the `DiagonalWrapper` can participate in compile-time expression evaluation.  I haven't verified that this elides the copy, but based on a deep dive into `CoreEvaluator.h` and friends, I think it does, and I'm pretty confident that this is at least as good as the existing code.

peddie: Apparently this does something shady with references and causes a segfault with clang.  Oh well.

PR #320, opened 2022-04-28
Chi Squared Bug in Limits

akleeman:
I discovered a few cases where the chi squared cdf function was numerically unstable.  In particular when evaluating the function for very large `x` (so when asking how far in the tails an extremely large outlier is) the `chi_squared_cdf` should be returning `1` but was returning erratic values:

image

The chi squared cdf relies heavily on the incomplete gamma function and it turns out these instabilities were happening when we start dividing really small values by other really small values.  The denominator in the incomplete gamma is approximating through continuous fractions and small inaccuracies can cause the denominator to be slightly larger than it should be which causes the drops.  Here for example are the numerator and denominator (on log scale) separately.  Notice that for large values of `z` the two are always very nearly identical.

image

If you look at the difference between the numerator and denominator you can see it becomes very unstable:

image

We had some tresholds for when to treat the incomplete gamma as `1`, but those thresholds were on the absolute values of the numerator and denominator and I think it makes more sense to be have them on the difference between the two.

### DIFF
--- a/include/albatross/src/stats/chi_squared.hpp
+++ b/include/albatross/src/stats/chi_squared.hpp
@@ -26,7 +26,7 @@ namespace albatross {
 
 namespace details {
 
-inline double chi_squared_cdf_unsafe(double x, std::size_t degrees_of_freedom) {
+inline double chi_squared_cdf_unsafe(double x, double degrees_of_freedom) {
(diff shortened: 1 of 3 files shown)


peddie (approved):
I'm not familiar enough with the incomplete gamma function to know whether there's a better approach here or whether 1e-12 is good enough, but you have a test . . .

PR #321, opened 2022-04-29
Remove copy of covariance matrix in LOO conditional calculation

peddie:
See [this answer](https://github.com/swift-nav/albatross/pull/319/files/53468b92dc2cdf856e9c2995c0e286c899affe94#r861503228) on the original PR.  I spent way too much time figuring this out . . .


peddie:
Apparently this does something shady with references . . . better abandon.

PR #322, opened 2022-05-02
Move Parameter Handling Logic to Free Functions.

akleeman:
This fixes a bit of tech debt around parameter handling.

As a reminder the `Parameter` is albatross consists of both the `value` of the parameter and a `prior`.  If you have a class which holds parameters you can access them by calling `x.get_params()` which returns a `map<std::string, Parameter>` which maps a parameter name to the value and prior.  When manipulating parameters you don't always care about both the value and the prior.  For example, if you're tuning a model you may only want to set the parameter values.  We have a number of helper methods which make it easy to set only the values (`set_param_values`) or set only the parameters which exist (`set_params_if_exists`).

Before this change to access these helper methods you would inherit from `ParameterHandlingMixin`, implement the two virtual methods (`get_params` and `unchecked_set_param`) and would then be able to use the helper methods in the mixin class (`set_param`, `set_param_values` ... etc). In practice the models / classes which deal with parameters often end up getting wrapped in other interfaces at which point you're forced to either 1) expose all the parameter get/set helpers in the high level interface or 2) write your own equivalents of things like (`set_param_values`, `set_params_if_exists`).

The primary motivation for this PR was to allow reuse of the helper functions without explicitly inheriting from `ParameterHandlingMixin` by breaking out all the class methods into free functions.  You can continue to use the `ParameterHandlingMixin` to do things like:
```
  ClassExtendingParameterHandlingMixin param_handler;
  param_handler.set_param_values({{"foo", 1.}, {"bar", 2.}});
  param_handler.set_param_value_if_exists("foo", 1.);
```
But can now also use free functions:
```
  AnyClassWithGetSetParams param_handler;
  set_param_values({{"foo", 1.}, {"bar", 2.}}, &param_handler);
  set_param_value_if_exists("foo", 1., &param_handler);
```
The free functions use duck typing to determine whether `param_handler` is a valid parameter handler which requires methods with the same signature as the virtual methods in `ParameterHandlingMixin`:
```
  ParameterStore get_params() const;

  void set_param(const ParameterKey &name, const Parameter &param);
```

A couple of other nice things come out of this. The free functions can now deal with multiple parameter handlers in a single call.  This is realized in `set_params_if_exists_in_any(params, &a, &b, ..., &k)` which should simplify a lot of logic in some of the composite classes we have in other swift repositories.


include/albatross/src/core/parameters.hpp
```diff
@@ -202,6 +202,107 @@ inline bool params_are_valid(const ParameterStore &params) {
...
+  }
+
+  if (assert_exists) {
+    std::cerr << "Error: Parameter `" << name << "` not found";
+    ALBATROSS_ASSERT(false && "Attempt to modify a param which doesn't exist");
+  }
+  return false;
+}
+
+inline void set_param(const ParameterKey &name, const Parameter &param,
```
peddie, line 221: You could imagine avoiding these 3 wrappers by returning something like
`std::optional<std::reference_wrapper<Parameter>>` or `Parameter*`.

include/albatross/src/core/parameters.hpp
```diff
@@ -202,6 +202,107 @@ inline bool params_are_valid(const ParameterStore &params) {
...
+                                      ParameterStore *params) {
+  return modify_param(name, [&](Parameter *p) { p->prior = prior; }, params,
+                      false);
+}
+
+inline bool set_params_if_exists(const ParameterStore &input_params,
+                                 ParameterStore *params) {
+  bool all_set = input_params.size() > 0;
+  for (const auto &pair : input_params) {
+    all_set = set_param_if_exists(pair.first, pair.second, params);
```
peddie, line 283: Should this be `&=`?

Or `std::accumulate` with `std::logical_and<>`?

include/albatross/src/core/parameters.hpp
```diff
@@ -202,6 +202,107 @@ inline bool params_are_valid(const ParameterStore &params) {
...
+  }
+  return all_set;
+}
+
+inline bool set_param_values_if_exists(
+    const std::map<ParameterKey, ParameterValue> &param_values,
+    ParameterStore *params) {
+  bool all_set = param_values.size() > 0;
+  for (const auto &pair : param_values) {
+    all_set = set_param_value_if_exists(pair.first, pair.second, params);
```
peddie, line 293: Same as above

include/albatross/src/core/parameters.hpp
akleeman: :scream: nice catch!

include/albatross/src/core/parameters.hpp
akleeman: Can you explain?  We actually don't have `optional` in albatross, it's not part of c++14 and while we could pull in a non standard implementation I've trying to limit dependencies ... but we could make it happen if we need to.

peddie: What I mean is that if you return some sort of reference to the `Parameter` in question, the caller can get this reference and modify the parameter, the value, the key etc. as needed, without the need for multiple similar functions in the API (kind of like `std::map<Key, T>::at()` returns `T&`).

Then the question becomes how to return such a reference when the right parameter exists but also return something meaningful when it does not.  The two types I mentioned are ways to do that; the first has the drawbacks you mentioned, and the second has the drawback that the compiler doesn't really encourage you to check that it's non-`nullptr` before you dereference it.

akleeman: Ah yeah, I think I actually tried that first and you've identified some of the pain points which made me change course.

The API involved:
```
Parameter *get_param_pointer(const ParameterKey &key)
```
instead of:
```
void set_param(const ParameterKey &key, const Parameter &param);
```

What I found was that this greatly simplified all the `if_exists` logic and was almost certainly more computationally efficient ... but the deal breaker was that once a class returns a pointer to a parameter it has no clue if that parameter has been changed and that's critical to some of our applications in which we have parameters which require post processing.  For example, a class which requires a grid, you might parameterize it with a `grid_origin` and `grid_spacing` parameters, but internally you'd store a `std::vector<double> grid_points`.  With the `set_param` logic you give the implementing class the opportunity to update the grid points when a parameter has changed, with the `get_param_pointer` approach you have no clue when (or if) the parameter has changed.

Here's the branch if you're curious: https://github.com/swift-nav/albatross/compare/9479ecd2a7608530d5b0e3ed8e71c8b6c2237582...akleeman/param_free_functions_get_ptr

peddie: The need to trigger further processing on changes seems like a great reason to stick with the current API.  Thanks for clarifying.

peddie (approved):
Looks good.

PR #323, opened 2022-05-02
Improve Assertion Handling

akleeman:
The use of `assert` can be complicated.  The macro is disabled in release mode which can cause a discrepancy between what gets evaluated in debug mode and release mode.  This can also lead to compilation errors in release mode when the only use of a variable is in any assert statement (so when it get's disabled the variable appears unused).

Here we introduce a new macro which follows the advice in: https://web.archive.org/web/20201129200055/http://cnicholson.net/2009/02/stupid-c-tricks-adventures-in-assert/.

The new macro makes sure even in release mode an expression will get evaluated and avoids the unused variable problem.

### DIFF
--- a/include/albatross/src/cereal/eigen.hpp
+++ b/include/albatross/src/cereal/eigen.hpp
@@ -62,7 +62,7 @@ inline void load(Archive &ar, Eigen::Matrix<_Scalar, _Rows, _Cols> &m,
   const std::string decompressed =
       gzip::decompress(payload.data(), payload.size());
 
-  assert(size_in_bytes == decompressed.size());
+  ALBATROSS_ASSERT(size_in_bytes == decompressed.size());
 
   std::vector<_Scalar> decoded_data(size);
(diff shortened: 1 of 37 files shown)


jbangelo (commented):
I hate to rain on your parade, but I think wrapping the expression up in a `sizeof()` [isn't strictly correct](https://en.cppreference.com/w/cpp/language/expressions#Unevaluated_expressions), it seems like the compiler doesn't have to execute that wrapped expression at runtime if it doesn't want to. Why wasn't it possible to just have `(void) x;`?

akleeman:
Hmm yeah @jbangelo I think it's actually the opposite.  A closer re-read of this: https://web.archive.org/web/20201129200055/http://cnicholson.net/2009/02/stupid-c-tricks-adventures-in-assert/ and it looks like they added `sizeof` explicitly so that the expression is **not** evaluated.

Indeed, removing `sizeof` seems to work just fine.

peddie (approved):
Seems fine to me.

PR #328, opened 2022-05-09
Ability to Align Datasets With Custom Compare

akleeman:
This adds an optional argument to `align_datasets(x, y)` which lets you define your own custom equality comparator.  For example, you might have two datasets which each represent data from different times.  If you want to align the two you then might do something like:
```
RegressionDataset<X> data_prev = dataset_from_previous_time();
RegressionDataset<X> data_next = dataset_from_next_time();

auto equal_except_for_time = [](const auto &x, const auto &y) {
  return x.not_time == y.not_time;
};

align_datasets(&data_prev, &data_next, equal_except_for_time);
```

### DIFF
--- a/include/albatross/src/core/dataset.hpp
+++ b/include/albatross/src/core/dataset.hpp
@@ -105,14 +105,15 @@ deduplicate(const RegressionDataset<FeatureType> &dataset) {
   return albatross::subset(dataset, unique_inds);
 }
 
-template <typename X>
-inline auto align_datasets(RegressionDataset<X> *x, RegressionDataset<X> *y) {
+template <typename X, typename EqualTo>
+inline auto align_datasets(RegressionDataset<X> *x, RegressionDataset<X> *y,
+                           EqualTo equal_to) {
 
   std::vector<std::size_t> x_inds;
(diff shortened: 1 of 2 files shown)


peddie (commented):
I don't see any major problem, but I have questions:

1. Why not just ask the user to customize `operator==()`?
2. If you default to `std::equal_to<>`, couldn't the user specialize that instead and avoid the extra argument?

akleeman:
@peddie yeah, mostly I think defining `operator==()` on the classes involved is the way to go, but consider the case in the example.  If you have a class like this:
```
class X {
 public:
  bool operator==(const X &other) const {
    return (foo == other.foo && time == other.time);
  }
  
  Foo foo;
  Time time;
}
```
Notice that you might want the `operator==()` to be strict equality, but you also might want to align two datasets ignoring time:
```
auto equal_except_for_time = [](const auto &x, const auto &y) {
  return x.foo == y.foo;
};

align_datasets(&data_prev, &data_next, equal_except_for_time);
```
I think for the same reasoning you may not want to specialize `equal_to<X>` (which may have other consequences).

peddie (approved):
Sounds good.

PR #331, opened 2022-05-31
[DO NOT MERGE] Use SPQR for sparse GP solves

peddie:
On a few hundred typical problems ~8000x1000 with a fill of ~0.3, I observed a median speedup of over 5x when using a single thread.  4 threads gave a speedup of over 7x.


peddie:
See https://github.com/swift-nav/albatross/pull/453

PR #345, opened 2022-08-17
Bump clang format to 14 [BUILD-269]

krisukox:
I added another job in the workflow so that I can use ubuntu 22.04.


isaactorz (commented):
LGTM but lets get a look at these from @akleeman

akleeman (approved):
lgtm!

PR #351, opened 2022-09-08
Improve documentation around GPs

akleeman:
I don't really know a good way of showing a preview of the resulting latex renderings (which are probably hard to review directly in the markdown. So here are some screen shots to help.  Alternatively you should be able to checkout this branch, move in to the `doc` directory and run `make html` (it'll require python with sphinx).

image
image
image
image
image
image

### DIFF
--- a/doc/src/datasets.rst
+++ b/doc/src/datasets.rst
@@ -8,7 +8,7 @@ Datasets
 Regression Datasets
 -------------------
 
-One of the core data types in ``albatross`` is the ``RegressionDataset``.  These objects are built to store a set of measurements which we use all over albatross.  There are a number of additional helper methods in the ``RegressionDataset``, but you can think of it as a simple struct:
+One of the core data types in ``albatross`` is the ``RegressionDataset``.  These objects are built to store a set of measurements (or predictions) and are used throughout albatross.  There are a number of additional helper methods in the ``RegressionDataset``, but you can think of it as a simple struct:
 
 .. code-block:: c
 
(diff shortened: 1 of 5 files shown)


peddie (commented):
Everything here looks fine to me, but I'll wait for someone less familiar with the material to approve.

wlanghans (approved):
Really intuitive and easy to follow. Just a few remarks that you can consider. 
I haven't seen the notation you use before. Usually y=f is assumed and noise is implied, or it it written as y = f(x) + eps as you describe and from then on y is used instead of f. At the end you drop f and say f=y anyway. The way I'm used to seeing this is by writing distributions like P(y*|y,x,x*).
That said, since familiarity with Gaussian process regression is assumed at the top, everyone will get it the way it is.
The LDL' decomposition is missing a "T" for transpose when it's introduced for the first time. I'm not sure what the "P" matrix stands for. Maybe a quick mention or reference would be helpful.
I like that you add the code snippet for the fit. It would be cool to have this example code also for the various predictions.

akleeman:
@wlanghans good catch, I fixed the LDLT description.

I also (hopefully) clarified the difference between `f` and `y`. I agree that `f` is often dropped ... but there is a difference. In English I would write `f^* | f = y` as "the process at some new locations given some measurements of the process" and `f^* | y` "the process at some new locations given some measurements".  Basically I think it's reasonable to just say "measurements" instead of "measurements of the process", the process is implied.

Things like `y* | y` start to feel wrong though. I think it's OK to write `y^* | y` if it's clear that you're using `y` to make predictions of some hypothetical measurements `y^*`, but `y^*` isn't _really_ a measurement, it's a prediction of a possible measurement.  For example, what if you're evaluating a model, so you fit a model with some measurements `y` and make a prediction of a held out group of measurements, `y^*`.  Then say you want to compute the error, what would that be? `y^* - y^*|y`?  This is where I think `f^*` helps dis-disambiguate, it's much more obvious if you write `y^* - f^*|y`.

I also think it helps to keep a mental picture of what's happening which involves a process, `f`, which is some nebulous unknown thing that you're trying to understand and you do so by taking measurements of that process, `y`, which help (but don't perfectly) define what the process must be at certain locations.

akleeman:
I'm not really sure it's worth being religious about this distinction between a process and measurements though.  For example, in [Gaussian Processes for Machine Learning](https://gaussianprocess.org/gpml/chapters/RW2.pdf) there are points where they also use `f` to be explicit that we're dealing with a function, or process:
image
a few paragraphs later there's a similar distinction between measurements with noise versus the process:
image
so they seem to happily re-use `y` as both the random variable described by some prior and the real valued vector used in the posterior mean computation :shrug: which is what I was hoping to distinguish between.

sort of related (from that book):
image

wlanghans:
In your example of residual computation
`y^* - y^*|y`
you could distinguish between two sets of observations `y_train` and `y_test`. The residual would then be 
`y_test - y^*|y_train` whereby `y` is used for all observations and asterisk denotes predictions. 
More importantly, I think in the end everyone will get what you mean no matter how you write it. From that perspective I thought it would be easier to keep it simple.

PR #357, opened 2022-10-04
Stricter Composition Rules

akleeman:
This PR handles a subtle unintended behavior in the templating logic which deals with sums, products and scalings of covariance functions. On master we end up defining far more templated methods than we technically need which can lead to some confusing results if you try to ask which methods are supported by a composition of covariance functions.  Examples should help.

Consider the case where you define two covariance functions,
```
class Foo : public CovarianceFunction<Foo> {
public:
  double _call_impl(const X &, const X &) const;
};

class Bar : public CovarianceFunction<Bar> {
public:
  double _call_impl(const X &, const X &) const;
};
```
Notice that they are each defined (only) for type `X`.  Now you can instantiate two and sum them,
```
Foo foo;
Bar bar;
const auto sum = foo + bar;
```
The expected outcome _ideally_ would be the equivalent of:
```
class Sum : public CovarianceFunction<Sum> {
public:
  double _call_impl(const X &a, const X &b) const {
    return foo(a, b) + sum(a, b);
  }
};
```
but on master you'd actually be able to evaluate `_call_impl` with composite types such as:
```
LinearCombination<X> combo;
variant<X, Y> variant;
sum._call_impl(combo, variant);
```
This isn't really a bug, we ultimately do want to be able to call `sum` with these types, but I don't think we want to define the `_call_impl` for these types and would rather leave it to the `CovarianceFunction::call` logic to realize you can do so.  In otherwords we want to be able to do:
```
sum(combo, variant);
```
but don't want to be able to do:
```
sum._call_impl(combo, variant);
```
(not that a user should never be using `_call_impl`, but even internal to albatross I don't think we want to be able to do so).

### DIFF
--- a/include/albatross/src/covariance_functions/callers.hpp
+++ b/include/albatross/src/covariance_functions/callers.hpp
@@ -488,6 +488,35 @@ template <typename CovFunc, typename... Args>
 class has_valid_caller
     : public caller_has_valid_call<DefaultCaller, CovFunc, Args...> {};
 
+/*
+ * This defines a helper trait which indicates whether a call to a
+ * function has an equivalent call. This is different from the valid
+ * caller which might do more complicated operations such as unpacking
+ * a variant, or integrating over linear combinations, here we just want
+ * to know if there is a call method available for types such as
+ * the Measurement<> wrapper for which we'd happily unwrap the type.
+ *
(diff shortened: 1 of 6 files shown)


peddie (approved):
I can't say I understand 100% of the test cases, but I learned a lot trying to understand this patch.

PR #358, opened 2022-10-04
Add Nugget Covariance Function

akleeman:
We've used this `Nugget` covariance function in packages which depend on albatross, but it seems like it should live directly in albatross.

The only real difference between this and the `IndependentNoise<X>` covariance function is that the nugget applies across the board to all features, while the `IndpendentNoise<X>` only evaluates non-zero for `X`.

### DIFF
--- a/include/albatross/src/covariance_functions/traits.hpp
+++ b/include/albatross/src/covariance_functions/traits.hpp
@@ -226,6 +226,25 @@ struct has_valid_variant_mean_caller<U, Caller, variant<A, Ts...>,
       has_valid_variant_mean_caller<U, Caller, variant<Ts...>>::value;
 };
 
+/*
+ * Sometimes we need to distinguish between raw types for which we would
+ * explicitly write covariance function _call_impl methods and wrapped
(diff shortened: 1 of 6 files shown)


include/albatross/src/covariance_functions/traits.hpp
```diff
@@ -226,6 +226,25 @@ struct has_valid_variant_mean_caller<U, Caller, variant<A, Ts...>,
       has_valid_variant_mean_caller<U, Caller, variant<Ts...>>::value;
 };
 
+/*
```
peddie, line 229: Does this belong in #357?

peddie (approved):

include/albatross/src/covariance_functions/traits.hpp
akleeman: It's only used in this PR, but it's true they address very similar issues.  I didn't want the templating:
```
  template <typename X>
  double _call_impl(const X &x, const X &y) const;
```
to capture any arbitrary type, only those which are "basic", it not composites of multiple types.

jbangelo (approved):

PR #359, opened 2022-10-08
Transform Vectors and Datasets

akleeman:
This PR adds a number of operators which let you do things like:

```
Eigen::MatrixXd A;
RegressionDataset<Foo> dataset;
RegressionDataset<LinearCombination<Foo>> = A * dataset;
```
and handles the case where `Foo` is itself a `LinearCombination`:
```
Eigen::MatrixXd A;
RegressionDataset<LinearCombination<Foo>> dataset;
RegressionDataset<LinearCombination<Foo>> = A * dataset;
```
similarly you can transform vectors:
```
Eigen::MatrixXd A;
std::vector<Foo> foos;
std::vector<LinearCombination<Foo>> = A * foos;
```

the dense version would lead to some very large `LinearCombination<>` features, so sparse matrices are supported (and dense matrices are converted to sparse representations first).


include/albatross/src/core/linear_combination.hpp
```diff
@@ -74,6 +74,63 @@ inline auto difference(const X &x, const Y &y) {
...
+
+  for (Eigen::Index k = 0; k < matrix.outerSize(); ++k) {
+    for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
+      output[cast::to_size(it.row())].values.push_back(
+          features[cast::to_size(it.col())]);
+      output_coefs[cast::to_size(it.row())].push_back(it.value());
+    }
+  }
+
+  for (std::size_t i = 0; i < output_coefs.size(); ++i) {
```
peddie, line 92: Can we just output directly into the coefficients of the `LinearCombination`s above instead of doing another copy here?  Might need to reserve the right size for the coefficient vectors ahead of time?

include/albatross/src/core/linear_combination.hpp
```diff
@@ -74,6 +74,63 @@ inline auto difference(const X &x, const Y &y) {
...
+template <typename Derived, typename X>
+inline auto operator*(const Eigen::SparseMatrixBase<Derived> &matrix,
+                      const std::vector<X> &features) {
+  std::vector<LinearCombination<X>> output(cast::to_size(matrix.rows()));
+
+  std::vector<std::vector<double>> output_coefs(cast::to_size(matrix.rows()));
+
+  for (Eigen::Index k = 0; k < matrix.outerSize(); ++k) {
+    for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
+      output[cast::to_size(it.row())].values.push_back(
```
peddie, line 86: nit: might be easier to read if you define separate variables for `std::vector` indices.

include/albatross/src/core/linear_combination.hpp
```diff
@@ -74,6 +74,63 @@ inline auto difference(const X &x, const Y &y) {
...
+                      const std::vector<LinearCombination<X>> &features) {
+  std::vector<LinearCombination<X>> output(cast::to_size(matrix.rows()));
+
+  std::vector<std::vector<double>> output_coefs(cast::to_size(matrix.rows()));
+
+  for (Eigen::Index k = 0; k < matrix.outerSize(); ++k) {
+    for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
+      const auto &combo = features[cast::to_size(it.col())];
+      for (Eigen::Index i = 0; i < combo.coefficients.size(); ++i) {
+        output[cast::to_size(it.row())].values.push_back(
```
peddie, line 111: This statement doesn't require iteratively pushing, right?  can't we just copy `combo.values` outside this loop?

include/albatross/src/core/linear_combination.hpp
```diff
@@ -74,6 +74,63 @@ inline auto difference(const X &x, const Y &y) {
...
+            combo.values[cast::to_size(i)]);
+        output_coefs[cast::to_size(it.row())].push_back(it.value() *
```
peddie, line 113: If we can output directly to the coefficient vector as suggested above in this function as well, this loop can go away entirely and we can just multiply `combo.coefficients` by `it.value()`.

peddie (approved):
Great idea to make operators for these.  I put some structural comments inline.

I found this tough to read, and I think it's because the `std::vector` and `LinearCombination` indexing is mixed in with the calculation.  I can't help wondering whether a clearer primitive would be some way to view a vector of linear combinations via a value matrix and a coefficient matrix?

include/albatross/src/core/linear_combination.hpp
akleeman: Mind putting a code snippet suggestion up? I'm not sure I follow. Effectively what we're doing here is concatenating all the linear combo values and coeffs together:
```
  output[i].values = concatenate(f[i_0].values, ..., f[i_n].vlaues);
  output[i].coefficients = concatenate(w[i_0] * coefs[i_0], ... , w[i_n] * coefs[i_n]);
  // i_j indicates the j^th non zero index in row i
```
So I can see how we could first loop over and learn the size of the output coefs, allocate that and copy (with a multiplication) into the `Eigen::VectorXd` with blocks ... but I don't see how loops go away unless something really creative happens inside the `concatenate` place holders? Maybe that's what you meant by just multiplying coefs by the value?

include/albatross/src/core/linear_combination.hpp
akleeman: I gave that a shot, it'll avoid copying the `.data()` from a `vector<double>` into the matrix ... but I'm pretty sure it introduces another loop including static casting and I'm unsure whether that means another loop of copying (if so I think the new version will be less efficient right?).

akleeman: Actually, I'm going to scrap my modified version. it doesn't pass tests what I tried was something like:
```
  for (Eigen::Index k = 0; k < matrix.outerSize(); ++k) {
    Eigen::Index elem_count = 0;
    for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
      const auto srow = cast::to_size(it.row());
      const auto scol = cast::to_size(it.col());
      output[srow].values.push_back(features[scol]);
      ++elem_count;
    }

    output[cast::to_size(it.row())].coefficients.resize(elem_count);
    Eigen::Index j = 0;
    for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
      output[it.row()].coefficients[j] = it.value();
      ++j
    }
  }
```
but notice that outer `k` loop isn't a loop over rows necessarily, we need to access `it.row()` and `it.col()` which means I'd have to either 1) make some assumptions about the way the sparse matrix is stored which aren't clearly held here (see ["iterating over non-zero elements"](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html)) or 2) keep a bookkeeping vector around in which I increment the element counts in each row.

akleeman:
> I found this tough to read, and I think it's because the `std::vector` and `LinearCombination` indexing is mixed in with the calculation.

I think part of it is also the awkwardness of iterating over non zero values in an eigen sparse matrix. I maybe made the indexing easier to read?

> I can't help wondering whether a clearer primitive would be some way to view a vector of linear combinations via a value matrix and a coefficient matrix?

Not sure I follow, you mean changing the types stored in a `LinearCombination<X>` to be `Eigen::Vector<X>` types? I agree that'd help with these methods but I think it'd make other parts of the code awkward. So far arbitrary feature types are always in `std::vector` and numerical values in `Eigen::*` (like in `RegressionDataset`). I can see how switching would lead to less awkward indexing since we wouldn't need to cast left and right to appease the compiler, but we'd also lose the ability to iterate over features since Eigen doesn't have that, which might lead to more indexing. (maybe that's not what you meant though?)

peddie:
I had in mind some kind of class like `Eigen::Map` that provides a sparse-matrix view of the coefficients, so the calculations just look like normal matrix calculations without interleaving indexing of all structures in one place.  I agree we shouldn't change `std::vector` in this situation!

include/albatross/src/core/linear_combination.hpp
peddie: This passes all tests for me:

```
@@ -102,26 +102,16 @@ inline auto operator*(const Eigen::SparseMatrixBase<Derived> &matrix,
                       const std::vector<LinearCombination<X>> &features) {
   std::vector<LinearCombination<X>> output(cast::to_size(matrix.rows()));
 
-  std::vector<std::vector<double>> output_coefs(cast::to_size(matrix.rows()));
-
   for (Eigen::Index k = 0; k < matrix.outerSize(); ++k) {
     for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
       const auto row = cast::to_size(it.row());
       const auto col = cast::to_size(it.col());
       const auto &combo = features[col];
-      for (Eigen::Index i = 0; i < combo.coefficients.size(); ++i) {
-        const auto si = cast::to_size(i);
-        output[row].values.push_back(combo.values[si]);
-        output_coefs[row].push_back(it.value() * combo.coefficients[i]);
-      }
+      output[row].values = combo.values;
+      output[row].coefficients = it.value() * combo.coefficients;
     }
   }
 
-  for (std::size_t i = 0; i < output_coefs.size(); ++i) {
-    output[i].coefficients = Eigen::Map<Eigen::VectorXd>(
-        &output_coefs[i][0], static_cast<Eigen::Index>(output_coefs[i].size()));
-  }
-
   return output;
 }
```

include/albatross/src/core/linear_combination.hpp
peddie: This passes all tests for me:

```
@@ -102,26 +102,16 @@ inline auto operator*(const Eigen::SparseMatrixBase<Derived> &matrix,
                       const std::vector<LinearCombination<X>> &features) {
   std::vector<LinearCombination<X>> output(cast::to_size(matrix.rows()));
 
-  std::vector<std::vector<double>> output_coefs(cast::to_size(matrix.rows()));
-
   for (Eigen::Index k = 0; k < matrix.outerSize(); ++k) {
     for (typename Derived::InnerIterator it(matrix.derived(), k); it; ++it) {
       const auto row = cast::to_size(it.row());
       const auto col = cast::to_size(it.col());
       const auto &combo = features[col];
-      for (Eigen::Index i = 0; i < combo.coefficients.size(); ++i) {
-        const auto si = cast::to_size(i);
-        output[row].values.push_back(combo.values[si]);
-        output_coefs[row].push_back(it.value() * combo.coefficients[i]);
-      }
+      output[row].values = combo.values;
+      output[row].coefficients = it.value() * combo.coefficients;
     }
   }
 
-  for (std::size_t i = 0; i < output_coefs.size(); ++i) {
-    output[i].coefficients = Eigen::Map<Eigen::VectorXd>(
-        &output_coefs[i][0], static_cast<Eigen::Index>(output_coefs[i].size()));
-  }
-
   return output;
 }
```

akleeman: I don't think that's doing the same thing, if I'm correct that ^ will fill `output[row]` with only the product from the last entry in that `row`. I guess this means there's a gap in the tests (they mostly just check the output types, probably should add some validation on the values).

akleeman: I've updated the tests to actually check the content of the transformations after which that ^ no longer passes, feel free to experiment if you think there's a better way (but if you find yourself looking at this chart https://xkcd.com/1691/ ...).

peddie: It took me a while of staring at it, but I get it now.  Sorry for the confusion.  Glad the tests are better at least!

PR #361, opened 2022-10-14
Dense/Sparse Matrix by Distribution Multiplication

akleeman:
This adds the ability to lazily multiply a matrix by a distribution, then decide whether you'd like the result to be a joint distribution or a marginal distribution (in which case the laziness lets you save computations).  They syntax looks like this,

```
MarginalDistribution transformed = (matrix * distribution).marginal();
JointDistribution transformed = (matrix * distribution).joint();
```

It also enables multiplication with sparse matrices. In a future PR I'd like to use this to be able to do sparse transformations of datasets:

```
RegressionDataset<LinearCombination<X>> transformed = T * x_dataset;
```

which would make it a lot easier to build a single transformation and apply it not only to a dataset (which internally has a MarginalDistribution) but also to any corresponding predictions, for example:

```
compute_likelihood(diff_operator * dataset.targets,
                   diff_operator * model.predict(dataset.features).joint())
```

### DIFF
--- a/tests/test_core_distribution.cc
+++ b/tests/test_core_distribution.cc
@@ -31,33 +31,94 @@ TYPED_TEST_P(DistributionTest, test_subset) {
   expect_subset_equal(dist.covariance, ss.covariance, indices);
 };
 
-TYPED_TEST_P(DistributionTest, test_multiply_with_matrix) {
+TYPED_TEST_P(DistributionTest, test_multiply_with_matrix_marginal) {
 
   TypeParam test_case;
   const auto dist = test_case.create();
 
   Eigen::MatrixXd mat = Eigen::MatrixXd::Random(dist.size() - 1, dist.size());
 
-  const auto transformed_distribution = dist * mat;
+  const auto transformed_distribution = (mat * dist).marginal();
+  const MarginalDistribution alternate = mat * dist;
+  EXPECT_EQ(alternate, transformed_distribution);
 
   EXPECT_EQ(transformed_distribution.mean, mat * dist.mean);
-  EXPECT_EQ(transformed_distribution.covariance,
-            mat * dist.covariance * mat.transpose());
+  const Eigen::MatrixXd cov = mat * dist.covariance * mat.transpose();
+  EXPECT_LT(
+      (transformed_distribution.covariance.diagonal() - cov.diagonal()).norm(),
+      1e-8);
+};
+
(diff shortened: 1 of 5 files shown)


tests/test_core_distribution.cc
```diff
@@ -31,33 +31,94 @@ TYPED_TEST_P(DistributionTest, test_subset) {
...
+      1e-8);
 };
 
-TYPED_TEST_P(DistributionTest, test_multiply_with_vector) {
+TYPED_TEST_P(DistributionTest, test_multiply_with_matrix_joint) {
 
   TypeParam test_case;
   const auto dist = test_case.create();
 
+  Eigen::MatrixXd mat = Eigen::MatrixXd::Random(dist.size() - 1, dist.size());
```
peddie, line 57: Does this line do anything?

peddie (approved):
Any reason this PR comment is not part of the commit message?

Seems like a reasonable thing to add.  Did you hit a performance problem and make this in response, or is this done just based on a priori knowledge of the sizes and asymptotics involved?

tests/test_core_distribution.cc
akleeman: good catch.

akleeman:
> Any reason this PR comment is not part of the commit message?

I think that would only be visible during review right? After squash merge we'll just get the PR title and a link to the PR (https://github.com/swift-nav/albatross/commits/master).

peddie:
The commit message goes into the git history forever, so this contextual information is available right along with the change whenever you look for it in the future -- no references necessary.

akleeman:
Here's an example commit which makes me think squash merge just replaces the entire commit message: https://github.com/swift-nav/albatross/commit/38b08f28c083c9362ad32941523113aa5efafd89

peddie:
I think I just understood the concern about how it's merged.  In git documentation, "squash" means to combine two commits and also concatenate their commit messages.  But maybe github uses "squash" to mean "remove the messages?"

In any case, I'm just trying to advocate for putting this valuable contextual information right in the git commit history rather than just have a reference to a github page.  If I'm searching through the git history for something, it's a lot quicker to just read what happened than to follow a link for every commit I may be interested in, not to mention the text can be searched.

akleeman:
I can see how searching through git log full descriptions would be nice.  There's another camp which prefers short descriptions so it's easy to quickly glance at `git log` and see overviews of recent changes, then PRs for context.  I don't have a super strong opinion, but for now this is the swift policy :shrug:

PR #362, opened 2022-10-18
Add methods to fill a covariance matrix by group with async

akleeman:
This adds a method `compute_covariance_by_group` which lets you fill a [cross] covariance by group in parallel, meaning you provide a group which divides data into chunks, then these chunks get filled asynchronously.

Probably there could be some optimization to be done around making sure your filling contiguous chunks, but the more general approach here (groups don't need to be ordered) was done in anticipation of making the groups meaningful in the sense that some pairs of groups may be known to be zero.  So in a future change we could provide a `Grouper` which also has a `is_zero(group_a, group_b)` method which lets us avoid computations of chunks all together.

### DIFF
--- a/include/albatross/src/indexing/covariance.hpp
+++ b/include/albatross/src/indexing/covariance.hpp
@@ -1,0 +1,138 @@
+/*
+ * Copyright (C) 2022 Swift Navigation Inc.
+ * Contact: Swift Navigation <dev@swiftnav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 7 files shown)


include/albatross/src/indexing/covariance.hpp
```diff
@@ -0,0 +1,138 @@
...
+          }
+        }
+      } else if (x_key == y_pair.first) {
+        // Diagonal case, fill both triangles directly
+        details::fill_covariance_subset(caller, subset(xs, x_inds), x_inds, &C);
+      }
+    }
+  };
+
+  async_apply(x_indexers, fill_row);
```
peddie, line 131: It would be nice to decouple the calculation strategy (async vs. not) from the accelerations possible by grouping before evaluation.  We may still want to use a thread pool to evaluate covariances that don't benefit from grouping, and conversely we may want to group wisely but not evaluate blocks concurrently.

akleeman:
closed in favor of: https://github.com/swift-nav/albatross/pull/377

PR #363, opened 2022-10-20
Move stream overloads into albatross namespace

akleeman:
This moves any of the `operator <<` overloads in albatross into the same namespace as the class it operates on.

This is done to avoid method hiding in which having any `operator <<` defined in a non-albatross namespace means any globally defined `operator <<` will get hidden.  For example:

```
#include <iostream>

namespace foo {
  struct Foo {};
}

void call(const foo::Foo &) {
  std::cout << "foo::call" << std::endl;
}

namespace bar {

struct Bar {};

//void call(const Bar &) {}

void call_foo() {
  std::cout << "bar::";
  foo::Foo x;
  call(x);
}

}

void call_foo() {
  foo::Foo x;
  call(x);
}

int main() {

  call_foo();
  bar::call_foo();
  return 0.;
}
```
will not compile if you uncomment the `void call(const Bar &) {}` line because including that line will lead to the compiler hiding any `call()` method in any parent namespaces (including the global namespace).

### DIFF
--- a/include/albatross/src/models/ransac.hpp
+++ b/include/albatross/src/models/ransac.hpp
@@ -529,13 +529,10 @@ ModelBase<ModelType>::ransac(const StrategyType &strategy,
   return Ransac<ModelType, StrategyType>(derived(), strategy, config);
 }
 
-} // namespace albatross
-
-template <typename X, typename std::enable_if_t<
-                          albatross::is_streamable<X>::value, int> = 0>
-inline std::ostream &
-operator<<(std::ostream &os, const albatross::RansacIteration<X> &iteration) {
-
+template <typename X,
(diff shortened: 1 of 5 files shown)


include/albatross/src/models/ransac.hpp
```diff
@@ -529,8 +529,6 @@ ModelBase<ModelType>::ransac(const StrategyType &strategy,
   return Ransac<ModelType, StrategyType>(derived(), strategy, config);
 }
 
-} // namespace albatross
-
 template <typename X, typename std::enable_if_t<
                           albatross::is_streamable<X>::value, int> = 0>
```
jbangelo, line 533:
```suggestion
                          is_streamable<X>::value, int> = 0>
```

PR #364, opened 2022-10-26
Build albatross with bazel [BUILD-364]

krisukox:
This PR adds Bazel to albastross.

This PR needs this PR to be merged:
https://github.com/swift-nav/swiftnav-bazel/pull/9


.github/workflows/ci.yaml
```diff
@@ -87,6 +87,30 @@ jobs:
...
+  bazel:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_KEY }}
+
+      - name: Install clang-14
```
isaactorz, line 99: I think we can just use the system CC for now.

I'm going to punt on requiring to use clang as our compiler until I see if my bazel PR can get merged.

BUILD.bazel
```diff
@@ -0,0 +1,212 @@
...
+        "tests/test_traits_core.cc",
+        "tests/test_traits_covariance_functions.cc",
+        "tests/test_traits_details.cc",
+        "tests/test_traits_evaluation.cc",
+        "tests/test_traits_indexing.cc",
+        "tests/test_tune.cc",
+        "tests/test_utils.h",
+        "tests/test_variant_utils.cc",
+    ],
+    copts = ["-std=c++14"],
```
isaactorz, line 204: For the time being I think the way we should set the C++ standard is to enable it via `BAZEL_CXXOPTS` in a `.bazelrc` file.

BUILD.bazel
```diff
@@ -0,0 +1,212 @@
...
+        "include/albatross/utils/AsyncUtils",
+        "include/albatross/utils/CsvUtils",
+        "include/albatross/utils/LinalgUtils",
+        "include/albatross/utils/MapUtils",
+        "include/albatross/utils/RandomUtils",
+        "include/albatross/utils/VariantUtils",
+    ],
+    includes = [
+        "include",
+        "tests",
```
isaactorz, line 134: Does albatross include files that are part of its test directory?
Is it possible to set these via `copts` instead (i.e. using -I).

BUILD.bazel
```diff
@@ -0,0 +1,212 @@
+load("//bazel:swift_cc_defs.bzl", "UNIT", "swift_cc_library", "swift_cc_test")
+
+swift_cc_library(
```
isaactorz, line 3: Make sure that we also set the warning flags for this project: https://github.com/swift-nav/albatross/blob/01e266027fa4096165966f1b6b847b4f11944a2d/CMakeLists.txt#L45

BUILD.bazel
```diff
@@ -0,0 +1,212 @@
...
+        "@eigen",
+        "@fast_csv",
+        "@gzip",
+        "@nlopt",
+        "@variant",
+    ],
+)
+
+swift_cc_test(
+    name = "albatorss-test",
```
isaactorz, line 148: Let's add the compile options for this target as well: https://github.com/swift-nav/albatross/blob/01e266027fa4096165966f1b6b847b4f11944a2d/tests/CMakeLists.txt#L54

BUILD.bazel
krisukox: Since this target contains only headers, I added warning flags to the `albatorss-test` target and examples binary.

.github/workflows/ci.yaml
```diff
@@ -87,6 +87,23 @@ jobs:
...
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_KEY }}
+
+      - name: Run Bazel
+        env:
+          CC: clang-14
```
isaactorz, line 101: Let's build with the same compiler that the cmake build is using (assuming thats not clang-14).

BUILD.bazel
```diff
@@ -0,0 +1,210 @@
...
+        "tests/test_traits_core.cc",
+        "tests/test_traits_covariance_functions.cc",
+        "tests/test_traits_details.cc",
+        "tests/test_traits_evaluation.cc",
+        "tests/test_traits_indexing.cc",
+        "tests/test_tune.cc",
+        "tests/test_utils.h",
+        "tests/test_variant_utils.cc",
+    ],
+    copts = COPTS,
```
isaactorz, line 202:
```suggestion
    copts = COPTS,
    local_defines ["CSV_IO_NO_THREAD"],
```

This is in the cmake build: https://github.com/swift-nav/albatross/blob/01e266027fa4096165966f1b6b847b4f11944a2d/tests/CMakeLists.txt#L55.

isaactorz (commented):
Just a couple more things but otherwise looks good

.github/workflows/ci.yaml
krisukox: I changed it to gcc-6.

BUILD.bazel
krisukox: Done

BUILD.bazel
```diff
@@ -0,0 +1,211 @@
+load("//bazel:swift_cc_defs.bzl", "UNIT", "swift_cc_library", "swift_cc_test")
+load("//:copts.bzl", "COPTS")
+
+swift_cc_library(
```
isaactorz, line 4: Sorry, one last thing.

Only use the `swift_cc` wrappers for code that's using our custom swift cmake wrappers.

Albatross is just using regular cmake `add_library`.

It probably doesn't make much of a difference here, but we'll likely be unable to compile code that's not already using those wrappers in other projects (it doesn't respect all the warnings/settings).

krisukox: Done

isaactorz:
cc @akleeman this PR adds bazel support for albatross.

We need this to support building the projects that depend on it with bazel. We've set it up so that the CI pipeline won't fail if the bazel build doesn't pass.

Regardless of how this bazel project develops my thinking is that for open source projects, cmake will continue to be the primary build system.

Adding you for a +1 for adding support for another build system.

isaactorz (approved):
LGTM I've added @akleeman for review too.

BUILD.bazel
```diff
@@ -0,0 +1,211 @@
...
+        "include/albatross/serialize/Evaluation",
+        "include/albatross/serialize/GP",
+        "include/albatross/serialize/LeastSquares",
+        "include/albatross/serialize/Ransac",
+        "include/albatross/src/cereal/block_utils.hpp",
+        "include/albatross/src/cereal/dataset.hpp",
+        "include/albatross/src/cereal/distribution.hpp",
+        "include/albatross/src/cereal/eigen.hpp",
+        "include/albatross/src/cereal/evaluation.hpp",
+        "include/albatross/src/cereal/fit_model.hpp",
```
akleeman, line 36: interesting, do you have to manually add each new header for bazel?

akleeman (approved):
lgtm, I know nothing about bazel setup ... but see no issue with adding it.

BUILD.bazel
isaactorz: Yes. The way bazel works is it copies every file into a "sandbox" environment where it performs the build. The idea is to isolate the build from the host environment.

So they require that you explicitly declare all the inputs rather than just giving it include paths like cmake.

PR #365, opened 2022-11-04
Traits member value should be `constexpr`

peddie:


include/albatross/src/details/traits.hpp
```diff
@@ -84,9 +84,12 @@ template <typename T> class is_streamable {
   template <typename> static std::false_type test(...);
 
 public:
-  static const bool value = decltype(test<T>(0))::value;
+  static constexpr bool value = decltype(test<T>(0))::value;
```
akleeman, line 87: :+1:

include/albatross/src/details/traits.hpp
```diff
@@ -84,9 +84,12 @@ template <typename T> class is_streamable {
...
 };
 
+template <typename T>
+constexpr bool is_streamable<T>::value;
```
akleeman, line 91: this seems strange to me, why is this needed for only this trait?

peddie: I don't have a good answer for this -- the linking issue that led to this patch occurred when using runtime sanitizers, so it's possible the program I was linking (a test suite) simply didn't use the other traits?  Perhaps they are usually all inlined long before the linking stage, and turning on the sanitizer somehow prevents this one value from being inlined away in the template system?

Please double-check the following reasoning, but if I understand correctly, it's normal to have to separately define *static* member variables like this outside a class.  For an untemplated class, you would be expected to do it in a source file so it's linked only once.  For a templated class, you can define it in a header here because the compiler understands how templates get expanded.  Finally, you may sometimes get away without separately defining them if they are all inlined before linking and ODR etc. come into play, but that's no guarantee.

akleeman: Had a bit of a hard time finding similar traits defined with similar style in other packages, but I did find this:

https://github.com/USCiLab/cereal/blob/master/include/cereal/details/traits.hpp#L460

which is actually closer to `master` (`const bool` not `constexpr bool` and no (re)definition outside of the struct).

When it comes to method (not member variable) definition vs implementation I see (looking at Eigen for example) a mix of approaches with things like the `PlainObjectBase` class ([here](https://github.com/swift-nav/eigen/blob/master/Eigen/src/Core/PlainObjectBase.h#L154)) which is entirely implementation when defined.  While other things like `MatrixBase` [have implementations outside of defintion](https://github.com/swift-nav/eigen/blob/master/Eigen/src/Core/MatrixBase.h#L485).


Ie, I see things like this:
```
template <typename T>
struct Foo {
  bool get_value();
}

template <typename T>
bool Foo<T>::get_value() { return m_value;};

```
but haven't found any examples of:
```
template <typename T>
struct Foo {
  constexpr bool value;
}

template <typename T>
constexpr bool Foo<T>::value;
```

Anyway, I'm happy to put this in the massive category of "things I don't understand about compilers", mostly just curious why this specific trait gets treated different from the rest. but I don't see anything that feels dangerous here.

peddie: I don't have a better explanation than "probably inlining," but I can add the same out-of-line definition for all the other traits in this file if you prefer.  If the understanding I detailed above is correct, then technically I should be doing that and not relying on inlining / template magic.

peddie: Actually this is still a bit of a mystery to me, because (again, if I understand correctly) `constexpr` should imply `inline`, which should obviate the need for out-of-line definition.  So there's something else going on with this linker failure -- perhaps a clang bug of some kind.

akleeman: I think I'd vote just add this line to fix your issue (rather than adding it everywhere).  Definitely still don't understand the problem entirely.

akleeman: I wonder if it has to do with the use of `auto` in the return type for the test case for this trait?
```
  template <typename C>
  static auto test(int)
      -> decltype(std::declval<std::ostream &>() << std::declval<C>(),
                  std::true_type());
```
other traits are usually more explicit about the return type: https://github.com/swift-nav/albatross/blob/master/include/albatross/src/core/traits.hpp#L25

peddie: I was able to link successfully using `clang++-14` by removing the out-of-line definition and instead declaring it `static inline constexpr`.  If I understand correctly, this is redundant; `constexpr` should imply `idris`.  Perhaps a compiler bug, or perhaps an exception to a special case in the language.

Unfortunately according to the CI system, this is an error.  The OS X build complains that declaring variables `inline` is a C++17 extension.

I'm going to revert to the previous version and merge that if you approve it, I guess.

akleeman (approved):

PR #370, opened 2022-12-07
Use a thread pool instead of `std::async`

peddie:
- Pull in a minimal third-party thread pool definition

 - Replace `apply_async()` and `solve_*async()` with overloads
   accepting a pointer to a thread pool

 - Add overloads for `compute_covariance_matrix()` to allow parallel
   threaded calculation of covariance matrices

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -139,11 +139,12 @@ cc_library(
         "@gzip",
         "@nlopt",
         "@variant",
+        "@ThreadPool",
     ],
 )
 
 swift_cc_test(
-    name = "albatorss-test",
+    name = "albatross-test",
     srcs = [
         "tests/mock_model.h",
         "tests/test_apply.cc",
@@ -189,6 +190,7 @@ swift_cc_test(
         "tests/test_serialize.h",
         "tests/test_sparse_gp.cc",
         "tests/test_stats.cc",
+        "tests/test_thread_pool.cc",
         "tests/test_traits_cereal.cc",
         "tests/test_traits_core.cc",
         "tests/test_traits_covariance_functions.cc",
@@ -207,5 +209,6 @@ swift_cc_test(
     deps = [
         ":albatross",
         "@gtest//:gtest_main",
+        "@zlib",
     ],
 )
--- a/include/albatross/src/core/model.hpp
+++ b/include/albatross/src/core/model.hpp
@@ -28,7 +28,13 @@ template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
   template <typename T, typename FeatureType> friend class fit_model_type;
 
 protected:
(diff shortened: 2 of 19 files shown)


include/albatross/src/core/model.hpp
```diff
@@ -28,7 +28,13 @@ template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
   template <typename T, typename FeatureType> friend class fit_model_type;
 
 protected:
-  ModelBase() : insights(), use_async_(DEFAULT_USE_ASYNC){};
+  ModelBase()
+      : insights(),
+        threads_(DEFAULT_USE_ASYNC
+                     ? std::make_shared<ThreadPool>(std::max(
```
akleeman, line 34: should this use `get_default_thread_count`?

BUILD.bazel
```diff
@@ -139,11 +140,12 @@ cc_library(
         "@gzip",
         "@nlopt",
         "@variant",
+        "@ThreadPool",
     ],
 )
 
 swift_cc_test(
-    name = "albatorss-test",
+    name = "albatross-test",
```
akleeman, line 148: ha, nice catch

akleeman (commented):
lgtm, if it's easy enough maybe worth splitting this into two changes? The async -> thread pool.  and then a follow up which adds block covariance computations?

include/albatross/src/core/model.hpp
peddie: I could do that, but it would change the `#include` structure and include the async utils into `core/`.  If that's OK, I can do it; otherwise I thought I'd write it out to avoid changing the graph of inclusions.

peddie:
Sure, I'll split into two different PRs, but . . . chaining PRs is :(

peddie:
Covariance function change is at #377.

include/albatross/src/core/model.hpp
akleeman: I see. I guess the alternative would be to add a separate header which basically just holds this and include it in both core and async tooling?

Anyway, the include dependencies are far from perfect, no need to change this if you don't think it's worth it.

BUILD.bazel
```diff
@@ -86,6 +86,7 @@ cc_library(
         "include/albatross/src/evaluation/traits.hpp",
         "include/albatross/src/graph/minimum_spanning_tree.hpp",
         "include/albatross/src/indexing/apply.hpp",
+        "include/albatross/src/indexing/block.hpp",
```
akleeman, line 89: leftover from splitting things apart?

akleeman (approved):
lgtm, can't wait to use it!

BUILD.bazel
peddie: D'oh.  Yeah, I'll remove.

Issue #372, opened 2022-12-09
Possible heap memory error in group-by code


peddie:
In [this CI build using the thread sanitizer](https://github.com/swift-nav/albatross/actions/runs/3654531262/jobs/6175036522), we get the following error:
```
[----------] 10 tests from test_groupby/GroupByTester/0, where TypeParam = albatross::BoolClassMethodGrouper
[ RUN      ] test_groupby/GroupByTester/0.test_groupby_access_methods
==================
WARNING: ThreadSanitizer: heap-use-after-free (pid=6243)
  Read of size 8 at 0x7b0400005210 by main thread:
    #0 std::_Bit_reference::operator bool() const /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_bvector.h:96:17 (albatross_unit_tests+0x68360e) (BuildId: 8d743b0ee98587d31ebbfc53e9fff979f9e97fa7)
    #1 albatross::gtest_suite_GroupByTester_::test_groupby_access_methods<albatross::BoolClassMethodGrouper>::TestBody() /home/runner/work/albatross/albatross/tests/test_group_by.cc:231:19 (albatross_unit_tests+0x68360e)
    #2 void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) <null> (albatross_unit_tests+0xa45d95) (BuildId: 8d743b0ee98587d31ebbfc53e9fff979f9e97fa7)

  Previous write of size 8 at 0x7b0400005210 by main thread:
    #0 operator delete(void*) <null> (albatross_unit_tests+0x3bd36e) (BuildId: 8d743b0ee98587d31ebbfc53e9fff979f9e97fa7)
    #1 std::__new_allocator<unsigned long>::deallocate(unsigned long*, unsigned long) /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/new_allocator.h:158:2 (albatross_unit_tests+0x683606) (BuildId: 8d743b0ee98587d31ebbfc53e9fff979f9e97fa7)
    #2 std::allocator_traits<std::allocator<unsigned long> >::deallocate(std::allocator<unsigned long>&, unsigned long*, unsigned long) /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/alloc_traits.h:496:13 (albatross_unit_tests+0x683606)
[...]

PR #373, opened 2022-12-12
Add address/undefined behavior/thread sanitizer builds to CI

peddie:
This PR fixes existing runtime sanitizer failures and then adds CI jobs to run the sanitizers on every PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,6 +63,39 @@ jobs:
           bash ./ci/run_tests.sh
           bash ./ci/ensure_copyright.sh
 
+  ubuntu-clang14-sanitizers:
+    runs-on: ubuntu-22.04
+    strategy:
+      os:
+        - { name: "address-undefined-leak", script: "./ci/run_address_sanitizer_tests.sh" }
+        - { name: "thread", script: "./ci/run_thread_sanitizer_tests.sh" }
+        # Memory sanitizer is not run because it requires building a
+        # custom instrumented libc.
+
+    steps:
+
+      - name: Checkout source
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_KEY }}
+
+      - name: Install clang-14
+        run: |
+          sudo apt-get update
+          sudo apt-get install build-essential software-properties-common -y
+          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+          sudo apt-get update
+          sudo apt-get install clang-14
+          clang -v
+      - name: Run (${{ matrix.name }}) sanitizer tests
+        env:
+          CC: clang-14
+          CXX: clang++-14
+          TESTENV: clang14
+        run: |
+          bash "${{ matrix.script }}"
+
   format:
     runs-on: ubuntu-22.04
     steps:
--- a/tests/test_group_by.cc
+++ b/tests/test_group_by.cc
@@ -227,7 +227,8 @@ TYPED_TEST_P(GroupByTester, test_groupby_access_methods) {
 
   const auto const_groups = grouper.groups();
   assert(const_groups.size() > 0);
-  const auto first_key = const_groups.keys()[0];
+  const typename decltype(const_groups)::key_t first_key =
+      const_groups.keys()[0];
   const_groups.at(first_key);
 
   auto groups = grouper.groups();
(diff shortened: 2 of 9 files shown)


peddie:
This cannot be merged for two reasons:
 - Part of #374 fixes an error that prevents `clang-14` from building
 - The address and thread sanitizers both fail due to #372 
   - I cannot mask the failure at the function level because function attributes do not seem to work when applied to templated functions, and the test case that fails is a `TYPED_TEST_P`, so it's impractical to even pull out the failing bit into a separate function that I can annotate instead.

peddie:
I've addressed the issues I mentioned above; it's ready for review.

tests/test_group_by.cc
```diff
@@ -262,7 +263,7 @@ TYPED_TEST_P(GroupByTester, test_groupby_modify_combine) {
   const auto grouped = group_by(parent, this->test_case.get_grouper());
   auto groups = grouped.groups();
 
-  const auto first_key = map_keys(groups)[0];
+  const typename decltype(groups)::key_t first_key = map_keys(groups)[0];
```
akleeman, line 266: curious why `auto` doesn't work here?

akleeman (approved):
lgtm!

.github/workflows/ci.yaml
```diff
@@ -63,6 +63,61 @@ jobs:
...
+      - name: Checkout source
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_KEY }}
+
+      - name: Install clang-14
+        run: |
+          sudo apt-get update && \
+          sudo apt-get install build-essential software-properties-common -y && \
```
john-michaelburke, line 79: nit. In github actions you don't have to add a `&& \` for each line when using the default shell. It will evaluate the entire run step as a shared shell.

.github/workflows/ci.yaml
```diff
@@ -63,6 +63,61 @@ jobs:
...
+          sudo apt-get update && \
+          sudo apt-get install clang-14 && \
+          gcc -v
+      - name: Run build
+        env:
+          CC: clang-14
+          CXX: clang++-14
+          TESTENV: clang14
+        run: |
+          bash ./ci/run_thread_sanitizer_tests.sh
```
john-michaelburke, line 119: extreme nit. This and the `address undefined leak` jobs are relatively similar so we can matricize them. Something like:
```suggestion
  ubuntu-clang14-sanitizer-tests:
    runs-on: ubuntu-22.04
    strategy:
        os:
          - { name: "address-undefined-leak", script: "./ci/run_address_sanitizer_tests.sh" }
          - { name: "thread", script: "./ci/run_thread_sanitizer_tests.sh" }
    steps:

      - name: Checkout source
        uses: actions/checkout@v2
        with:
          submodules: recursive
          ssh-key: ${{ secrets.SSH_KEY }}

      - name: Install clang-14
        run: |
          sudo apt-get update
          sudo apt-get install build-essential software-properties-common -y
          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
          sudo apt-get update
          sudo apt-get install clang-14
          gcc -v
      - name: Run (${{ matrix.name }} ) Sanitizer Tests
        env:
          CC: clang-14
          CXX: clang++-14
          TESTENV: clang14
        run: |
          bash "${{ matrix.script }}"
```

.github/workflows/ci.yaml
```diff
@@ -63,6 +63,61 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+
+      - name: Checkout source
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_KEY }}
+
+      - name: Install clang-14
```
john-michaelburke, line 105: We might be able to replace this with a premade action, such as:
https://github.com/marketplace/actions/install-llvm-and-clang

Maybe in a separate PR.

john-michaelburke (approved):

tests/test_group_by.cc
peddie: It deduces a reference type which has lifetime issues, triggering the address sanitizer!

.github/workflows/ci.yaml
peddie: Done; let's see if I copied correctly :)

.github/workflows/ci.yaml
peddie: Feel free to make a ticket for the future -- for now I think I'll just keep it simple.

.github/workflows/ci.yaml
peddie: Excellent; thanks.

PR #374, opened 2022-12-12
Use `swift_add_tool()` for unit test suite

peddie:
This enables additional warnings.

### DIFF
--- a/include/albatross/src/core/model.hpp
+++ b/include/albatross/src/core/model.hpp
@@ -48,8 +48,8 @@ protected:
       typename std::enable_if<has_possible_fit<ModelType, FeatureType>::value &&
                                   !has_valid_fit<ModelType, FeatureType>::value,
                               int>::type = 0>
-  void _fit(const std::vector<FeatureType> &features,
-            const MarginalDistribution &targets) const
+  void _fit(const std::vector<FeatureType> &features ALBATROSS_UNUSED,
+            const MarginalDistribution &targets ALBATROSS_UNUSED) const
       ALBATROSS_FAIL(FeatureType,
                      "The ModelType *almost* has a _fit_impl method for "
                      "FeatureType, but it appears to be invalid");
@@ -59,8 +59,8 @@ protected:
                 !has_possible_fit<ModelType, FeatureType>::value &&
                     !has_valid_fit<ModelType, FeatureType>::value,
                 int>::type = 0>
-  void _fit(const std::vector<FeatureType> &features,
-            const MarginalDistribution &targets) const
+  void _fit(const std::vector<FeatureType> &features ALBATROSS_UNUSED,
+            const MarginalDistribution &targets ALBATROSS_UNUSED) const
       ALBATROSS_FAIL(
           FeatureType,
           "The ModelType is missing a _fit_impl method for FeatureType.");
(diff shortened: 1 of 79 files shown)


include/albatross/src/core/model.hpp
```diff
@@ -59,8 +59,8 @@ template <typename ModelType> class ModelBase : public ParameterHandlingMixin {
                 !has_possible_fit<ModelType, FeatureType>::value &&
                     !has_valid_fit<ModelType, FeatureType>::value,
                 int>::type = 0>
-  void _fit(const std::vector<FeatureType> &features,
```
akleeman, line 62: the other option is (I think) leaving out the variable name,
```
void _fit(const std::vector<FeatureType> &,
```

include/albatross/src/details/typecast.hpp
```diff
@@ -0,0 +1,53 @@
...
+  ALBATROSS_ASSERT(input >= 0);
+  return static_cast<std::size_t>(input);
+}
+
+constexpr Eigen::Index to_index(std::size_t input) {
+  ALBATROSS_ASSERT(input <= to_size(std::numeric_limits<Eigen::Index>::max()));
+  return static_cast<Eigen::Index>(input);
+}
+
+constexpr std::size_t kMaxIntegerPrecisionDoubleSize = 1ULL << 53;
```
akleeman, line 30: magic number: maybe a comment about why 53?

include/albatross/src/details/typecast.hpp
```diff
@@ -0,0 +1,53 @@
...
+constexpr double to_double(std::size_t input) {
+  ALBATROSS_ASSERT(input <= kMaxIntegerPrecisionDoubleSize);
+  return static_cast<double>(input);
+}
+
+// We have to be sure to handle the case of 32-bit `Eigen::Index`
+constexpr Eigen::Index kMaxIntegerPrecisionDoubleIndex =
+    static_cast<Eigen::Index>(std::min(
+        static_cast<long long>(std::numeric_limits<Eigen::Index>::max()),
+        1LL << 53));
```
akleeman, line 41: same

tests/mock_model.h
```diff
@@ -102,12 +101,12 @@ class MockModel : public ModelBase<MockModel> {
...
 
 static inline RegressionDataset<MockFeature>
-mock_training_data(const int n = 10) {
+mock_training_data(const Eigen::Index n = 10) {
   std::vector<MockFeature> features;
   Eigen::VectorXd targets(n);
   for (int i = 0; i < n; i++) {
     features.push_back(MockFeature(i));
-    targets[i] = static_cast<double>(i + n);
+    targets[static_cast<Eigen::Index>(i)] = cast::to_double(i + n);
```
akleeman, line 109: `to_index`?

akleeman (approved):

include/albatross/src/core/model.hpp
peddie: Yes -- `clang-tidy` complains about this, however.

include/albatross/src/details/typecast.hpp
peddie: Good call, will do

tests/mock_model.h
peddie: Not in this case -- it's not being cast from an STL index (`std::size_t`) but from a mock feature of type `int`.  I can add some more checks for other cases like this, but I tried to focus here mainly on safely handling index conversions.

include/albatross/src/core/model.hpp
akleeman: Ah weird, anyway this works!

tests/mock_model.h
akleeman: Ah, I see.  Yeah all good, just thought it was an overlooked case.

PR #377, opened 2022-12-12
Compute covariance functions in parallel in sparse GP

peddie:
This modifies the covariance operators to accept an optional `ThreadPool` pointer to be used for parallel calculation.  The sparse GP model now uses this facility by default in fitting (but not in prediction).


akleeman (approved):
Nice one! Mind adding a description to the PR before merging to give context if anyone in the future is looking back at this, otherwise good to go.

PR #378, opened 2022-12-12
Add clang-tidy check in CI

peddie:


peddie:
I've added a ratcheting script to let us enable `clang-tidy` checking without fixing the existing warnings.

woodfell (approved):
I like it, seems like a good way of incrementally decreasing the number of defects

jbangelo (approved):

PR #381, opened 2022-12-13
Serialization for ThreadPool objects

peddie:
This commit adds serialization support for `ThreadPool`s.

### DIFF
--- a/tests/test_serialize.cc
+++ b/tests/test_serialize.cc
@@ -449,4 +449,45 @@ TEST(test_serialize, test_joint_streamable) {
   oss << dist;
 }
 
+TEST(test_serialize, test_ThreadPool_threads) {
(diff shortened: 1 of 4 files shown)


tests/test_serialize.cc
```diff
@@ -449,4 +449,22 @@ TEST(test_serialize, test_joint_streamable) {
   oss << dist;
 }
 
+TEST(test_serialize, test_ThreadPool) {
```
akleeman, line 452: nit: might be worth a `nullptr` test as well

akleeman (approved):
lgtm!

tests/test_serialize.cc
peddie: Will do!

peddie: I expanded test coverage substantially, have a look.

PR #384, opened 2022-12-14
Fix typo in github actions workflow.

john-michaelburke:
Fixed a minor typo that was preventing the workflow from running.
Removed extra `&& \` instances.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,7 +1,7 @@
 name: CI
 
 on:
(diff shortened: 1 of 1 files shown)


john-michaelburke:
> Looks good -- is there a way to ensure that if there is another error in the CI config, we will get a failure?

Not that I'm aware of. It is kind of a catch 22 because it fails to parse the workflow yaml. I don't know why github doesn't mark this as a failure rather than just omitting it entirely.

PR #391, opened 2022-12-29
Remove Patchwork GP

akleeman:
This model hasn't proven quite as useful as expected, we can always dig it up in git history if we want it back.


akleeman:
@peddie can you explain how I should inspect the clang tidy output? I rebuilt and ran `ci/run_tidy.sh` which produced similar warnings ... but I don't see any output which points me to what got added, only that there are more warnings? This seems particularly strange since this is all deletions :shrug:

peddie (approved):
I'll allow incrementing one count because I also have no idea how a deletion could cause that . . .

PR #394, opened 2023-01-09
Add missing test indclude path [BUILD-384]

wwieclaw:
Those files are used in the orion-engine tests


BUILD.bazel
```diff
@@ -130,17 +130,26 @@ cc_library(
         "include/albatross/utils/MapUtils",
         "include/albatross/utils/RandomUtils",
         "include/albatross/utils/VariantUtils",
+        "tests/mock_model.h",
+        "tests/test_core_distribution.h",
+        "tests/test_covariance_utils.h",
+        "tests/test_models.h",
+        "tests/test_serialize.h",
+        "tests/test_utils.h",
```
krisukox, line 138: I think we should extract these files to a new `cc_test_library` and use this library in `albatross-test` and the orion-engine tests.

isaactorz (commented):
Agree with @krisukox suggestion!

isaactorz (approved):
lgtm assuming CI passes

PR #395, opened 2023-01-10
Test clang block

akleeman:


akleeman:
@peddie not sure if this is what you were expecting, but it looks like it'll give a success checkmark, but if you dig in tidy actually failed :shrug:

peddie:
That's a bit frustrating (I was hoping for like a yellow X or something), but let's merge it so as not to block anyone else.

peddie:
Feel free to close; the parent is merged.

PR #397, opened 2023-01-13
Fix bazel build

isaactorz:
Fixes up the bazel build to account for changes introduced in https://github.com/swift-nav/albatross/pull/390


isaactorz:
ubuntu repos are being a real pain.

peddie (approved):

PR #401, opened 2023-01-20
ci: bazel: require bazel builds pass to merge

isaactorz:
Removes the condition that passes the CI pipeline regardless of whether
the bazel build passes or not.

Also updates the bazel build to use the same ubuntu/gcc as the other stages.


akleeman (approved):
lgtm!

PR #404, opened 2023-02-08
Add rational quadratic covariance

peddie:


peddie:
I haven't got gpytorch oracle tests passing yet, but at this point I suspect it's an issue with using a plain gpytorch kernel outside the context of a giant pile of torch jacobians on a GPU, etc.

PR #409, opened 2023-03-01
Add Helper Functions for Leave One Group Out Cross Validation

akleeman:
I want to be able to take a prediction of some data,
```
const auto pred = model.predict(dataset.features).joint();
```
And then compute leave one group out predictions on the result,
```
const auto loo_preds = leave_one_group_out_joints(pred, dataset.targets, group_inds);
```
which is technically already possible (put `pred` in a `ConditionalGaussian` model and then cross validate), but doing so would not take advantage of the same linear algebra trick we use in the GP implementation. 

In this PR I've moved the leave one group out cross validation routines out of `gp.hpp` and into `cross_validation_utils.hpp` and added some helper functions which make it possible to efficiently compute leave one group out predictions using free functions.

In the process I realized that the cross validation trick isn't very well documented, so I added it to the readthedocs pages:

image

image

### DIFF
--- a/doc/src/gp-details.rst
+++ b/doc/src/gp-details.rst
@@ -176,3 +176,85 @@ would:
 * Evaluate the prior covariance :math:`K_{**}`
 * Compute the posterior covariance :math:`K_{**} - Q_{f*}^T Q_{f*}`
 
+++++++++++++++++++++++++++++++++++++++++
+Cross Validation Implementation Details
+++++++++++++++++++++++++++++++++++++++++
+
+In `Gaussian Processes for Machine Learning`_ (Section 5.4.2) they describe an efficient way for making leave one out predictions, here we expand that same trick to enable making leave one group out predictions.
+
+Consider the case where we have a set of observations, :math:`y`, and we would like to make leave one group out cross validated predictions and by groups we mean independent sets of one or more variables.
+
+We start with our GP,
+
+.. math::
+
+    \mathbf{f} \sim \mathcal{N}\left(0, \Sigma \right)
+
+Which we can then break into groups,
+
+.. math::
+
+    \begin{bmatrix} \mathbf{\hat{y}} \\ \mathbf{y_i} \end{bmatrix} \sim \mathcal{N}\left(\begin{bmatrix} 0 \\ 0 \end{bmatrix}, \begin{bmatrix}\hat{\Sigma} & S \\ S^T & C \end{bmatrix}\right)
+
+Where we will be using a subset of observations, :math:`\hat{y}` to make predictions for a held out set of locations, :math:`x_i`.  We can do this directly using the Gaussian process predict formula,
+
+.. math::
+
+    [\mathbf{y_i}|\mathbf{\hat{y}}=\hat{y}] \sim \mathcal{N}\left(S^T \hat{\Sigma}^{-1} \hat{y}, C - S^T \hat{\Sigma}^{-1} S\right)
+
+But doing so would require computing :math:`\hat{\Sigma}^{-1}` for every group, :math:`i`, that we hold out.  So if we're doing leave one out with :math:`n` observations we have to do the :math:`\mathcal{O}(n^3)` inversion :math:`n` times leading to :math:`\mathcal{O}(n^4)` complexity which will quickly get infeasible.
+
+However, in the process of fitting our GP we'll need to end up computing the inverse of the full covariance, :math:`\Sigma^{-1}` as well as what we've been calling the information vector, :math:`v = \Sigma^{-1} y`.  By using block inversion we get,
+
+.. math::
+
+    \Sigma^{-1} = \begin{bmatrix}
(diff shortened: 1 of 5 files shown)


doc/src/gp-details.rst
```diff
@@ -176,3 +176,85 @@ would:
...
+    [\mathbf{y_i}|\mathbf{\hat{y}}=\hat{y}] \sim \mathcal{N}\left(S^T \hat{\Sigma}^{-1} \hat{y}, C - S^T \hat{\Sigma}^{-1} S\right)
+
+But doing so would require computing :math:`\hat{\Sigma}^{-1}` for every group, :math:`i`, that we hold out.  So if we're doing leave one out with :math:`n` observations we have to do the :math:`\mathcal{O}(n^3)` inversion :math:`n` times leading to :math:`\mathcal{O}(n^4)` complexity which will quickly get infeasible.
+
+However, in the process of fitting our GP we'll need to end up computing the inverse of the full covariance, :math:`\Sigma^{-1}` as well as what we've been calling the information vector, :math:`v = \Sigma^{-1} y`.  By using block inversion we get,
+
+.. math::
+
+    \Sigma^{-1} = \begin{bmatrix}
+    \left(\hat{\Sigma} - S C^{-1} S^T\right)^{-1} & -\left(\hat{\Sigma} - S C^{-1} S^T\right)^{-1}SC^{-1} \\
```
peddie, line 212: I think there might be some typos in this formula.  We should never see `\hat{\Sigma} - S C^{-1} S^T` as the middle term of any of these blocks -- I think `\Sigma` and `C` are swapped here.

peddie: Also it looks like we are post-multiplying by `S C^{-1}` rather than pre-multiplying by `S \Sigma^{-1}`.

include/albatross/src/evaluation/cross_validation_utils.hpp
```diff
@@ -162,6 +162,104 @@ leave_one_out_conditional(const JointDistribution &prior,
...
+  const auto A_ldlt = Eigen::SerializableLDLT(inverse_block);
+  const Eigen::VectorXd mean = y - A_ldlt.solve(v);
+  return MarginalDistribution(mean, A_ldlt.inverse_diagonal());
+}
+
+inline JointDistribution
+held_out_prediction(const Eigen::MatrixXd &inverse_block,
+                    const Eigen::VectorXd &y, const Eigen::VectorXd &v,
+                    PredictTypeIdentity<JointDistribution>) {
+  const auto A_inv = inverse_block.inverse();
```
peddie, line 192: :O 

1) Can we form the mean by using `.ldlt().solve(v)`?

2) `.inverse()` uses the LU decomposition.  Since we know it's a sub-diagonal of the inverse of a covariance matrix, is there some way we can either not re-invert or preserve the triangle and permutation until we get here?

I would place less importance on this if I hadn't just spent several days trying to find numerically negative eigenvalues and asymmetry.

akleeman: Yeah we can do `.ldlt().solve(v)` ... but we need the actual inverse a few lines below.

doc/src/gp-details.rst
akleeman: There are two different equivalent forms for the block inversion (you either form schur complements over the top left or bottom right block), https://en.wikipedia.org/wiki/Block_matrix#Block_matrix_inversion. I was picking and choosing between them to get terms which cancel easier.

include/albatross/src/evaluation/cross_validation_utils.hpp
peddie: That actual inverse is the inverse of an inverse, right?

doc/src/gp-details.rst
peddie: Thanks!  I didn't think about this.

include/albatross/src/evaluation/cross_validation_utils.hpp
akleeman: I see, I'd have to think about that a bit more. Is the inverse of a block of the inverse just a block  of the original matrix (that's a mouthful)? I do see how it seems we could avoid a step, but I'd have to do some proof by python to know for sure.

peddie: The inverse of a block of the inverse is definitely not just a block of the original matrix, sadly.  I just wondered whether there is a way to either avoid inverting again or at least use parts of the LDL decomp of the original matrix to form this inverse, rather than pivoting and LU decomposing from scratch.

akleeman: Any actions we want to take on this? If there's a suggestion for a simple change to improve things I'm happy to do it, but don't think refactoring the cross validation code to return decompositions is a good use of time at the moment.

peddie: OK no problem.

peddie (approved):

PR #421, opened 2023-05-18
Async threads should play nicely with MKL

peddie:
MKL uses OpenMP threads under the hood; we use our own C++11 thread pool.  When requested to set up a thread pool, the easiest approach is to set MKL to use 1 (OpenMP) thread for any operations in async worker threads.  This should avoid oversubscribing CPU resources.


akleeman (approved):
lgtm!

Issue #419, opened 2023-05-18
Failed Eigen assertion when testing with MKL enabled


peddie:
```
[ RUN      ] test_core_dataset/DatasetOperatorTester/0.test_multiply_dataset
albatross_unit_tests: albatross/third_party/eigen/Eigen/src/Core/Assign_MKL.h:132: static void Eigen::internal::Assignment<Eigen::Array<double, -1, 1, 0>, Eigen::CwiseUnaryOp<Eigen::internal::scalar_sqrt_op<double>, const Eigen::ArrayWrapper<const Eigen::Matrix<double, -1, 1, 0>>>, Eigen::internal::assign_op<double, double>, Eigen::internal::Dense2Dense>::run(DstXprType &, const Eigen::internal::Assignment<type-parameter-0-0, CwiseUnaryOp<Eigen::internal::scalar_sqrt_op<double>, type-parameter-0-1>, Eigen::internal::assign_op<double, double>, Eigen::internal::Dense2Dense, typename enable_if<vml_assign_traits<DstXprType, SrcXprNested>::EnableVml, void>::type>::SrcXprType &, const assign_op<double, double> &) [DstXprType = Eigen::Array<double, -1, 1, 0>, SrcXprType = Eigen::CwiseUnaryOp<Eigen::internal::scalar_sqrt_op<double>, const Eigen::ArrayWrapper<const Eigen::Matrix<double, -1, 1, 0>>>, Functor = Eigen::internal::assign_op<double, double>, Kind = Eigen::internal::Dense2Dense, EnableIf = void]: Assertion `dst.rows() == src.rows() && dst.cols() == src.cols()' failed.
Aborted
```
Looks like a mistaken assumption about sizes somewhere.  This failure also occurs in:
 - `test_multiply_with_matrix_joint`
 - `test_multiply_with_matrix_marginal`
 - `test_multiply_with_sparse_matrix_joint`
 - `test_multiply_with_sparse_matrix_marginal`
 - `test_multiply_with_vector`

peddie:
This is triggered by the matrix with 0 rows at https://github.com/swift-nav/albatross/blob/master/tests/test_core_dataset.cc#L193:
```
struct EmptyTransform {
  auto get_matrix(Eigen::Index n) {
    Eigen::SparseMatrix<double> matrix(0, n);
    return matrix;
  }
};
```
[...]

PR #422, opened 2023-05-18
Add bazel --config=mkl

john-michaelburke:
## Implements
* Uprevs rules_swiftnav version.
* Adds MKL config which triggers Eigen linkage with intel mkl for linux.

## Testing
* The unittests fail with this configuration but we can address them in a future PR.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,5 +1,10 @@
 build --repo_env=BAZEL_CXXOPTS="-std=c++14"
 
+# Build with mkl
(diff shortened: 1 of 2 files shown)


peddie:
I don't think the failure is due to problems with async calculation.  I would look into https://github.com/swift-nav/albatross/issues/419 since the failure occurred in the same unit test.

akleeman (approved):
lgtm

PR #423, opened 2023-05-18
Replace ad-hoc DiagonalWrapper with built-in Eigen approach

peddie:
As far as I can tell, this is mathematically and operationally equivalent, but using Eigen's built-in functionality for producing a diagonal matrix expression from a 1-d array solves issue https://github.com/swift-nav/albatross/issues/419.

### DIFF
--- a/include/albatross/src/core/transformed_distribution.hpp
+++ b/include/albatross/src/core/transformed_distribution.hpp
@@ -36,17 +36,12 @@ namespace albatross {
 
 namespace details {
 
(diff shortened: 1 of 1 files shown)


john-michaelburke (approved):
Good hunting!

peddie:
When building with `g++`, I still see the warning `Intel MKL ERROR: Parameter 6 was incorrect on entry to DGEMV .`  I'm not clear whether this is a problem for us in practice if the results for the test are still computed correctly.

akleeman (approved):
lgtm!

PR #426, opened 2023-06-01
[WIP] Use zstd compression for serialization of data

peddie:
Previously the serialization routines used gzip.  This commit modifies them to use zstd, which pareto-dominates gzip apart from breadth of distribution.


peddie:
See https://github.com/swift-nav/albatross/pull/451

PR #427, opened 2023-06-02
[WIP] use zfp compression for serialization of matrix data

peddie:
All tests pass, but this has not been benchmarked.


peddie:
This was a total bust, zfp is super slow in lossless mode.

PR #428, opened 2023-06-02
[WIP] Use blosc2 compression for serialization of matrix data

peddie:
Passes all tests, but not benchmarked yet.


peddie:
This worked OK, but the minor space savings were judged not worth the time cost or infrastructure overhead.

PR #429, opened 2023-06-06
Added last_value() function

wlanghans:
`group_by.hpp` is given a function named last_value() in analogy to the existing first_value() function. The new function returns the last indices.

The PR has been tested as part of [this orion engine PR](https://github.com/swift-nav/orion-engine/pull/6390/)

Further, a unit test has been added. 

This PR also fixes a bug in the `number_of_digits` function. `number_of_digits(1.)` would return 0 and `number_of_digits(1000.)` would return 3 and this was fixed by redefining the function.


akleeman (commented):
looks good, it seems like we don't have great unit test coverage of these methods, mind expanding those a bit?

We already have [a test](https://github.com/swift-nav/albatross/blob/6b49f69410bec4e339c00b74c548c12922bdb284/tests/test_group_by.cc#L458) that checks that `x.group_by(...).first_group() == x.group_by(...).groups().first_group()` but I couldn't find anything that actually checks that first/last work the way we think. It'd be nice to have something like:

```
std::map<int, std::string> example = {
  {1, "one"},
  {2, "two"},
  {0, "zero"}
};

albatross::Grouped<int, std::string> grouped(example);

EXPECT_EQ(grouped.first_value(), "zero");
EXPECT_EQ(grouped.last_value(), "two");
```
there are a lot of edge cases we might want to think about (what happens when the map is empty?) ... but handling all those is a big ask so I think a simple sanity check would be a good start.

tests/test_group_by.cc
```diff
@@ -383,7 +385,7 @@ double mean(const std::vector<double> &xs) {
   return mean / cast::to_double(xs.size());
 }
 
-long int number_of_digits(double x) { return lround(ceil(log10(x))); }
+long int number_of_digits(double x) { return (long int)floorl(log10(x) + 1.); }
```
akleeman, line 388: I see, this was fixing the edge cases were `log(x)` falls on an integer? Nice catch. We should change the casting to `static_cast<long int>(...)` though, or you could stick with the `lround` approach,
```suggestion
long int number_of_digits(double x) { return lround(floor(log10(x))); }
```

tests/test_group_by.cc
```diff
@@ -471,6 +473,40 @@ TEST(test_groupby, test_group_by_first_group) {
...
+
+  const auto grouped_indexers = group_by(fib, number_of_digits).indexers();
+  const auto smallest_number_of_digits_inds = grouped_indexers.first_value();
+  const auto largest_number_of_digits_inds = grouped_indexers.last_value();
+  auto it_min =
+      std::distance(fib.begin(), std::min_element(fib.begin(), fib.end()));
+  auto it_max =
+      std::distance(fib.begin(), std::max_element(fib.begin(), fib.end()));
+  unsigned long it_min_u_l = (unsigned long)it_min;
+  unsigned long it_max_u_l = (unsigned long)it_max;
```
akleeman, line 488: I think that this is checking that `first_value()` returns the value associated with the minimum key ... which works out here because internally we're using `std::map` which sorts by key, but I _think_ I would have expected `first_value()` and `last_value()` to simply return the first and last elements in the container (sorted or not).

Basically, I think if we changed the container to `std::unordered_map<>` instead of `std::map` these tests could fail even if `first_value()` and `last_value()` continued to do the expected thing.
```suggestion
const auto grouped = group_by(fib, number_of_digits);
const unsigned long expected_first = grouped.begin()->second;
const unsigned long expected_last = std::prev(grouped.end())->second;
EXPECT_EQ(expected_first, grouped.first_value());
EXPECT_EQ(expected_last, grouped.last_value());
```

akleeman:
Not for this PR, just a general comment:

looking over the code related to this made me think about the `grouped.min()`, `min_key()` and `min_value()` methods. Those names are a big ambiguous I think. The goal there was to be able to do something like:
```
key_corresponding_to_smallest_group = data.group_by(get_group).apply(count_something).min_key();
```
which means the `min` and `max` methods sort by values. But I can imagine use cases where you want the minimum key. In which case, what would those methods be called? `min_by_key`? what would be the value associated with a minimum key be, `min_value_by_key`? Not great. I wonder if there's a better vocabulary/syntax to make methods like that more intuitive?

Maybe something like?
`grouped.keys.min().value` -- value associated with the smallest key
`grouped.values.min().key` -- key associated with the smallest value

tests/test_group_by.cc
wlanghans: lround rounds up or down, but it should be down only. So I'll implement the casting as 
`long int number_of_digits(double x) {static_cast<long int>(lround(ceil(log10(x))));}`

tests/test_group_by.cc
wlanghans: first_value and last_value return all values assigned to the first and last key respectively. So in my case, which is based on the use case, the indexer has keys for each time, sorted by time. And the map has all indices for those entries that have those times. I agree we are also testing the indexer here, but that was actually my point based on the use case.

wlanghans: The test also tests that all the right indices are chosen from fib. That the first and last value vectors are complete.

wlanghans:
@akleeman I added the snippet you shared to extend the unit test to make sure it works without sorted input.

wlanghans:
Regarding future ideas to rename or repurpose some of the min/max functions. One thought would be to have the `min_value` also report back the index or the key.
Alternatively, renaming `min_key` to argmin (Python) or minloc (Fortran) would seem intuitive.

akleeman:
> @akleeman I added the snippet you shared to extend the unit test to make sure it works without sorted input.

I guess my point was that the tests as is would already fail if the underlying container switched from `std::map` to `std::unordered_map`

For example:
```
  const std::map<int, std::string> example = {{1, "one"}, {2, "two"}, {0, "zero"}};

  for (const auto &pair : example) {
    std::cout << pair.first << std::endl;
  }
```
is going to return
```
0
1
2
```
because `std::map` sorts internally. Inside `albatross::GroupBy` we [store things in a map](https://github.com/swift-nav/albatross/blob/master/include/albatross/src/indexing/group_by.hpp#L165)

include/albatross/src/indexing/group_by.hpp
```diff
@@ -111,6 +111,12 @@ template <typename KeyType, typename ValueType> class GroupedBase {
 
   ValueType first_value() const { return map_.begin()->second; }
```
akleeman, line 112: nit: maybe worth adding a comment here which says:
```
returns the value associated with the minimum key
```
and similar for `last_value`

akleeman (approved):

include/albatross/src/indexing/group_by.hpp
```diff
@@ -109,8 +109,16 @@ template <typename KeyType, typename ValueType> class GroupedBase {
 
   std::pair<KeyType, ValueType> first_group() const { return *map_.begin(); }
 
+  //returns the value associated with the maximum key
```
akleeman, line 112: type: `minimum key`

PR #430, opened 2023-06-08
Test serialization for binary archives

peddie:
This commit runs the existing serialization tests on cereal's `Binary` and `PortableBinary` archive types as well as JSON ones.


akleeman (approved):
lgtm!

PR #431, opened 2023-06-10
Ensure unit tests are always built with assertions enabled

peddie:
This commit modifies the built-in cmake flags to remove `-DNDEBUG` for the test suite.


akleeman (approved):
thanks!

PR #432, opened 2023-06-12
Derive Length Scales

akleeman:
When writing covariance functions it's best to work with parameters which can be directly used into the equations. Consider the squared exponential covariance for example,

$$
k_{se}(d) = \sigma^2 \mbox{exp}\left[-\left(\frac{d}{\ell}\right)^2\right]
$$

In our models we directly store the prior standard deviation, $\sigma$, and length scale, $\ell$. That way evaluating the covariance doesn't require unnecessary computations which could become problematic since covariance functions are often evaluated millions of times. However, the most computationally efficient choice of parameters isn't necessarily the most intuitive. One alternate representation which is often used in Kalman filtering is to define the initial standard deviation and the "process noise" or the amount the standard deviation increases between updates. If we pick a reference distance we can actually compute the equivalent process noise for any radial covariance function by asking for the predictive variance at some location given a known value some distance, $\Delta$, away. To do so we start with the prior:

$$
\begin{bmatrix}
f_0 \\
f_\Delta
\end{bmatrix}
\sim \mathcal{N}\left(0, \begin{bmatrix}k(0) & k(\Delta) \\ k(\Delta) & k(0) \end{bmatrix}\right)
$$

Then we can ask for $\sigma\left(f_\Delta | f_0\right)$,

$$
    \sigma\left(f_\Delta | f_0\right) = \sqrt{k(0) - \frac{k(\Delta)^2}{k(0)}}
$$

If we plug the squared exponential covariance function in we get,

$$
    \sigma_{se}\left(f_\Delta | f_0\right) = \sigma \sqrt{1 - \mbox{exp}\left[-2\left(\frac{\Delta}{\ell}\right)^2\right]}
$$

This offers us a way to go from the less intuitive length scale parameter to a more intuitive increase in standard deviation, $\delta=\sigma\left(f_\Delta | f_0\right)$, over some distance, $\Delta$.

This change adds methods for deriving the length scale required to achieve a given increase in standard deviation over a reference distance for all of the radial covariance functions in albatross (exponential, squared exponential, matern 3/2, matern 5/2).

Note that while the exponential and squared exponential functions had closed form solutions, the Matern covariance functions do not. As a result we use a customized newton root finding algorithm to solve for the length scales. This isn't cheap, but would only need to be used when a parameter gets set, not when the covariance function is evaluated, so should suffice.

### DIFF
--- a/include/albatross/src/covariance_functions/radial.hpp
+++ b/include/albatross/src/covariance_functions/radial.hpp
@@ -18,6 +18,7 @@ constexpr double default_radial_sigma = 10.;
 
 namespace albatross {
 
+constexpr std::size_t MAX_NEWTON_ITERATIONS = 50;
 constexpr double MAX_LENGTH_SCALE_RATIO = 1e7;
 constexpr double MIN_LENGTH_SCALE_RATIO = 1e-7;
 
@@ -90,7 +91,7 @@ inline double derive_squared_exponential_length_scale(double reference_distance,
   //   VAR[f_d|f_0] = k(0) - k(d) k(d) / k(0)
   //   STD[f_d|f_0] = sqrt(k(0) - k(d)^2/ k(0))
   //
-  // for the squared exponential funciton this means an increase in
+  // for the squared exponential function this means an increase in
   // standard deviation, d_sd, can be written:
   //
   //   d_sd = sqrt(sigma^2 - sigma^2 exp[-(d / length_scale)^2]^2)
@@ -100,10 +101,15 @@ inline double derive_squared_exponential_length_scale(double reference_distance,
   // solving for length_scale gives us
   //
   //   d_sd / sigma = sqrt(1 - exp[-2 * (d / length_scale)^2])
+  //
   //   (d_sd / sigma)^2 = 1 - exp[-2 * (d / length_scale)^2]
+  //
   //   exp[-2 * (d / length_scale)^2] = 1 - (d_sd / sigma)^2
+  //
   //   -2 (d / length_scale)^2 = log[1 - (d_sd / sigma)^2]
+  //
   //   (d / length_scale) = sqrt(-1/2 log[1 - (d_sd / sigma)^2])
+  //
   //   length_scale = d / sqrt(-1/2 log[1 - (d_sd / sigma)^2])
   const double ratio = std_dev_increase / prior_sigma;
   assert(ratio > 0.);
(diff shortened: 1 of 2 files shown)


include/albatross/src/covariance_functions/radial.hpp
```diff
@@ -193,11 +432,28 @@ inline double matern_52_covariance(double distance, double length_scale,
...
+  const double sqrt_5_d = sqrt(5.) * distance / length_scale;
   return sigma * sigma * (1 + sqrt_5_d + sqrt_5_d * sqrt_5_d / 3.) *
          exp(-sqrt_5_d);
 }
 
+inline double derive_matern_52_length_scale(double reference_distance,
+                                            double prior_sigma,
+                                            double std_dev_increase) {
+
+  auto func = [&](double ratio) { return matern_52_covariance(1., ratio, 1.); };
```
peddie, line 444: nit: what does this capture by reference (same below)?

tests/test_radial.cc
```diff
@@ -38,6 +38,108 @@ inline auto random_spherical_dataset(std::vector<Eigen::VectorXd> points,
...
+                << " expected: " << expected_sd_increase << std::endl;
+      std::cerr << "FAIL: absolute_sd error: " << absolute_sd_error
+                << " relative_sd_error: " << relative_sd_error << std::endl;
+    }
+    return relative_sd_error <= 1e-3;
+  };
+
+  // large sigmas are likely lead to numerical instabilities
+  // in matrix inversions, so that seems like a reasonable limit;
+  std::vector<double> sigmas = {1e-6, 1e-4, 1e-2, 1, 1e2, 1e4, 1e6};
```
peddie, line 128: nit: `const`?

peddie (approved):
Logic looks great -- just a couple minor formatting / organizational suggestions

tests/test_radial.cc
```diff
@@ -38,6 +38,108 @@ inline auto random_spherical_dataset(std::vector<Eigen::VectorXd> points,
...
+  // extremely close points should be almost perfectly correlated
+  EXPECT_NEAR(this->test_case(M_PI, M_PI + 1e-16), this->test_case(M_PI, M_PI),
+              1e-8);
+  // extremely distant points should be independent
+  EXPECT_EQ(this->test_case(0., 1e32), 0.);
+}
+
+TYPED_TEST(RadialCovarianceTester, test_derive_length_scale) {
+
+  auto set_sigma_length_scale = [&](double sigma, double length_scale) {
```
peddie, line 70: do these capture anything but `this`?  maybe explicitly capture `this` rather than make the reader search for what's captured?

include/albatross/src/covariance_functions/radial.hpp
```diff
@@ -149,10 +274,118 @@ inline double matern_32_covariance(double distance, double length_scale,
...
+  for (std::size_t i = 0; i < 50; ++i) {
+    const double log_f = log_f_eval(guess);
+    const double f_i = log_target - log_f;
+    if (!std::isfinite(f_i)) {
+      break;
+    }
+    double g = log_g_eval(guess);
+    if (!std::isfinite(g) || g == 0.) {
+      g = f_i > 0. ? -1e-8 : 1e-8;
+    };
```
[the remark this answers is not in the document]
akleeman: Ah yeah that actually shouldn't be needed any more, those were some hacks before I realized working in log space was more stable.

PR #438, opened 2023-06-26
ci: use bazel as formatting source of truth

isaactorz:
albatross has no `.clang-format` file so the formatting in CI is based on the default settings of clang-format-14.

This PR makes this implicit config explicit by checking in the output of `clang-format-14 -dump-config` and also sets up bazel to use this file to check formatting.

It also sets up bazel as the "source of truth" which should make it much easier for developers to match the results locally as in CI regardless of platform/local environment setup.


akleeman (approved):
lgtm, thanks!

PR #439, opened 2023-07-12
Set default thread pool smaller to avoid hyperthreading

peddie:


john-michaelburke (approved):

include/albatross/src/utils/async_utils.hpp
```diff
@@ -189,10 +189,21 @@ inline auto apply(const Grouped<KeyType, ValueType> &map, ApplyFunction &&f,
...
+// try to return the actual number of hardware cores, not the number
+// of hardware threads.
 inline std::size_t get_default_thread_count() {
   // This standard function is not guaranteed to return nonzero.
   return std::max(std::size_t{1},
-                  std::size_t{std::thread::hardware_concurrency()});
+#if defined(EIGEN_USE_MKL_ALL) || defined(EIGEN_USE_MKL_VML)
+                  std::size_t{mkl_get_max_threads()}
+#else  // EIGEN_USE_MKL_ALL || EIGEN_USE_MKL_VML
+                  std::size_t{std::thread::hardware_concurrency()} / 2
```
jbangelo, line 204: Potential future improvement would be to do some magic with the [`CPUID`](https://en.wikipedia.org/wiki/CPUID) instruction to detect number of cores and if Hyperthreading is available/enabled.

jbangelo (approved):

akleeman:
do we still want this in?

peddie:
Sorry, I thought I had replied to this several days ago, but it got lost in the cloud somewhere.

> do we still want this in?

It's hard for me to tell.  I don't think its behaviour will hurt anything, since albatross is mainly used in production on x86 machines as far as we know; I think it's a reasonable default.

On the other hand, I think the original problem motivating it went away (@john-michaelburke?), and if not, I'd ideally like to
 - verify that this change fixes the original issue
 - confer with the Kubernetes crew about whether this is a step in the right direction long-term, or whether there is a better way to deal with the broader issue of calculations that are harmed by hyperthreading
 
Personally I am super careful about explicit thread configuration when developing locally, since our development cycle is so long that I want to always leave a few cores free for back-burner tasks when a model is running, but maybe this behaviour would help others avoid some extra workflow steps?

I'm going to close it for now, but I don't mind bringing it back if anyone wants it.

john-michaelburke:
> Sorry, I thought I had replied to this several days ago, but it got lost in the cloud somewhere.
> 
> > do we still want this in?
> 
> It's hard for me to tell. I don't think its behaviour will hurt anything, since albatross is mainly used in production on x86 machines as far as we know; I think it's a reasonable default.
> 
> On the other hand, I think the original problem motivating it went away (@john-michaelburke?), and if not, I'd ideally like to
> 
> * verify that this change fixes the original issue
> * confer with the Kubernetes crew about whether this is a step in the right direction long-term, or whether there is a better way to deal with the broader issue of calculations that are harmed by hyperthreading
> 
> Personally I am super careful about explicit thread configuration when developing locally, since our development cycle is so long that I want to always leave a few cores free for back-burner tasks when a model is running, but maybe this behaviour would help others avoid some extra workflow steps?
> 
> I'm going to close it for now, but I don't mind bringing it back if anyone wants it.

I'll take a stab at setting this up this week.

PR #441, opened 2023-07-13
Add float comparison for MarginalDistribution 

lkloh:


include/albatross/src/core/distribution.hpp
```diff
@@ -100,7 +100,28 @@ struct MarginalDistribution : public DistributionBase<MarginalDistribution> {
   }
 
   bool operator==(const MarginalDistribution &other) const {
-    return (mean == other.mean && covariance == other.covariance);
+    const double epsilon = 0.1;
```
lkloh, line 103: Had to set it to be fairly loose to support https://github.com/swift-nav/orion-engine/pull/6673 

Any concerns about it? I could just write a custom comparator that lets people pass in `epsilon` if there are

akleeman: I fear people might assume that `x == y` would provide a strict equality, but I do see the value in enabling inexact equality checks. What about having the syntax be something like:
```
x.approx_equal(y, 1e-6)
```
?

include/albatross/src/core/distribution.hpp
```diff
@@ -100,7 +100,28 @@ struct MarginalDistribution : public DistributionBase<MarginalDistribution> {
...
+      return false;
+    }
+    for (int i = 0; i < covariance.diagonal().size(); i++) {
+      if (fabs(covariance.diagonal()[i] - other.covariance.diagonal()[i]) >
+          epsilon) {
+        return false;
+      }
+    }
+
+    return true;
```
akleeman, line 124: I'm not really sure what this will do with different sized distributions ... but you could take advantage of Eigen's helpers:
```suggestion
    const bool mean_approx_equal = mean.isApprox(other.mean, epsilon);
    const bool cov_approx_equal = covariance.diagonal().isApprox(other.covariance.diagonal(), epsilon);
    return mean_approx_equal && cov_approx_equal;
```

include/albatross/src/core/distribution.hpp
lkloh: Done

include/albatross/src/core/distribution.hpp
```diff
@@ -99,8 +99,17 @@ struct MarginalDistribution : public DistributionBase<MarginalDistribution> {
...
+    const bool mean_approx_equal = mean.isApprox(other.mean, epsilon);
+    const bool cov_approx_equal =
+        covariance.diagonal().isApprox(other.covariance.diagonal(), epsilon);
+    return mean_approx_equal && cov_approx_equal;
+  }
+
   bool operator==(const MarginalDistribution &other) const {
-    return (mean == other.mean && covariance == other.covariance);
+    // Use the smallest possible value to approximate
+    return approximately_equal(other, std::numeric_limits<double>::epsilon());
```
lkloh, line 112: The mean and covariance used by `MarginalDistribution` are of `double` type

include/albatross/src/core/distribution.hpp
```diff
@@ -99,8 +99,17 @@ struct MarginalDistribution : public DistributionBase<MarginalDistribution> {
     return covariance.diagonal()[i];
   }
 
+  bool approximately_equal(const MarginalDistribution &other,
+                           const double epsilon = 1e-3) const {
```
akleeman, line 103: nit: let's pull out the default value:
```
constexpr double default_approximately_equal_epsilon = 1e-3;
```
and remove the `const` since we're passing in `epsilon` by value.
```suggestion
                           double epsilon = default_approximately_equal_epsilon) const {
```

tests/test_core_distribution.cc
```diff
@@ -140,6 +140,48 @@ TYPED_TEST_P(DistributionTest, test_multiply_by_scalar) {
...
+  }
+
+  {
+    const MarginalDistribution dist1(mean, var);
+    const MarginalDistribution dist2(mean + Eigen::VectorXd::Constant(n, 1e-5),
+                                     var);
+    EXPECT_FALSE(dist1.approximately_equal(
+        dist2, std::numeric_limits<double>::epsilon()));
+    EXPECT_FALSE(dist1.approximately_equal(dist2, 1e-6));
+  }
```
akleeman, line 182: The `DistributionTest` case is a helper to enable the testing of all possible types of distributions (`MarginalDistribution` and `JointDistribution`).

https://github.com/swift-nav/albatross/blob/master/tests/test_core_distribution.cc#L208

To use the `DistributionTest` fixture you create a distribution [like this](https://github.com/swift-nav/albatross/blob/master/tests/test_core_distribution.cc#L145C1-L146C40
):
```
  TypeParam test_case;
  const auto dist = test_case.create();
```
and then the test you write will get executed on all current and future types. So you'll probably want to do something like this:

```suggestion
  TypeParam test_case;
  const auto dist = test_case.create();
  {
    auto copied = dist;
    EXPECT_TRUE(dist.approximately_equal(
        copied, std::numeric_limits<double>::epsilon()));
  }

  {
    auto perturbed = dist;
    purturbed.mean += Eigen::VectorXd::Constant(n, 1e-4);
    EXPECT_FALSE(dist.approximately_equal(
        purturbed, std::numeric_limits<double>::epsilon()));
    EXPECT_TRUE(dist.approximately_equal(purturbed, 1e-3));
  }
```
Note that for that to pass you'll need to have `approximately_equal` work for both `MarginalDistribution` and `JointDistribution`, which is probably a good idea anyway since these distribution classes share the same base class and most other methods ... but could arguably be out of scope.

An alternative if you don't want to add the corresponding method to `JointDistribution` is to remove the `DistributionTest` and just do,
```
TEST(test_core_distribution, test_marginal_distribution_approximately_equal) {
```

include/albatross/src/core/distribution.hpp
akleeman, line 112: I think @RReichert  should weigh in here since I think he has a stronger sense of best programming practices than I do.

On one hand treating two distributions equal if their mean and covariance are the same down to float precision seems reasonable to me. On the other, I think it's generally assumed that equality operators will return true only if the two objects are _exactly_ the same.  In other words, maybe `dist_x == dist_y` should behave the same way `double_x == double_y` does (tiny difference lead to inequality). and a user should be choosing `dist_x.approximately_equal(dist_y)` if they don't care about perfect equality?

lkloh: Hmm reading this again, I think having `==` mean exactly equal makes sense, especially since we have `approximately_equal` for float point equality. 

Reverted

include/albatross/src/core/distribution.hpp
```diff
@@ -99,8 +101,17 @@ struct MarginalDistribution : public DistributionBase<MarginalDistribution> {
     return covariance.diagonal()[i];
   }
 
+  bool approximately_equal(
+      const MarginalDistribution &other,
+      const double epsilon = cDefaultApproximatelyEqualEpsilon) const {
```
akleeman, line 106: nit: no const needed here to follow coding style.

include/albatross/src/core/distribution.hpp
```diff
@@ -165,8 +176,17 @@ struct JointDistribution : public DistributionBase<JointDistribution> {
     return covariance.diagonal()[i];
   }
 
+  bool approximately_equal(
+      const JointDistribution &other,
+      const double epsilon = cDefaultApproximatelyEqualEpsilon) const {
+    const bool mean_approx_equal = mean.isApprox(other.mean, epsilon);
+    const bool cov_approx_equal =
+        covariance.diagonal().isApprox(other.covariance.diagonal(), epsilon);
```
akleeman, line 184: I think we should drop the `diagonal()` here, so it's just:
```
covariance.isApprox(other.covariance, epsilon);
```

akleeman (approved):
looks good, just one small change (and a nit), feel free to merge after!

tests/test_core_distribution.cc
lkloh: I decided to add the `approximately_equal` for both.

include/albatross/src/core/distribution.hpp
lkloh: Yeah. just saw the Joint type could be populated everywhere in the cov matrix

PR #444, opened 2023-08-01
Fix GCC 13.1.1 missing cstdlib errors

RReichert:
# Changes

The `gzip-hpp` library had an issue that fails in GCC 13.1.1 (rightfully so). The changes made here are to now point at our own fork of the `gzip-hpp` library with the applied fix.

See https://github.com/swift-nav/gzip-hpp/pull/1 for details.


akleeman (approved):
lgtm

PR #445, opened 2023-08-01
Use copy_if for filtering

akleeman:
Switch to standard library tools for filtering.


include/albatross/src/indexing/filter.hpp
```diff
@@ -24,11 +24,8 @@ template <typename ToKeepFunction, typename ValueType,
 inline std::vector<ValueType> filter(const std::vector<ValueType> &values,
                                      ToKeepFunction &&to_keep) {
   std::vector<ValueType> output;
-  for (const auto &v : values) {
-    if (to_keep(v)) {
-      output.emplace_back(v);
-    }
-  }
+  std::copy_if(output.begin(), output.end(), std::back_inserter(output),
```
jbangelo, line 27:
```suggestion
  std::copy_if(values.begin(), values.end(), std::back_inserter(output),
```

jbangelo (approved):

PR #448, opened 2023-08-14
Remove map_get_or_construct in favor of map_at_or

akleeman:
The existing `map_get_or_construct` method was unused, not super efficient (it'll end up doing two `find` operations) and doesn't let you provide an alternate value.

This PR changes the name and functionality to allow a user provided alternative if a key is not found in a map.


JADC362 (approved):
This is really handy! LGTM!

PR #449, opened 2023-08-14
Add vector_contains Helper

akleeman:
We already have `map_contains`, this adds the equivalent for vectors.


include/albatross/src/utils/vector_utils.hpp
```diff
@@ -49,7 +49,7 @@ inline bool all(const std::vector<bool> &xs) {
 }
 
 inline bool any(const std::vector<bool> &xs) {
-  // The thinking on having any({}) return false comes from interpreting "amy"
+  // The thinking on having any({}) return false comes from interpreting "any"
```
peddie, line 52: Poor Amy!

peddie (approved):

PR #451, opened 2023-08-28
binary serialization and zstd compression for Eigen types [OC-293]

peddie:
Currently `albatross` serializes large vectors and matrices by
 - compressing the binary data with gzip
 - base64-encoding the compressed data
 - serializing to JSON

In a head-to-head test on systems internal to Swift, we were able to reduce serialization time to about 0.4 of the current approach by using Cereal's portable binary archive type and compressing with zstd.  This patch switches Albatross to use the latter approach.


WORKSPACE.bazel
```diff
@@ -4,8 +4,8 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
     
 http_archive(
     name = "rules_swiftnav",
-    strip_prefix = "rules_swiftnav-e79fa5d8d8cbb19974877397bf045822af9a5125",
-    url = "https://github.com/swift-nav/rules_swiftnav/archive/e79fa5d8d8cbb19974877397bf045822af9a5125.tar.gz",
+    strip_prefix = "rules_swiftnav-70823deec53c70820bffc3ae4cb295c88bbdd735",
```
isaactorz, line 7: Needs to be updated to non-branch commit before merging.

peddie: thanks!

include/albatross/src/utils/compress.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+inline std::string compress(const void *data, std::size_t size_bytes,
+                            int compression_level) {
+  const std::size_t compressed_bound = ZSTD_compressBound(size_bytes);
+
+  std::string compressed{};
+  compressed.resize(compressed_bound);
+
+  const std::size_t compressed_size =
+      // I solemnly swear that I am up to no good.
```
john-michaelburke, line 30: Maybe describe the situation you think could cause a failure here and/or why it won't happen, in place of this and the below comments.

tests/test_compression.cc
```diff
@@ -0,0 +1,197 @@
...
+          albatross::zstd::maybe_decompress(compressed, &maybe_outputs);
+      EXPECT_TRUE(result);
+      EXPECT_EQ(inputs, maybe_outputs);
+    }
+  }
+}
+
+TEST(Compress, DecompressEmpty) {
+  std::string inputs;
+  ASSERT_DEATH({ const auto result = albatross::zstd::decompress(inputs); },
```
john-michaelburke, line 70: Is this something we would want to assert on? One recent issue that has come up in the AQ, the current solution is to return an empty model instead of asserting. Could this cause issues on the receiving end? @jbangelo 
https://github.com/swift-nav/orion-engine/pull/6939

john-michaelburke (approved):
Add a description to the PR and take a look at my comments. Otherwise LGTM!

akleeman (approved):
lgtm

include/albatross/src/utils/compress.hpp
peddie: Lol, I'll remove those.

tests/test_compression.cc
peddie: I'm not sure if we have a better way to handle failures here.  These routines get used in the `cereal` serialisation / deserialisation routines, and I'm not aware of a way to propagate deserialisation failures via the `cereal` API.

The code this is replacing also asserted on any detected decompression failure, so for this change I chose to preserve that behavior.

akleeman (approved):
lgtm, nit: the title mostly says it, but the PR is missing a description (why the switch to zstd?)

PR #453, opened 2023-09-19
Use suitesparse QR for sparse GP models [OC-310]

peddie:
When the observation matrix (covariance between measurements to inducing points) is sufficiently sparse, a fit or update to a `SparseGP` model can be computed significantly more quickly using a sparse solver.  This patch adds support for doing this calculation with the multifrontal SPQR solver from Suitesparse rather than with Eigen's built-in Householder QR routine.

## Measurements

On a Swift model with fill ratio around 0.3, around 10k measurements and 1k inducing points, adopting SPQR alone provided over 4x speedup in the _O(N^3)_ solve step.  This increased to nearly 7x when SPQR was instructed to use 8 threads internally.  In combination with MKL, the speedup vs. the original dense QR approached 9x (again with 8 threads).

Your results may vary depending on your problem, architecture and more.

### DIFF
--- a/include/albatross/src/cereal/eigen.hpp
+++ b/include/albatross/src/cereal/eigen.hpp
@@ -139,6 +139,74 @@ inline void load_lower_triangle(Archive &archive,
   }
 }
 
+template <class Archive, class _Scalar, int _Options, typename _StorageIndex>
+inline void save(Archive &ar,
+                 Eigen::SparseMatrix<_Scalar, _Options, _StorageIndex> const &m,
+                 const std::uint32_t version ALBATROSS_UNUSED) {
+  const _StorageIndex rows = m.rows();
+  const _StorageIndex cols = m.cols();
+  const std::size_t nnz = albatross::cast::to_size(m.nonZeros());
+  const std::size_t size_bytes = nnz * sizeof(_Scalar);
+
+  std::vector<Eigen::Triplet<_Scalar>> elems;
+  elems.reserve(nnz);
+  for (_StorageIndex col = 0; col < m.outerSize(); ++col) {
+    for (typename Eigen::SparseMatrix<_Scalar, _Options,
+                                      _StorageIndex>::InnerIterator it(m, col);
+         it; ++it) {
+      elems.emplace_back(it.col(), it.row(), it.value());
+    }
+  }
+
+  std::string payload =
+      gzip::compress(reinterpret_cast<const char *>(elems.data()), size_bytes);
+  if (::cereal::traits::is_text_archive<Archive>::value) {
+    payload =
+        base64::encode(reinterpret_cast<const unsigned char *>(payload.data()),
+                       payload.size());
+  }
+
+  ar(CEREAL_NVP(rows));
+  ar(CEREAL_NVP(cols));
+  ar(CEREAL_NVP(nnz));
+  ar(CEREAL_NVP(payload));
+}
+
+template <class Archive, class _Scalar, int _Options, typename _StorageIndex>
+inline void load(Archive &ar,
+                 Eigen::SparseMatrix<_Scalar, _Options, _StorageIndex> &m,
+                 const std::uint32_t version ALBATROSS_UNUSED) {
+  _StorageIndex rows = 0;
+  _StorageIndex cols = 0;
+  std::size_t nnz = 0;
+  std::string payload;
+  ar(CEREAL_NVP(rows));
+  ar(CEREAL_NVP(cols));
+  ar(CEREAL_NVP(nnz));
+  ar(CEREAL_NVP(payload));
+  const std::size_t size_bytes = nnz * sizeof(_Scalar);
+
+  if (::cereal::traits::is_text_archive<Archive>::value) {
+    payload = base64::decode(payload);
+  }
+
+  const std::string decompressed =
+      gzip::decompress(payload.data(), payload.size());
+
+  ALBATROSS_ASSERT(size_bytes == decompressed.size());
+  m.resize(rows, cols);
+
+  if (size_bytes == 0) {
+    m.setZero();
+  } else {
+    std::vector<Eigen::Triplet<_Scalar>> elems(nnz);
+    std::memcpy(elems.data(), decompressed.data(), size_bytes);
+    m.setFromTriplets(elems.begin(), elems.end());
+  }
+
+  m.makeCompressed();
+}
+
 } // namespace cereal
 
 #endif
--- a/include/albatross/src/cereal/suitesparse.hpp
+++ b/include/albatross/src/cereal/suitesparse.hpp
@@ -56,10 +56,10 @@ inline void serialize(Archive &ar, cholmod_common::cholmod_method_struct &cms,
 namespace detail {
 
 template <class Archive>
-inline void encode_array(Archive &ar, const char *name, void *source,
+inline void encode_array(Archive &ar, const char *name, const void *source,
                          std::size_t size_bytes) {
-  std::string payload =
-      gzip::compress(reinterpret_cast<const char *>(source), size_bytes);
+  std::string payload = albatross::zstd::internal::compress(
+      source, size_bytes, albatross::zstd::kDefaultCompressionLevel);
   if (::cereal::traits::is_text_archive<Archive>::value) {
     payload =
         base64::encode(reinterpret_cast<const unsigned char *>(payload.data()),
@@ -77,10 +77,7 @@ inline void decode_array(Archive &ar, const char *name, void *destination,
   if (::cereal::traits::is_text_archive<Archive>::value) {
     payload = base64::decode(payload);
   }
-  const std::string decompressed =
-      gzip::decompress(payload.data(), payload.size());
-  ALBATROSS_ASSERT(decompressed.size() == expected_bytes);
-  std::memcpy(destination, decompressed.data(), decompressed.size());
+  albatross::zstd::internal::decompress(payload, destination, expected_bytes);
 }
 
 // This exists because 1) the suitesparse memory functions require a
@@ -408,6 +405,8 @@ inline void load(Archive &ar, cholmod_sparse &m,
   }
   detail::suitesparse_cereal_realloc(&m.x, m.nzmax, element_size_bytes);
 
+  std::cout << "m.p size: " << p_size_bytes << " bytes" << std::endl;
+  std::cout << "m.ncol + 1: " << m.ncol + 1 << std::endl;
   detail::decode_array(ar, "m.p", m.p, (m.ncol + 1) * p_size_bytes);
   detail::decode_array(ar, "m.i", m.i, m.nzmax * integer_size_bytes);
   if (!m.packed) {
--- a/include/albatross/src/cereal/suitesparse.hpp
+++ b/include/albatross/src/cereal/suitesparse.hpp
@@ -288,7 +288,7 @@ inline void load(Archive &ar, cholmod_common &cc,
                        (cc.nrow + 1) * cholmod_int_size);
   detail::decode_array(ar, "cc.Xwork", cc.Xwork, cc.xworksize * sizeof(double));
 
-  cc.Iwork = cholmod_malloc(cc.iworksize, cholmod_int_size, &cc);
+  cc.Iwork = cholmod_l_malloc(cc.iworksize, cholmod_int_size, &cc);
   ar(CEREAL_NVP(cc.itype));
   ar(CEREAL_NVP(cc.dtype));
   ar(CEREAL_NVP(cc.no_workspace_reallocate));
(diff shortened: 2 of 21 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -370,9 +370,10 @@ class SparseGaussianProcessRegression
   // };
 
   // TODO(@peddie) SPQR configuration
```
akleeman, line 372: what does this TODO mean?

peddie: Hmm . . . that's a good question.  I think it's outdated, since `SPQR_create()` sets pivot tolerance and thread count.

include/albatross/src/cereal/eigen.hpp
```diff
@@ -147,6 +147,74 @@ inline void load_lower_triangle(Archive &archive,
...
+  for (_StorageIndex col = 0; col < m.outerSize(); ++col) {
+    for (typename Eigen::SparseMatrix<_Scalar, _Options,
+                                      _StorageIndex>::InnerIterator it(m, col);
+         it; ++it) {
+      elems.emplace_back(it.col(), it.row(), it.value());
+    }
+  }
+
+  std::string payload =
+      gzip::compress(reinterpret_cast<const char *>(elems.data()), size_bytes);
```
akleeman, line 170: I guess we can followup, but curious if this should end up using zstd?

include/albatross/src/cereal/eigen.hpp
```diff
@@ -147,6 +147,74 @@ inline void load_lower_triangle(Archive &archive,
...
+  ar(CEREAL_NVP(nnz));
+  ar(CEREAL_NVP(payload));
+  const std::size_t size_bytes = nnz * sizeof(_Scalar);
+
+  if (::cereal::traits::is_text_archive<Archive>::value) {
+    payload = base64::decode(payload);
+  }
+
+  const std::string decompressed =
+      gzip::decompress(payload.data(), payload.size());
```
john-michaelburke, line 202: At the risk of adding more work for you, when you have clearly put a lot of work into this PR already :grimacing: 

It seems like in another PR you have up, you switched to zstd compression in place of gzip. Are the benefits of gzip over zstd situational or was there another reason? I saw in your other PR you used a zstd::maybe_decompress and then would fall back on gzip, is that something we could do here and force all compression to use zstd?

john-michaelburke (approved):

include/albatross/src/cereal/eigen.hpp
peddie: This just uses gzip because it's not rebased on the other change.  Depending on which goes in first, I'll update this code accordingly.  We will use zstd in the end.

include/albatross/src/cereal/eigen.hpp
```diff
@@ -147,6 +147,74 @@ inline void load_lower_triangle(Archive &archive,
...
+  elems.reserve(nnz);
+  for (_StorageIndex col = 0; col < m.outerSize(); ++col) {
+    for (typename Eigen::SparseMatrix<_Scalar, _Options,
+                                      _StorageIndex>::InnerIterator it(m, col);
+         it; ++it) {
+      elems.emplace_back(it.col(), it.row(), it.value());
+    }
+  }
+
+  std::string payload =
```
akleeman, line 169: encode_array work here?

include/albatross/src/cereal/suitesparse.hpp
```diff
@@ -0,0 +1,486 @@
...
+  for (std::size_t m = 0; m < max_methods; ++m) {
+    ar(cc.method[m]);
+  }
+  // method
+  ar(CEREAL_NVP(cc.postorder));
+  ar(CEREAL_NVP(cc.default_nesdis));
+  ar(CEREAL_NVP(cc.metis_memory));
+  ar(CEREAL_NVP(cc.metis_dswitch));
+  ar(CEREAL_NVP(cc.metis_nswitch));
+  ar(::cereal::make_nvp("nrow", cc.nrow));
```
akleeman, line 147: why does it sometimes use `CEREAL_NVP` and othertimes explicit names?

include/albatross/src/cereal/suitesparse.hpp
```diff
@@ -0,0 +1,486 @@
...
+
+  const std::size_t cholmod_int_size =
+      cc.itype == CHOLMOD_LONG ? sizeof(SuiteSparse_long) : sizeof(int);
+
+  detail::encode_array(ar, "cc.Flag", cc.Flag, cc.nrow * cholmod_int_size);
+  detail::encode_array(ar, "cc.Head", cc.Head,
+                       (cc.nrow + 1) * cholmod_int_size);
+  detail::encode_array(ar, "cc.Xwork", cc.Xwork, cc.xworksize * sizeof(double));
+
+  // Iwork omitted; must reinitialize
```
akleeman, line 162: Not sure I follow this comment?

include/albatross/src/cereal/suitesparse.hpp
```diff
@@ -0,0 +1,486 @@
...
+
+  detail::encode_array(ar, "m.p", m.p, (m.ncol + 1) * p_size_bytes);
+  detail::encode_array(ar, "m.i", m.i, m.nzmax * integer_size_bytes);
+  if (nullptr != m.nz && !m.packed) {
+    detail::encode_array(ar, "m.nz", m.nz, m.ncol * integer_size_bytes);
+  }
+  if (nullptr != m.x) {
+    detail::encode_array(ar, "m.x", m.x, x_elements * element_size_bytes);
+  }
+  // This should only be non-null for "zomplex" elements
```
akleeman, line 401: complex?

include/albatross/src/cereal/suitesparse.hpp
```diff
@@ -0,0 +1,486 @@
...
+    detail::suitesparse_cereal_realloc(&m.nz, m.ncol, integer_size_bytes);
+  }
+  detail::suitesparse_cereal_realloc(&m.x, m.nzmax, element_size_bytes);
+
+  detail::decode_array(ar, "m.p", m.p, (m.ncol + 1) * p_size_bytes);
+  detail::decode_array(ar, "m.i", m.i, m.nzmax * integer_size_bytes);
+  if (!m.packed) {
+    detail::decode_array(ar, "m.nz", m.nz, m.ncol * integer_size_bytes);
+  }
+  // TODO(@peddie): in what cases may `x` be absent?
```
akleeman, line 435: is this a concern? should we add a ticket to follow up?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -317,7 +319,8 @@ class SparseGaussianProcessRegression
       B.col(pi).topRows(i + 1) = old_fit.sigma_R.col(i).topRows(i + 1);
     }
     B.bottomRows(n_new) = A_ldlt.sqrt_solve(K_fu);
-    const auto B_qr = B.colPivHouseholderQr();
+    auto B_qr = SPQR_create(B.sparseView());
```
akleeman: curious how long `SPQR_create` takes, would it be helpful if `B` were already a sparse matrix type?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -350,14 +359,25 @@ class SparseGaussianProcessRegression
   //
   //   Sigma = (B^T B)^-1
   //
-  Eigen::ColPivHouseholderQR<Eigen::MatrixXd>
+  // Eigen::ColPivHouseholderQR<Eigen::MatrixXd>
```
akleeman, line 362: delete?

akleeman (commented):
mostly looks good, handful of nits/questions.

I guess we just need to trust unit tests to know everything was serialized. Are there any other edge cases on the matrix usage that we should check?

Very curious to see how this changes run time on some of our use cases.

akleeman:
could use an improved description

include/albatross/src/cereal/eigen.hpp
peddie: See JM's comment above!

include/albatross/src/cereal/suitesparse.hpp
peddie: no . . . sadly "zomplex" is an internal suitesparse representation.

include/albatross/src/cereal/suitesparse.hpp
peddie: I will check.

include/albatross/src/cereal/eigen.hpp
peddie: It should -- I could expose that to this code; so far it's local to the suitesparse serialisation code.

include/albatross/src/cereal/suitesparse.hpp
peddie: In most of the occurrences, this happens because I'm serialising elements of a fixed-length array.

The reason it happens this way for `nrow`, `iworksize` and `xworksize` is that I want to deserialise them to stack variables so I can calculate allocation sizes etc., rather than deserialising them to elements of `cholmod_common` directly (because initialisation functions can and do modify various elements of `cholmod_common` structures).  I'll add a comment here.

include/albatross/src/cereal/suitesparse.hpp
peddie: This is related to the previous thing -- I don't serialise the workspaces, I just reallocate new ones when I deserialise.  (I confirmed that this is safe both with tests and by reading the SPQR / cholmod code.)  I'll expand the comment to make it clearer.

include/albatross/src/models/sparse_gp.hpp
peddie: It would definitely be helpful, but it significantly complicates assembling `B`, as you can't just assign to block views of sparse matrices.  I don't mind making that change if you'd like to see it.

I haven't done any benchmarks on this whole integrated system yet; my benchmarks were just on solving sparse problems I'd dumped out from a model already in sparse format.  I'll put some head-to-head test results up for the Swift model that uses a sparse GP in that repo.

include/albatross/src/models/sparse_gp.hpp
peddie: Yes, thanks!

akleeman:
How crazy would it be to copy the `sparse_gp.hpp` to a `suitesparse_gp.hpp` (or something like that) to make it easy to switch?

peddie:
> How crazy would it be to copy the `sparse_gp.hpp` to a `suitesparse_gp.hpp` (or something like that) to make it easy to switch and compare performance?

Check the most recent patch.  The changes are much smaller than that.

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -317,7 +341,9 @@ class SparseGaussianProcessRegression
       B.col(pi).topRows(i + 1) = old_fit.sigma_R.col(i).topRows(i + 1);
     }
     B.bottomRows(n_new) = A_ldlt.sqrt_solve(K_fu);
-    const auto B_qr = B.colPivHouseholderQr();
+    auto B_qr = QRImplementation::compute(
+        B, Base::threads_
+               .get()); // SPQR_create(B.sparseView(), Base::threads_.get());
```
akleeman, line 346: nit: remove comment. Should `B_qr` be `const`?

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -349,15 +382,14 @@ class SparseGaussianProcessRegression
...
-  Eigen::ColPivHouseholderQR<Eigen::MatrixXd>
+  std::unique_ptr<Eigen::SPQR<Eigen::SparseMatrix<double>>>
   compute_sigma_qr(const Eigen::SerializableLDLT &K_uu_ldlt,
                    const BlockDiagonalLDLT &A_ldlt,
                    const Eigen::MatrixXd &K_fu) const {
     Eigen::MatrixXd B(A_ldlt.rows() + K_uu_ldlt.rows(), K_uu_ldlt.rows());
     B.topRows(A_ldlt.rows()) = A_ldlt.sqrt_solve(K_fu);
     B.bottomRows(K_uu_ldlt.rows()) = K_uu_ldlt.sqrt_transpose();
-    return B.colPivHouseholderQr();
+    return SPQR_create(B.sparseView(), Base::threads_.get());
```
akleeman, line 392: So we're keeping the QR of sigma sparse for both implementations? or should this be using the `QRImplementation`?

peddie: Oops, I updated the update but not the initial fit!

include/albatross/src/models/sparse_gp.hpp
peddie: Done.

akleeman:
Mind adding a description of the changes (maybe some of the timing comparison plots from the first prototype?) for any future developers.

PR #454, opened 2023-09-19
[SRV-225] standardize thread pool size

RReichert:
# Changes

See: https://github.com/swift-nav/ThreadPool/pull/2 for background information.

The changes here make it so that the thread size is explicit and avoids stack overflow as the stack size differs between platform. This fixes issues on MacOS, among other systems.

### DIFF
--- a/include/albatross/src/utils/async_utils.hpp
+++ b/include/albatross/src/utils/async_utils.hpp
@@ -17,6 +17,8 @@ namespace albatross {
 
 namespace detail {
 
+static constexpr std::size_t cDefaultStackSize{8 * 1024 * 1024};
+
 inline bool should_serial_apply(ThreadPool *pool) {
   return (nullptr == pool) || (pool->thread_count() <= 1);
 }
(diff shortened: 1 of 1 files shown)


include/albatross/src/utils/async_utils.hpp
```diff
@@ -17,6 +17,8 @@ namespace albatross {
 
 namespace detail {
 
+static constexpr std::size_t cDefaultStackSize{8 * 1024 * 1024};
```
RReichert, line 20: defaulting to using GNU Linux's default of 8MiB. In one of the runs GM runs I recorded 2.1MiB of stack usage.

peddie: Is it possible to leave this out of albatross and just expose the ability to set it, then set it appropriately from other programs that use albatross?

RReichert: I'll update the interface to see if we can pass the configuration up the stack

RReichert: all done

peddie (approved):

jbangelo (approved):

PR #456, opened 2023-10-20
Allow dense GP fitting using thread pool

peddie:
We already use a thread pool when available to compute the fit covariance for the SparseGP model; this commit adds the equivalent plumbing to the (dense) GP class.  This yields significant speedups on large, dense GPs, especially if the covariance function is expensive.


akleeman (approved):
lgtm

PR #463, opened 2023-11-03
Add linear_combination::to_linear_combination()

akleeman:
Sometimes you have a single type which may or may not be a `LinearCombination` already and you want to either turn it into a trivial linear combination (ie, length one set of values with coefficient of 1) or just use the existing linear combo. You can now do that with:
```
int foo;
// convert to a combo
LinearCombination<int> foo_combo linear_combination::to_linear_combination(foo);
// already a combo so just return the input
LinearCombination<int> foo_combo linear_combination::to_linear_combination(foo_combo);
```

In the process I added a new namespace `linear_combination::` which contains things like `sum()` to avoid possible conflicts with other `sum` methods.

### DIFF
--- a/include/albatross/src/core/linear_combination.hpp
+++ b/include/albatross/src/core/linear_combination.hpp
@@ -42,6 +42,8 @@ template <typename X> struct LinearCombination {
   Eigen::VectorXd coefficients;
 };
 
+namespace linear_combination {
+
 template <typename X> inline auto sum(const X &x, const X &y) {
   Eigen::Vector2d coefs;
   coefs << 1., 1.;
@@ -73,6 +75,16 @@ inline auto difference(const X &x, const Y &y) {
   variant<X, Y> vy = y;
   return difference(vx, vy);
 }
+
+template <typename X> inline auto to_linear_combination(const X &x) {
+  return LinearCombination<X>({x}, Eigen::VectorXd::Ones(1));
+}
+
+template <typename X>
+inline auto to_linear_combination(const LinearCombination<X> &x) {
+  return x;
+}
+} // namespace linear_combination
 
 template <typename Derived, typename X>
 inline auto operator*(const Eigen::SparseMatrixBase<Derived> &matrix,
--- a/include/albatross/src/core/linear_combination.hpp
+++ b/include/albatross/src/core/linear_combination.hpp
@@ -76,6 +76,10 @@ inline auto difference(const X &x, const Y &y) {
   return difference(vx, vy);
 }
 
+// to_linear_combination(x) lets you construct a LinearCombination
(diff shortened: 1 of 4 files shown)


include/albatross/src/core/linear_combination.hpp
```diff
@@ -74,6 +76,16 @@ inline auto difference(const X &x, const Y &y) {
   return difference(vx, vy);
 }
 
+template <typename X> inline auto to_linear_combination(const X &x) {
+  return LinearCombination<X>({x}, Eigen::VectorXd::Ones(1));
+}
+
+template <typename X>
+inline auto to_linear_combination(const LinearCombination<X> &x) {
```
peddie, line 84: What is the point of this function?

include/albatross/src/core/linear_combination.hpp
```diff
@@ -74,6 +76,16 @@ inline auto difference(const X &x, const Y &y) {
   return difference(vx, vy);
 }
 
+template <typename X> inline auto to_linear_combination(const X &x) {
```
peddie, line 79: Why make this a free function rather than a constructor of `LinearCombination`?

akleeman: One reason is type inference. If, for example, you have some type:
```
SomeWrapper<variant<Foo, Bar>, WowThatsALongTypeName> x;
```
If you want to create a linear combination with that type using a constructor I think you'd have to do:
```
LinearCombination<SomeWrapper<variant<Foo, Bar>, WowThatsALongTypeName>> combo(x);
```
this function let's you do:
```
const auto combo = to_linear_combination(x);
```

include/albatross/src/core/linear_combination.hpp
akleeman: If you aren't sure what the type of `X` is it could be handy to be able to call
```
Foo foo;
LinearCombination<Foo> a = to_linear_combination(foo);
```
and
```
LinearCombination<Foo> already_a_combo;
LinearCombination<Foo> b = to_linear_combination(already_a_combo);
```
without the second returning:
```
LinearCombination<LinearCombination<Foo>>
```

include/albatross/src/core/linear_combination.hpp
peddie: I see.  What I envisioned only works in C++17 and later.

include/albatross/src/core/linear_combination.hpp
peddie: Might this be worth a comment clarifying it for the next reader?  (Possibly me in a year?)

peddie (approved):

PR #464, opened 2023-11-04
Unique Value Utilities

akleeman:
A common problem is to start with some data, and wonder which unique fields the data contains. You might, for example, do
```
struct Foo {
  double bar;
}

std::set<double> unique_bars(const std::vector<Foo> &foos) {
  std::set<double> bars;
  for (const auto &f : foos) {
    bars.emplace(f.bar);
  }
  return bars;
}
```
this PR adds some helper functions which let you do things like:
```
auto get_bar = [](const auto &f) { return f.bar; };
const auto bars = unique_values(foos, get_bar);
```
If you think that all the values in a vector contain a shared field and you'd like to get the value (and assert that there is in fact only one) you can do:
```
auto get_bar = [](const auto &f) { return f.bar; };
const double unique_bar = unique_value(foos, get_bar);
```
or if there are lots of duplicate `Foo` types in the vector (and `Foo` has a `<` operator) you can do:
```
const auto unique_foos = unique_values(foos);
```
the function was named `unique_values` to avoid conflicting with `std::unique` which doesn't quite do the same thing. `std::unique` will return a new iterator which has collapsed neighboring unique values, for example:
```
  std::vector<int> foo = {1, 3, 2, 2, 1, 5};

  std::vector<int> copy(foo);
  const auto last = std::unique(copy.begin(), copy.end());
  copy.erase(last, copy.end());
  // copy = 1,3,2,1,5

  const auto unique_vals = unique_values(foo);
  // unique_vals = 1, 2, 3, 5
```


include/albatross/src/indexing/unique.hpp
```diff
@@ -0,0 +1,68 @@
...
+  return output;
+}
+
+// unique_value
+//   assumes there is one single value in the iterator
+//   and will return that value or assert if that assumption is false
+
+template <typename ValueType>
+inline ValueType unique_value(const std::set<ValueType> &values) {
+  assert(values.size() == 1);
```
peddie, line 44: Minor nit: I don't know whether we have a general policy on this, but so far I've often found asserts like `assert(values.size == 1 && "expected only one unique value")` to be helpful.

peddie (approved):
Good to capture this pattern.

PR #467, opened 2023-11-09
Make unique_values(xs, function) work with sets

akleeman:
We have some utilities for `set<T> unique_values(vector<X>, func)`, but sometimes it's nice to be able to manipulate a set.
```
bool is_even(int x) {
  return x & 2 == 0;
}

std::set<int> xs = {0, 1, 2, 3, 4};
std::set<bool> even_or_odd = unique_values(xs, is_even);
```
I thought about calling this `apply` and having it operate the same way `apply` does on vectors, but I think it would be easy to assume that `apply` won't change the size of a container and I think calling it `unique_values` makes it more obvious the result will be a set.


peddie (approved):
Looks good to me!

PR #476, opened 2024-02-02
Rebase Inducing Points Using QR

akleeman:
The existing `rebase_inducing_points` method relied heavily on the `fit_from_prediction` method. It was computing a prediction of the new inducing points, then `fit_from_prediction` to that result. This involved a few questionable steps which probably aren't a problem ... until you accidentally introduce fully correlated inducing points. When this happens the prediction is probably not full rank, and the `fit_from_prediction` step might have undefined behavior.

This PR more directly rebases the inducing points using one single QR decomposition. A description of the algorithm was added to the documentation. A screenshot is included here:

image


akleeman:
I tried pulling this into our models and it actually significantly changed performance :thinking:. Going to need to root cause why, in the meantime I won't merge this (but don't let that stop you from taking a look with an eye out for any mistakes I may have made).

peddie:
I don't see any obvious performance missteps here.  But Householder QR is always twice as expensive as an equivalently-sized Cholesky decomposition.

I am glad to see the permutation matrix cleanups going in -- might it make sense to pull those out as a separate first PR?  I would happily merge that stuff, and it would also leave fewer changes to examine for performance.

akleeman:
@peddie https://github.com/swift-nav/albatross/pull/475

akleeman:
> I don't see any obvious performance missteps here. But Householder QR is always twice as expensive as an equivalently-sized Cholesky decomposition.

Both this and the existing require LDLT _and_ QR decompositions. But it is true that this version will likely be slower since it requires two LDLTs and a larger QR.

That said, this approach opens the opportunity to do both a rebase and update in one single step (not implemented yet) at which point the cost might even out?

PR #478, opened 2024-02-28
Implement a log error msg

JADC362:
# Task
- [x] Implement an error logging msg if the assert `counter < gauss_legendre_50_points.size()` is false.


include/albatross/src/stats/incomplete_gamma.hpp
```diff
@@ -35,7 +35,12 @@ constexpr double INCOMPLETE_GAMMA_EQUALITY_TRESHOLD = 1e-12;
 
 inline double incomplete_gamma_quadrature_inp_vals(double lb, double ub,
                                                    std::size_t counter) {
-  ALBATROSS_ASSERT(counter < gauss_legendre_50_points.size());
+  if (counter >= gauss_legendre_50_points.size()) {
+    std::cout << "ERROR: Counter value " << counter
```
akleeman, line 39: maybe better to use `cerr`?

akleeman (approved):
lgtm!

PR #483, opened 2024-05-03
Small changes so it builds with cmake

wlanghans:


jbangelo:
@peddie it looks like this is a mix of several WIP branches. Can we close this one in particular?

peddie:
Yes, I don't really remember quite what the goal of this was.

PR #484, opened 2024-05-06
update swift_rules

notoriaga:


jbangelo:
@notoriaga do we still need this change?

PR #487, opened 2024-06-21
Add operator[] for MarginalDistribution

peddie:
This change is in support of the PIC implementation.

### DIFF
--- a/include/albatross/src/core/distribution.hpp
+++ b/include/albatross/src/core/distribution.hpp
@@ -20,6 +20,8 @@ inline bool operator==(const albatross::DiagonalMatrixXd &x,
 
 namespace albatross {
 
+struct MarginalDistribution;
+
 constexpr double cDefaultApproximatelyEqualEpsilon = 1e-3;
 
 template <typename Derived> struct DistributionBase {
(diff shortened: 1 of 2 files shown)


include/albatross/src/core/distribution.hpp
```diff
@@ -132,6 +132,11 @@ struct MarginalDistribution : public DistributionBase<MarginalDistribution> {
     return MarginalDistribution(scale * mean, scale * scale * covariance);
   }
 
+  MarginalDistribution operator[](std::size_t index) const {
```
akleeman, line 135: Nice! Should we move this to `DistributionBase<>` so it's available in `JointDistribution`s as well? (and maybe add a test to this suite: https://github.com/swift-nav/albatross/blob/master/tests/test_core_distribution.cc#L240)

peddie: I can definitely make those changes.  It's a bit odd because for a `JointDistribution` the return type is still a `MarginalDistribution`.

akleeman: Yeah I see, I guess if you want `JointDistribution` to preserve it's type you could return `DistributionBase`. I'm fine either way though.

peddie:
This should be ready to go.

akleeman (approved):

PR #488, opened 2024-06-21
Add wasserstein-2 metric calculation

peddie:
The Wasserstein metric measures the distance between two distributions, accounting for differences in both shape and location.  This can be used to compare two multivariate Gaussians.

This change is in support of PIC.


akleeman (commented):
probably worth a couple quick tests here? otherwise looks good.

peddie:
Tests added.

akleeman (approved):
lgtm

PR #489, opened 2024-06-21
Minor enhancements to custom LDLT implementations

peddie:
This change is in support of PIC.

### DIFF
--- a/include/albatross/src/eigen/serializable_ldlt.hpp
+++ b/include/albatross/src/eigen/serializable_ldlt.hpp
@@ -18,6 +18,10 @@ namespace Eigen {
 // See LDLT.h in Eigen for a detailed description of the decomposition
 class SerializableLDLT : public LDLT<MatrixXd, Lower> {
 public:
+  using RealScalar = double;
+  using Scalar = double;
+  using MatrixType = MatrixXd;
+
   SerializableLDLT() : LDLT<MatrixXd, Lower>(){};
 
   SerializableLDLT(const MatrixXd &x) : LDLT<MatrixXd, Lower>(x.ldlt()){};
@@ -42,6 +46,11 @@ public:
   bool &mutable_is_initialized() { return this->m_isInitialized; }
 
   bool is_initialized() const { return this->m_isInitialized; }
+
+  double l1_norm() const {
+    ALBATROSS_ASSERT(is_initialized() && "Must initialize first!");
+    return this->m_l1_norm;
+  }
 
   /*
    * Computes the inverse of the square root of the diagonal, D^{-1/2}
@@ -88,10 +97,10 @@ public:
    * Computes the product of the square root of A with rhs,
    *   D^{-1/2} L^-1 P rhs
    */
-  template <class Rhs>
-  Eigen::MatrixXd sqrt_solve(const MatrixBase<Rhs> &rhs) const {
+  template <class Rhs> Eigen::MatrixXd sqrt_solve(const Rhs &rhs) const {
     return diagonal_sqrt_inverse() *
-           this->matrixL().solve(this->transpositionsP() * rhs);
+           this->matrixL().solve(this->transpositionsP() *
+                                 Eigen::MatrixXd(rhs));
   }
 
   Eigen::MatrixXd sqrt_transpose() const {
--- a/include/albatross/src/linalg/block_diagonal.hpp
+++ b/include/albatross/src/linalg/block_diagonal.hpp
@@ -32,6 +32,14 @@ struct BlockDiagonalLDLT {
   Eigen::Matrix<_Scalar, _Rows, _Cols>
   sqrt_solve(const Eigen::Matrix<_Scalar, _Rows, _Cols> &rhs) const;
 
+  template <class _Scalar, int _Options, typename _StorageIndex>
+  Eigen::Matrix<_Scalar, Eigen::Dynamic, Eigen::Dynamic>
+  solve(const Eigen::SparseMatrix<_Scalar, _Options, _StorageIndex> &rhs) const;
+
+  template <class _Scalar, int _Options, typename _StorageIndex>
+  Eigen::Matrix<_Scalar, Eigen::Dynamic, Eigen::Dynamic> sqrt_solve(
+      const Eigen::SparseMatrix<_Scalar, _Options, _StorageIndex> &rhs) const;
+
   template <class _Scalar, int _Rows, int _Cols>
   Eigen::Matrix<_Scalar, _Rows, _Cols>
   solve(const Eigen::Matrix<_Scalar, _Rows, _Cols> &rhs,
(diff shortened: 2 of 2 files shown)


include/albatross/src/eigen/serializable_ldlt.hpp
```diff
@@ -88,10 +97,10 @@ class SerializableLDLT : public LDLT<MatrixXd, Lower> {
    * Computes the product of the square root of A with rhs,
    *   D^{-1/2} L^-1 P rhs
    */
-  template <class Rhs>
-  Eigen::MatrixXd sqrt_solve(const MatrixBase<Rhs> &rhs) const {
+  template <class Rhs> Eigen::MatrixXd sqrt_solve(const Rhs &rhs) const {
     return diagonal_sqrt_inverse() *
-           this->matrixL().solve(this->transpositionsP() * rhs);
+           this->matrixL().solve(this->transpositionsP() *
+                                 Eigen::MatrixXd(rhs));
```
akleeman, line 103: will this end up copying `rhs`?

include/albatross/src/linalg/block_diagonal.hpp
```diff
@@ -102,6 +126,55 @@ inline Eigen::Matrix<_Scalar, _Rows, _Cols> BlockDiagonalLDLT::sqrt_solve(
...
+    const auto rhs_chunk = rhs.block(i, 0, b.rows(), rhs.cols());
+    output.block(i, 0, b.rows(), rhs.cols()) = b.solve(rhs_chunk);
+    i += b.rows();
+  }
+  return output;
+}
+
+template <typename Op, typename LhsType, typename RhsType>
+inline Eigen::MatrixXd BlockDiagonalLDLT::solve(
+    const Eigen::CwiseBinaryOp<Op, LhsType, RhsType> &rhs_orig) const {
```
akleeman, line 147: What happens at the moment if you do:
```
Eigen::MatrixXd a, b;
block_diag.solve(a.cwiseProduct(b));
```
I would have guessed it'd find the `solve(Matrix<_Scalar, _Rows, _Cols>)` call [here](https://github.com/swift-nav/albatross/blob/master/include/albatross/src/linalg/block_diagonal.hpp#L29) and do effectively the same thing as these new methods?

What if you change `Matrix<...>` to `MatrixBase<...>`?

include/albatross/src/linalg/block_diagonal.hpp
```diff
@@ -102,6 +126,55 @@ inline Eigen::Matrix<_Scalar, _Rows, _Cols> BlockDiagonalLDLT::sqrt_solve(
...
+  ALBATROSS_ASSERT(cols() == rhs_orig.rows());
+  Eigen::MatrixXd rhs{rhs_orig};
+  return solve(rhs);
```
akleeman, line 150: what happens if you do; `solve(rhs_orig.eval())`?

include/albatross/src/linalg/block_diagonal.hpp
```diff
@@ -157,6 +230,17 @@ inline double BlockDiagonalLDLT::log_determinant() const {
   return output;
 }
 
+inline double BlockDiagonalLDLT::rcond() const {
+  // L1 induced norm is just the maximum absolute column sum.
+  // Therefore the L1 induced norm of a block-diagonal matrix is the
+  // maximum of the L1 induced norms of the individual blocks.
+  double l1_norm = -INFINITY;
```
akleeman, line 237: If blocks is empty what should we return? Probably NAN instead of -INFINITY (imagine someone checking `rcond() < 10000`, we'd probably want that to return false?)

peddie: That's a great point.

include/albatross/src/linalg/block_diagonal.hpp
peddie: I can test it, but I don't think forcing an expression type is enough here.

include/albatross/src/eigen/serializable_ldlt.hpp
peddie: That's a good point, I'll figure something else out.

peddie:
Thanks for the careful review.  I wrote this code in mad prototyping mode and failed to go back through some of these small patches in detail.  I'll rework this.

include/albatross/src/linalg/block_diagonal.hpp
peddie: `MatrixBase<Derived>` doesn't cover `CWiseBinaryOp` for sure.  But I think it could cover `Block`.

peddie: There are a couple problems here.  One is that we have to declare a type for `output`.  So if we use `MatrixBase<Derived>`, then if you pass it a `CWiseBinaryOp`, it tries to declare one and assign to it.  This is clearly no good as `CWiseBinaryOp`s are never lvalues.  But how else can we determine the return type so as to preserve sparsity?  We don't care about that for `rcond()`, but we do care about it (I think) in PIC.

peddie: The Right Thing to do here is to make `BlockDiagonalLDLT` cooperate with the Eigen expression system.

I think I'm going to just make it return `MatrixXd` for now.  But also I was wrong above that `MatrixBase` covers `Block`.

include/albatross/src/eigen/serializable_ldlt.hpp
peddie: Copy removed with no apparent issue.

include/albatross/src/linalg/block_diagonal.hpp
peddie: Added a check for this case.

include/albatross/src/linalg/block_diagonal.hpp
peddie: This change is up, which allows removing only the overload for `CWiseBinaryOp`.

peddie:
This should be good to go if you are happy with the rewrite.

include/albatross/src/eigen/serializable_ldlt.hpp
akleeman: Are we sure that `Eigen::MatrixXd(rhs)` isn't copying?

include/albatross/src/linalg/block_diagonal.hpp
```diff
@@ -74,12 +93,12 @@ struct BlockDiagonal {
 /*
  * BlockDiagonalLDLT
  */
-template <class _Scalar, int _Rows, int _Cols>
-inline Eigen::Matrix<_Scalar, _Rows, _Cols> BlockDiagonalLDLT::solve(
-    const Eigen::Matrix<_Scalar, _Rows, _Cols> &rhs) const {
+template <typename Derived>
+inline Eigen::MatrixXd
```
akleeman, line 97: There must be a way to do this via Eigen types such that if the input is static sized the output will be too? `Eigen::Matrix<_Scalar, _Rows, _Cols>` would have allowed for that, but `Eigen::MatrixXd` will force all output to be dynamic doubles. I don't think it's a problem for us, and feel free to proceed without changing this, it just feels like there should be a way, `Eigen::DenseBase<Derived>::PlainMatrix` or something like that?

akleeman (approved):

include/albatross/src/linalg/block_diagonal.hpp
peddie: I think the problem is not all expression types involved here derive from even `DenseBase`.

akleeman: `MatrixBase` inherits from `DenseBase` so it should right? https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/MatrixBase.h#L52

include/albatross/src/eigen/serializable_ldlt.hpp
peddie: Oh, I'm not sure why this is still in the change.  I thought I removed it.  I'll make sure it's gone.

include/albatross/src/linalg/block_diagonal.hpp
peddie: Yes, but this gets used with things other than a `MatrixBase`.

akleeman: I guess I meant that if the input argument is a `MatrixBase<Derived>` then maybe you could get the `PlainMatrix` from it?

I doubt this would work but if `Eigen::MatrixBase<Derived>` works, then `Eigen::DenseBase<Derived>` should right?
```
template <typename Derived>
Eigen::DenseBase<Derived>::PlainMatrix BlockDiagonalLDLT::solve(const Eigen::MatrixBase<Derived> &rhs) const {
```

Anyway everytime I dig into Eigen I start to think I understand how to extend it, then I find something like `Solve.h` and get thoroughly lost. This isn't important enough to spend the time figuring it out.

peddie: Sorry . . . I think the long-term "right way" is to make our `solve()` members look just like Eigen's ones and only overload `_solve_impl()` or whatever it is that takes an `RhsType` reference.

PR #496, opened 2024-10-30
Remove additional semicolon

armallen:
`DEFINE_GET_PARAMS`, `DEFINE_SET_PARAM` and `DECLARE_PARAMS` declare `;` by themselves, calling `ALBATROSS_DECLARE_PARAMS` leads to `;;` being declared, which is not allowed by compilers with strict warnings.

Tested in https://github.com/swift-nav/orion-engine/pull/8882


peddie (approved):
nit: semi _colon_ :)

PR #498, opened 2024-11-13
Tutorial

lkloh:


examples/temperature_example/temperature_example_utils.h
```diff
@@ -42,6 +42,30 @@ struct Station {
   }
 };
 
+/*
+ * Holds the information about a single station which
+ * is used as the FeatureType for our Gaussian process.
+ */
+struct ReducedStation {
+  double lat;
+  double height;
```
lkloh, line 51: Use only `lat` and `height` for prediction (what if we only had access to these variables?)

PR #501, opened 2025-01-24
Parallel calculation support in groupwise cross validation

peddie:
Previously the clever GP cross-validation calculations could leverage BLAS / LAPACK parallelism, but iteration across held-out groups happened in a serial for-loop.  This could lead to suboptimal performance in situations with many held-out groups to evaluate.

This commit plumbs thread pool support into the cross-validation and inverse covariance block calculation routines and uses the model's internal thread pool by default.

### DIFF
--- a/include/albatross/Distribution
+++ b/include/albatross/Distribution
@@ -16,6 +16,9 @@
 #include "Common"
 
 #include <albatross/src/core/distribution.hpp>
+
+#include "utils/AsyncUtils"
+
 #include <albatross/src/eigen/serializable_ldlt.hpp>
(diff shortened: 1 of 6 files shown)


peddie:
Note the age of this patch.  I have used it locally and in EC2 many times in the past year, but I never got a review for it.

What is the interface change you're referring to?  If it is the new parallel behaviour, I think it's OK.  Other operations in Albatross are automatically parallelised if you give the model a thread pool with multiple workers.  If there is some change in type signatures I missed, please let me know.  It clearly hasn't broken the test suite.

PR #502, opened 2025-03-18
Add getter and setter for the GrouperFunction param of SparseGaussianProcessRegression class [OC-2274]

lkloh:
Related PR: https://github.com/swift-nav/orion-engine/pull/9278

### DIFF
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -614,8 +614,8 @@ public:
     return independent_group_function_;
   }
 
-  GrouperFunction &get_grouper_function() {
-    return independent_group_function_;
+  void set_grouper_function(GrouperFunction &&independent_grouper_function) {
+    independent_group_function_ = independent_grouper_function;
   }
 
 private:
--- a/include/albatross/src/models/sparse_gp.hpp
+++ b/include/albatross/src/models/sparse_gp.hpp
@@ -610,6 +610,14 @@ public:
     return inducing_point_strategy_;
   }
 
+  GrouperFunction get_grouper_function() const {
+    return independent_group_function_;
+  }
+
+  GrouperFunction &get_grouper_function() {
+    return independent_group_function_;
+  }
+
 private:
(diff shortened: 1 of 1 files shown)


include/albatross/src/models/sparse_gp.hpp
```diff
@@ -610,6 +610,14 @@ class SparseGaussianProcessRegression
     return inducing_point_strategy_;
   }
 
+  GrouperFunction get_grouper_function() const {
+    return independent_group_function_;
+  }
+
+  GrouperFunction &get_grouper_function() {
+    return independent_group_function_;
```
lkloh, line 618: I can't exactly add a `setter` since `GrouperFunction` is a template type with no information, but we can enable the caller to modify it directly

peddie: Why does `void set_grouper_function(GrouperFunction &&) { ... }` not work?  What is the problem with using the template parameter as an argument type rather than a return type?

lkloh: I guess that should be ok actually

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -610,6 +610,14 @@ class SparseGaussianProcessRegression
     return inducing_point_strategy_;
   }
 
+  GrouperFunction get_grouper_function() const {
+    return independent_group_function_;
+  }
+
+  void set_grouper_function(GrouperFunction &&independent_grouper_function) {
+    independent_group_function_ = independent_grouper_function;
```
peddie, line 618: In this context, where `GrouperFunction` is a template type, `independent_grouper_function` is a [forwarding reference](https://lemire.me/blog/2024/05/13/forwarding-references-in-c/), so you can get the compiler to automatically do the right thing (move or copy) for the context by doing

    independent_group_function_ = std::forward<GrouperFunction>(independent_grouper_function);

peddie (commented):
Could you please also add a setter for the inducing point strategy while you're at it?

peddie (approved):
Perfect, thanks!

PR #503, opened 2025-03-19
fix build issue

lkloh:
Fix build issue, see https://github.com/swift-nav/albatross/pull/504 where before it failing

PR #506, opened 2025-03-21
Dependabot [OC-2291]

lkloh:
Recently had to deal with https://github.com/swift-nav/albatross/pull/503, hopefully dependabot will help


pcrumley (approved):
Sometimes it can be nice to group dependabot changes into one pr to avoid a zillion commits but LGTM as is

PR #515, opened 2025-06-04
Upgrade cereal dependency[OC-2654]

john-michaelburke:


peddie (approved):
Great to see these ancient deps getting a fresh look.

jbangelo:
How confident are we in their adherence to SemVer? Are we concerned if the serialization format isn't backwards compatible?

john-michaelburke:
> How confident are we in their adherence to SemVer? Are we concerned if the serialization format isn't backwards compatible?

Like I said, this pass through CI all the way to the top, bit of a convoluted PR but it includes this commit <ill send in a DM>

ljbade (approved):

PR #524, opened 2025-08-29
Cereal thread safety (cmake)

RReichert:


peddie:
I don't have any problem with modifying the build this way, but 1) do we need a corresponding Bazel change, 2) was this causing a problem in Albatross, or you are just locking it down everywhere to prevent future issues?

cmake/FindCereal.cmake
```diff
@@ -6,6 +6,7 @@ endif()
 
 if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/third_party/cereal/include")
   add_library(cereal INTERFACE)
-
   target_include_directories(cereal SYSTEM INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/third_party/cereal/include")
+  target_compile_definitions(cereal INTERFACE CEREAL_THREAD_SAFE=1)
```
RReichert, line 10: this forces all the libraries that use it to use the thread safe mode, which is better

RReichert:
> I don't have any problem with modifying the build this way, but was this causing a problem in Albatross, or you are just locking it down everywhere to prevent future issues?

This wasn't causing problems with albatross, this was causing problems downstream with code that used Cereal. We have multithreading application that use this and we noticed race conditions which caused Cereal to lock itself up in an infinite loop.

ljbade (approved):

peddie (approved):

PR #528, opened 2025-09-09
Update rules_swiftnav

martin4861:
# Description

- Update `rules_swiftnav` to downstream version.
- Move compile commands in correct subfolder.
- Use newer version of `cereal`, same as downstream.


peddie:
If you rebase on current `master`, your OS X build will probably start working . . . I just merged #532

martin4861:
> I don't see anything obviously wrong here, but I don't know much about `rules_swiftnav`, so you may want to ask someone more familiar to take a look as well.

Thanks for the review. I am quite confident that this is fine, because I made similar changes to other repos yesterday.

PR #530, opened 2025-09-10
Split test target

peddie:
Previously the unit test suite ran as a single Bazel target.  Since albatross is a header-only library, this often caused small edits to cause huge rebuilds during the development cycle (~every test module would be rebuild).  This commit splits the test target into multiple topical targets, enabling hacking with considerably shorter recompilation cycles.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -185,26 +185,31 @@ swift_cc_test_library(
     deps = ["@gtest"],
 )
 
+swift_cc_test_library(
+    name = "albatross-test-common",
+    srcs = [
+        "tests/mock_model.h",
+        "tests/test_utils.h",
+        "tests/test_models.h",
+        "tests/test_core_distribution.h",
+        "tests/test_covariance_utils.h",
+    ],
+    includes = ["tests"],
+)
+
 # Common test dependencies
 TEST_DEPS = [
     ":albatross",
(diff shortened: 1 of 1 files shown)


BUILD.bazel
```diff
@@ -185,77 +185,235 @@ swift_cc_test_library(
...
+# Common test dependencies
+TEST_DEPS = [
+    ":albatross",
+    ":serialize-testsuite",
+    "@zlib",
+    "@gtest//:gtest_main",
+    "@suitesparse//:spqr",
+]
+
+TEST_COMMON = [
```
martin4861, line 197: Can you extract a library out of that, e.g. `swift_cc_test_library`?

martin4861 (commented):
Thanks for splitting up the tests (which will increase the bazel cache hit rate) 🚀

BUILD.bazel
martin4861: I think it would not really change anything, except that the header files are modeled properly. But I think, it's also fine this way.

peddie: I am happy to fix it if it's meaningful to have a lib of just headers.  Please have a look at the latest version.

lkloh (approved):
thanks

jbangelo (approved):

PR #562, opened 2026-01-22
Support defining covariance functions on batches

peddie:
Previously the covariance machinery worked point-by-point -- users define covariance between scalar values using `_call_impl()`, and Albatross computes the required matrices.  In theory, this is typically correct, but in some cases, there may be a requirement to compute on batches of training data at once (for example, to perform an FFT and improve asymptotic scaling).

This commit adds an additional system to the covariance CRTP machinery, so that if a user defines `_call_impl_vector()`, which accepts `std::vector`s of feature types, this batch definition will be used instead, enabling things such as the FFT use case.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -229,6 +229,7 @@ swift_cc_test(
 swift_cc_test(
     name = "albatross-test-covariance",
     srcs = [
+        "tests/test_batch_covariance.cc",
         "tests/test_covariance_function.cc",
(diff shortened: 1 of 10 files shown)


peddie:
Ping @akleeman -- if you are interested in this kind of core change / have strong objections, otherwise please ignore.

jbangelo (approved):
:shipit:

PR #573, opened 2026-03-30
Migration to clang20 sanitizers with Bazel

martin4861:
Update sanitizers from clang14 to clang20 and from CMake to Bazel.
In order to pass the suppression file to `cc_test` we updated to Bazel 9.1.0 and used `additional_compiler_inputs`. To not affect any downstream targets we moved all tests into a separate `tools/BUILD.bazel` file and created a new folder structure for the shared test lib (which is consumed downstream).

Fixes:
- [x] Implement suppression for Eigen
- [x] Update cereal to latest version to fix `nullptr` arithmetic in `rapidjson`.


peddie (approved):
I can't say I fully understand all the changes to the build / CI here, but I am happy with the high-level change.

PR #575, opened 2026-04-17
Bazel 9 prepa

martin4861:
- Update dependencies to support Bazel 9

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,11 +1,10 @@
-# Enable bzl_mod support
 # In order to properly use the swiftnav-bazel-registry your local netrc needs to be configured
-common --enable_bzlmod
 common --registry=https://raw.githubusercontent.com/swift-nav/swiftnav-bazel-registry/refs/heads/main/
 common --registry=https://bcr.bazel.build
-common --@rules_swiftnav//cc:enable_bzlmod=true
-# Preparation for Bazel 8 behavior of paths in external (plus instead of tilde)
-common --incompatible_use_plus_in_repo_names
+
+# Compile all swift targets and all other codes with C++17 standards
+build --@rules_swiftnav//cc:cxx_standard=17
+build --@rules_swiftnav//cc:global_cxx_standard=17
 
 build --enable_platform_specific_config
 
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -1,11 +1,10 @@
-# Enable bzl_mod support
 # In order to properly use the swiftnav-bazel-registry your local netrc needs to be configured
-common --enable_bzlmod
 common --registry=https://raw.githubusercontent.com/swift-nav/swiftnav-bazel-registry/refs/heads/main/
 common --registry=https://bcr.bazel.build
-common --@rules_swiftnav//cc:enable_bzlmod=true
-# Preparation for Bazel 8 behavior of paths in external (plus instead of tilde)
-common --incompatible_use_plus_in_repo_names
+
+# Compile all swift targets and all other codes with C++14 standards
```
peddie: I thought we were on 17 everywhere?

tests/test_serializable_ldlt.cc
```diff
@@ -42,7 +42,7 @@ TEST_F(SerializableLDLTTest, test_inverse_diagonal) {
   const auto serializable_ldlt = Eigen::SerializableLDLT(ldlt);
   const auto inverse = cov.inverse();
   const auto diag = serializable_ldlt.inverse_diagonal();
-  EXPECT_LE(fabs((Eigen::VectorXd(inverse.diagonal()) - diag).norm()), 1e-8);
+  EXPECT_LE(fabs((Eigen::VectorXd(inverse.diagonal()) - diag).norm()), 1e-6);
```
peddie: I can't decide how unhappy this makes me.  1e-8 shouldn't be a high bar for whatever happens on Mac OS under the hood with Eigen (machine precision is better than 1e-15) and a 5x5 matrix.  It looks like it is chosen randomly, and we don't control the seed, so perhaps the fix is to explicitly seed it and ensure we test on a reasonable example every time?

peddie (changes requested):

tests/test_serializable_ldlt.cc
martin4861: you're right. The tests are anyway broken on master for Mac. Therefore I keep the current test implementation and they need to be fixed seperately

peddie (approved):

PR #583, opened 2026-06-25
Computational perf fixes: Polynomial, LDLT/block, indexing, inverse cholesky

armallen:
Computational performance fixes for the GP/covariance hot paths, validated with a new
Google Benchmark suite. Each commit is one self-contained, independently measured change.

> Note: this branch was cleaned up after benchmarking — three changes that measured as
> no-ops or regressions (an SE squared-distance fast path, a variogram allocation skip, and
> a ScalingTerm matrix-level fast path) were dropped. Only measured wins remain.

### 1. Add Google Benchmark target
Adds `//benchmarks:albatross-bench` covering covariance builds, `SerializableLDLT` ops, GP
fit/predict, and CV. This is the measurement harness for everything below — no production code.
Run with `bazel run -c opt //benchmarks:albatross-bench`.

### 2. Optimize Polynomial covariance function evaluation
`Polynomial` did two `std::map` lookups (one copying the whole param store) plus a `pow()` per
`(i,j)`. Cache `sigma^2` in an array and use running products instead. **~87× faster** at n=2048.

### 3. Optimize LDLT, BlockSymmetric, BlockDiagonal hot paths
`matrix()` now returns `const&` instead of deep-copying the factor each call; `BlockSymmetric`
stops solving the same system twice; block products accumulate in-place with `.noalias()`.
**~1.2–2.2× faster** on the affected ops.

### 4. Optimize indexing, folds, and joint LOO
`indices_complement` swaps a `std::set` set-difference for a `vector<bool>` mask + one linear
pass (**~100–230× faster**, dominant cost in K-fold setup); folds read the sorted set's ends
directly; joint LOO uses an SPD `LDLT` solve instead of `MatrixXd::inverse()`.

### 5. Relax inverse_diagonal test tolerance to 1e-6
Test-only. The `colwise().squaredNorm()` reformulation in #3 has slightly different roundoff
(~1.6e-7) than the old path; the old `1e-8` bound was too tight. Keeps a 10× margin.

### 6. Skip permuted-identity build in inverse_diagonal
Instead of materializing `(P·I)` as the solve RHS, solve against plain `I` and permute the
resulting n-vector at the end — drops an n² permutation pass. **~1.25–1.44× faster**.

### 7. Skip permuted-identity build in inverse_blocks
Same trick for `inverse_blocks`: solve against `I` and translate block column indices through
the permutation when subsetting, never materializing the full inverse cholesky. **~1.17–1.30× faster**.

---
All numbers are medians from `//benchmarks` (opt build, 5 reps). One pre-existing test failure
(`VariogramScorePropertiesTest.ProperScoringMeanPerturbation`) is unrelated to this branch — it
fails identically on `master`.


benchmarks/bench_kernels.cc
```diff
@@ -0,0 +1,140 @@
...
+  const auto xs = make_vector_features(n, 8);
+  albatross::Matern52<albatross::EuclideanDistance> cov;
+  for (auto _ : state) {
+    Eigen::MatrixXd K = cov(xs);
+    benchmark::DoNotOptimize(K);
+  }
+  state.SetItemsProcessed(state.iterations() * static_cast<int64_t>(n) *
+                          static_cast<int64_t>(n));
+}
+BENCHMARK(BM_Matern52_VectorXd_Symmetric)->Arg(256)->Arg(1024)->Arg(2048);
```
peddie, line 109: I don't understand what these kernel benchmarks are testing.  There don't appear to be any changes to `radial.hpp` that they would measure.

Note that because everything is headers, there is a pretty good chance the optimiser sees `pow(sqrt(x), 2.0)` and can do the right thing.

Could you take a look at https://github.com/swift-nav/albatross/compare/master...peddie/block-covariance#diff-b35466c157fbfa3fd3fc3941264991bc9ab08ba617ab32b6fb4867935989f8a6 which provides more flexible ways to compute these basic radial functions on vectors at a time?

peddie (commented):
Most of this looks fine -- building the permutation into subset indexing is a good idea.  I'm not sure about the intent of the benchmarks on standard radial kernel terms; can you look at the branch I linked and see whether there is some overlap there with what you're thinking?

armallen:
> Most of this looks fine -- building the permutation into subset indexing is a good idea. I'm not sure about the intent of the benchmarks on standard radial kernel terms; can you look at the branch I linked and see whether there is some overlap there with what you're thinking?

@peddie  Sorry I created this PR only to make sure to remember this work, it's not ready for review as I have not reviewed it myself 😬

PR #590, opened 2026-07-28
Fix five bugs found during code analysis

armallen:
One commit per bug, each with a regression test (behavioral ones verified failing pre-fix):

- SPQR pivot-threshold fall-through (small-matrix branch was dead code)
- Duplicate O(n^3) LDLT factorizations in LOGO-CV and `fit_from_prediction`
- `fit_from_prediction` discarded the mean-removed prediction (wrong results with non-zero mean functions)
- Data race: `std::map::operator[]` from thread-pool workers in `held_out_predictions`
- `SparseQRSparseGaussianProcessRegression` alias mis-bound template params; `get_predictions` returned in-sample predictions

### DIFF
--- a/include/albatross/src/covariance_functions/representations.hpp
+++ b/include/albatross/src/covariance_functions/representations.hpp
@@ -70,6 +70,12 @@ struct ExplainedCovariance {
     inner = inner_;
   }
 
+  // Use this constructor if the LDLT of the outer matrix has already
+  // been computed; it avoids re-factorizing the outer matrix.
+  ExplainedCovariance(const Eigen::SerializableLDLT &outer_ldlt_,
+                      const Eigen::MatrixXd &inner_)
+      : outer_ldlt(outer_ldlt_), inner(inner_) {}
+
   /*
    * Returns S^-1 rhs by using S^-1 = A^-1 B A^-1
    */
--- a/include/albatross/src/evaluation/cross_validation_utils.hpp
+++ b/include/albatross/src/evaluation/cross_validation_utils.hpp
@@ -21,7 +21,7 @@ get_predictions(const ModelType &model,
                 const RegressionFolds<GroupKey, FeatureType> &folds) {
 
   const auto predict_group = [&model](const auto &fold) {
-    return model.fit(fold.train_dataset).predict(fold.train_dataset.features);
+    return model.fit(fold.train_dataset).predict(fold.test_dataset.features);
   };
 
   return folds.apply(predict_group);
--- a/include/albatross/src/evaluation/cross_validation_utils.hpp
+++ b/include/albatross/src/evaluation/cross_validation_utils.hpp
@@ -218,8 +218,11 @@ held_out_predictions(const Eigen::SerializableLDLT &covariance,
   return apply(
              group_indexer,
              [&](const auto &key, const auto &indices) {
+               // `blocks` is shared across the pool's workers; use the
+               // const `at()` accessor since the non-const `operator[]`
+               // is a data race.
                return held_out_prediction(
-                   blocks[key], subset(target_mean, indices),
+                   blocks.at(key), subset(target_mean, indices),
                    subset(information, indices), predict_type);
              },
              pool)
--- a/include/albatross/src/evaluation/cross_validation_utils.hpp
+++ b/include/albatross/src/evaluation/cross_validation_utils.hpp
@@ -238,8 +238,11 @@ leave_one_group_out_conditional(const JointDistribution &prior,
   Eigen::SerializableLDLT ldlt(covariance);
   const Eigen::VectorXd deviation = truth.mean - prior.mean;
   const Eigen::VectorXd information = ldlt.solve(deviation);
-  return held_out_predictions(covariance, truth.mean, information,
-                              group_indexer, predict_type, pool);
+  // Pass the already computed LDLT here; passing the dense covariance
+  // would silently re-factorize it (an O(n^3) operation) through
+  // Eigen::SerializableLDLT's implicit converting constructor.
+  return held_out_predictions(ldlt, truth.mean, information, group_indexer,
+                              predict_type, pool);
 }
 
 } // namespace details
--- a/include/albatross/src/linalg/spqr_utils.hpp
+++ b/include/albatross/src/linalg/spqr_utils.hpp
@@ -43,8 +43,8 @@ constexpr Eigen::Index kMinSparsePivotSize = 300;
 // SPQR pivot threshold policy.
 constexpr double kSPQRPivotCoefficient = 0.1;
 
-// Set the pivot threshold of the solver based on the problem to be
-// solved.
+// Compute the pivot threshold of the solver based on the problem to
+// be solved.
 //
 // The default SPQR pivot threshold calculation is oriented towards
 // problems considerably larger than your typical Albatross use case.
@@ -52,13 +52,13 @@ constexpr double kSPQRPivotCoefficient = 0.1;
 // function to choose a smaller threshold than the default and trade a
 // minimal speed reduction for accuracy improvements.
 template <typename MatrixType>
-inline void SPQR_set_pivot_threshold(SPQR *spqr, const MatrixType &m,
-                                     Eigen::Index min_sparse_pivot_size,
-                                     double spqr_pivot_coefficient) {
+inline double SPQR_pivot_threshold(const MatrixType &m,
+                                   Eigen::Index min_sparse_pivot_size,
+                                   double spqr_pivot_coefficient) {
   if (m.rows() < min_sparse_pivot_size || m.cols() < min_sparse_pivot_size) {
     // For small matrices, just use a really tight threshold for pivot
     // detection to reduce the inaccuracy of the sparse method.
-    spqr->setPivotThreshold(1e-15);
+    return 1e-15;
   }
   // For larger matrices, use the default SuiteSparseQR policy to
   // choose the pivot threshold, but accept a user-controlled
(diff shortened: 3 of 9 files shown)


include/albatross/src/linalg/spqr_utils.hpp
```diff
@@ -43,22 +43,22 @@ constexpr Eigen::Index kMinSparsePivotSize = 300;
...
-                                     Eigen::Index min_sparse_pivot_size,
-                                     double spqr_pivot_coefficient) {
+inline double SPQR_pivot_threshold(const MatrixType &m,
+                                   Eigen::Index min_sparse_pivot_size,
+                                   double spqr_pivot_coefficient) {
   if (m.rows() < min_sparse_pivot_size || m.cols() < min_sparse_pivot_size) {
     // For small matrices, just use a really tight threshold for pivot
     // detection to reduce the inaccuracy of the sparse method.
-    spqr->setPivotThreshold(1e-15);
+    return 1e-15;
```
armallen, line 61: **Bug:** the small-matrix branch previously called `setPivotThreshold(1e-15)` but was missing a `return`, so the default policy below unconditionally overwrote it. The documented "small problem" threshold was dead code, and small matrices got a threshold roughly an order of magnitude looser than intended, changing which fits were treated as rank-deficient. The calculation is extracted into this pure function because Eigen's SPQR has no `pivotThreshold()` getter, so the behavior wasn't testable through the setter.

include/albatross/src/models/gp.hpp
```diff
@@ -143,9 +143,16 @@ gp_fit_from_prediction(const std::vector<FeatureType> &features,
                        const JointDistribution &prediction) {
   Fit<GPFit<ExplainedCovariance, FeatureType>> fit;
   fit.train_features = features;
+  // Factorize the prior once and share the decomposition between the
+  // covariance representation and the information vector.  Passing the
+  // dense `prior` to ExplainedCovariance (or calling `prior.ldlt()`
+  // again below) would trigger a second O(n^3) factorization, possibly
+  // silently through Eigen::SerializableLDLT's implicit converting
+  // constructor.
+  const Eigen::SerializableLDLT prior_ldlt(prior);
```
armallen, line 148: **Bug (performance):** `ExplainedCovariance(prior, ...)` factorizes `prior` internally, and the old next line called `prior.ldlt()` again — two full O(n^3) factorizations of the same matrix per `fit_from_prediction` call. This path runs per satellite in orion-engine's per-satellite ionosphere model.

include/albatross/src/models/gp.hpp
```diff
@@ -238,7 +245,7 @@ class GaussianProcessBase : public ModelBase<ImplType> {
     using FitType = Fit<GPFit<ExplainedCovariance, FeatureType>>;
     return FitModel<ImplType, FitType>(
         impl(), gp_fit_from_prediction(features, covariance_function_(features),
-                                       prediction));
+                                       zero_mean_prediction));
```
armallen, line 244: **Bug:** `zero_mean_prediction` was constructed and mean-removed a few lines up, but the original `prediction` was passed here, discarding the correction. The stored information vector then included the mean function's contribution, which `predict()` adds again — double-counting the mean. Invisible with the default `ZeroMean`; the new `test_model_from_prediction_with_mean` failed with a mean error of 16.45 before this fix.

include/albatross/src/evaluation/cross_validation_utils.hpp
```diff
@@ -21,7 +21,7 @@ get_predictions(const ModelType &model,
                 const RegressionFolds<GroupKey, FeatureType> &folds) {
 
   const auto predict_group = [&model](const auto &fold) {
-    return model.fit(fold.train_dataset).predict(fold.train_dataset.features);
+    return model.fit(fold.train_dataset).predict(fold.test_dataset.features);
```
armallen, line 24: **Bug:** this predicted `fold.train_dataset.features` — in-sample predictions masquerading as cross-validation. No in-tree callers (which is why it went unnoticed); fixed to mirror `predict_fold` in cross_validation.hpp.

include/albatross/src/evaluation/cross_validation_utils.hpp
```diff
@@ -218,8 +218,11 @@ held_out_predictions(const Eigen::SerializableLDLT &covariance,
   return apply(
              group_indexer,
              [&](const auto &key, const auto &indices) {
+               // `blocks` is shared across the pool's workers; use the
+               // const `at()` accessor since the non-const `operator[]`
+               // is a data race.
                return held_out_prediction(
-                   blocks[key], subset(target_mean, indices),
+                   blocks.at(key), subset(target_mean, indices),
```
armallen, line 225: **Bug:** `blocks[key]` (non-const `std::map::operator[]`) was invoked concurrently from thread-pool workers — a data race per [res.on.data.races], flagged by TSan. Benign in practice because every key already exists (no insertion/rehash), but still UB. `at()` is const and race-free here.

include/albatross/src/evaluation/cross_validation_utils.hpp
```diff
@@ -238,8 +241,11 @@ leave_one_group_out_conditional(const JointDistribution &prior,
   Eigen::SerializableLDLT ldlt(covariance);
   const Eigen::VectorXd deviation = truth.mean - prior.mean;
   const Eigen::VectorXd information = ldlt.solve(deviation);
-  return held_out_predictions(covariance, truth.mean, information,
-                              group_indexer, predict_type, pool);
+  // Pass the already computed LDLT here; passing the dense covariance
+  // would silently re-factorize it (an O(n^3) operation) through
+  // Eigen::SerializableLDLT's implicit converting constructor.
+  return held_out_predictions(ldlt, truth.mean, information, group_indexer,
```
armallen, line 247: **Bug (performance):** the dense `covariance` was passed here even though its LDLT was just computed above. `held_out_predictions` takes `const Eigen::SerializableLDLT &` and `SerializableLDLT(const MatrixXd&)` is a non-explicit converting constructor, so the same matrix was silently factorized a second time — an extra O(n^3) per call (runs per station in orion-engine's detailed evaluation).

include/albatross/src/models/sparse_gp.hpp
```diff
@@ -727,7 +727,7 @@ auto rebase_inducing_points(
 template <typename CovFunc, typename MeanFunc, typename GrouperFunction,
           typename InducingPointStrategy>
 using SparseQRSparseGaussianProcessRegression =
-    SparseGaussianProcessRegression<CovFunc, GrouperFunction,
+    SparseGaussianProcessRegression<CovFunc, MeanFunc, GrouperFunction,
```
armallen, line 730: **Bug:** the alias passed 4 arguments into the 5-parameter primary template: `GrouperFunction` bound to the `MeanFunc` slot, `InducingPointStrategy` to `GrouperFunction`, `SPQRImplementation` to `InducingPointStrategy`, and `QRImplementation` fell back to its `DenseQRImplementation` default — so the "SparseQR" alias would not have used SPQR at all. It compiled only because nothing instantiated it; the new static_assert test failed to compile before the fix.

include/albatross/src/covariance_functions/representations.hpp
```diff
@@ -70,6 +70,12 @@ struct ExplainedCovariance {
     inner = inner_;
   }
 
+  // Use this constructor if the LDLT of the outer matrix has already
+  // been computed; it avoids re-factorizing the outer matrix.
+  ExplainedCovariance(const Eigen::SerializableLDLT &outer_ldlt_,
```
armallen, line 75: Support for the duplicate-factorization fix in `gp_fit_from_prediction`: lets a caller that already holds the LDLT of the outer matrix construct the representation without re-factorizing.

armallen (commented):
Inline notes explaining each bug at its fix location.

tests/test_linalg_utils.cc
```diff
@@ -31,6 +31,29 @@ TEST(test_linalg_utils, test_qr_sqrt_solve) {
...
 }
 
+TEST(test_linalg_utils, test_spqr_pivot_threshold_small_matrix) {
+  // Regression test: matrices smaller than the minimum sparse pivot
+  // size must use the tight small-problem threshold (1e-15) instead of
+  // falling through to the large-problem SuiteSparseQR policy.
+  const Eigen::Index small_size = kMinSparsePivotSize - 1;
+  const Eigen::MatrixXd small =
+      Eigen::MatrixXd::Identity(small_size, small_size);
+  EXPECT_EQ(
```
michaudouble: EXPECT_DOUBLE_EQ

include/albatross/src/models/gp.hpp
```diff
@@ -143,9 +143,16 @@ gp_fit_from_prediction(const std::vector<FeatureType> &features,
                        const JointDistribution &prediction) {
   Fit<GPFit<ExplainedCovariance, FeatureType>> fit;
   fit.train_features = features;
+  // Factorize the prior once and share the decomposition between the
```
michaudouble, line 146: We could only keep the first sentence of this comment. Don't you think it's already sufficient?

tests/test_cross_validation.cc
```diff
@@ -301,6 +301,47 @@ TEST(test_crossvalidation, test_leave_one_group_out_conditional_reference) {
...
+      prior, dataset.targets, indexers, pool.get());
+
+  for (const auto &pair : indexers) {
+    const auto &key = pair.first;
+    EXPECT_EQ(threaded_means.at(key), serial_means.at(key));
+    EXPECT_EQ(threaded_marginals.at(key).mean, serial_marginals.at(key).mean);
+    EXPECT_EQ(Eigen::MatrixXd(threaded_marginals.at(key).covariance),
+              Eigen::MatrixXd(serial_marginals.at(key).covariance));
+    EXPECT_EQ(threaded_joints.at(key).mean, serial_joints.at(key).mean);
+    EXPECT_EQ(threaded_joints.at(key).covariance,
```
michaudouble: EXPECT_DOUBLE_EQ instead of EXPECT_EQ

tests/test_sparse_gp.cc
```diff
@@ -267,6 +267,31 @@ TEST(test_sparse_gp_sparse_qr, test_update_exists) {
   EXPECT_TRUE(bool(can_update_in_place<SparseGPR, FitType, double>::value));
 }
 
+TEST(test_sparse_gp_sparse_qr, test_alias_forwards_all_template_params) {
+  // Compile-time regression test: the SparseQRSparseGaussianProcessRegression
+  // alias must forward all four of its template parameters into the
+  // corresponding slots of SparseGaussianProcessRegression and select
+  // SPQRImplementation.  A previous version dropped MeanFunc, shifting
+  // every parameter into the wrong slot and silently falling back to
+  // DenseQRImplementation.
```
michaudouble: Is it useful to mention "A previous version dropped...". I would remove that part of the comment.

tests/test_linalg_utils.cc
armallen: Done in 011e4fa.

include/albatross/src/models/gp.hpp
armallen: Done in 011e4fa.

tests/test_cross_validation.cc
armallen: Done in 011e4fa.

tests/test_sparse_gp.cc
armallen: Done in 011e4fa.

michaudouble (approved):
Nice!

PR #592, opened 2026-07-28
Avoid lazy Eigen re-evaluation and gratuitous copies on hot paths

armallen:
BlockSymmetric: 4 -> 2 Schur solves, compute A^-1 B once. Materialize lazy `Inverse<>` in held-out prediction. `get_fit()` by reference, movable LDLT, copy-free `sqrt_solve`.

## Benchmark results

| benchmark | before | after | speedup |
|---|---|---|---|
| block_symmetric_construct/256 | 329 µs | 191 µs | **1.72x** |
| block_symmetric_construct/512 | 2283 µs | 1337 µs | **1.71x** |
| block_symmetric_solve_vector/512 | 147 µs | 96 µs | **1.53x** |
| block_symmetric_solve_matrix/512 | 732 µs | 493 µs | **1.48x** |
| sparse_gp_predict_joint/1024 | 9.63 ms | 8.79 ms | **1.10x** |
| loo-cv / fit / predict benches | — | — | within noise |

Stack 2/4, on #591.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #593, opened 2026-07-28
Compute inverse_diagonal directly from the inverse Cholesky factor

armallen:
Replaces the n-singleton-block machinery with column squared norms; removes per-element allocations and fixes the incorrect O(n^2) complexity comment. New regression test vs `inverse_blocks`.

## Benchmark results

| benchmark | speedup |
|---|---|
| inverse_diagonal/256 | **1.05x** |
| inverse_diagonal/512 | **1.02x** |
| inverse_diagonal/1024 | **~1.01x** |

Honest assessment: the O(n^3) triangular solve dominates, so the win is small — kept because the code is simpler, allocation-free, and never slower.

Stack 3/5, on #592.

### DIFF
--- a/include/albatross/src/eigen/serializable_ldlt.hpp
+++ b/include/albatross/src/eigen/serializable_ldlt.hpp
@@ -175,27 +175,27 @@ public:
   }
 
   /*
-   * The diagonal of the inverse of the matrix this LDLT
-   * decomposition represents in O(n^2) operations.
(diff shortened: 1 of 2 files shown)


armallen:
Parking this change for now — decision: not keeping the inverse_diagonal rewrite in the code at this time. The branch remains available if we revisit. Removed from the stack: #594 now targets perf/eigen-eval-and-copies directly.

PR #591, opened 2026-07-28
Add Google Benchmark micro-benchmark infrastructure

armallen:
Adds google_benchmark 1.9.5 (dev dep, BCR) and a `benchmarks/` package covering BlockSymmetric, LOO-CV, prediction (with/without thread pool), and Gram products. Baseline for the stacked perf PRs.

PR #596, opened 2026-07-29
Add opt-in matrix-level kernel evaluation API

armallen:
Covariance assembly currently fills K one scalar `k(x, y)` call at a time — a loop that never becomes a BLAS call, and the dominant cost of GP mean prediction. 

This adds an opt-in whole-matrix kernel path while leaving the per-pair path bit-for-bit unchanged as the fallback.

## Implementation

- **Trait + dispatch:** `has_valid_matrix_caller<CovFunc, X, Y>` detects an optional `Eigen::MatrixXd _call_matrix_impl(const std::vector<X>&, const std::vector<Y>&) const`; `CovarianceFunction::operator()` uses it when present (ThreadPool argument deliberately ignored on this path). Bitwise-fallback test guarantees kernels without the method are unaffected.
- **Composition:** `Sum`/`Product` combine per-term blocks (`+` / `cwiseProduct`), so composed kernels benefit term-wise even when some terms are scalar-only.
- **Flagship kernel:** `SquaredExponential<EuclideanDistance>` — pairwise squared distances via one GEMM (`||x||² + ||y||² − 2XᵀY`, clamped at 0 against cancellation) plus one vectorized exp. `double` features use direct differences.
- **Small-n threshold:** below ~16×16 outputs (measured crossover; raw matrix path was 2.5x slower at n=4) evaluation stays per-pair.
- **`Measurement<>`:** unwrapped for leaf kernels; composed kernels take the matrix path on predict (train features are stored unwrapped) and per-pair on fit.

⚠️ `SquaredExponential` covariance values change in the last ulps (GEMM evaluation order). No existing test needed tolerance changes.

## Results (Apple M-series, plain Eigen, -c opt, median of 5)

| n=m | assembly per-pair | assembly matrix | speedup | GP mean predict per-pair | matrix | speedup |
|---|---|---|---|---|---|---|
| 256 | 202.4 µs | 171.3 µs | **1.18x** | 0.203 ms | 0.182 ms | **1.12x** |
| 512 | 821.6 µs | 679.4 µs | **1.21x** | 0.798 ms | 0.724 ms | **1.10x** |
| 1024 | 3.60 ms | 2.73 ms | **1.32x** | 3.91 ms | 2.86 ms | **1.37x** |
| 2048 | 20.94 ms | 11.38 ms | **1.84x** | 22.6 ms | 11.9 ms | **1.90x** |

On this machine the computation is exp-bound (Apple libm exp ≈3 ns/pair; NEON is 2-wide for doubles), capping the gain near 2x. Linux/MKL (wider SIMD, `vdExp`, MKL GEMM) should do considerably better — worth benchmarking there before drawing conclusions.

Note: `//tests:albatross-test-stats-scores` fails identically on the parent commit (pre-existing Monte Carlo flake, unrelated).

Stacked on #591.

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,67 @@
+name: Benchmarks
+
+# Manual benchmark runs on Linux CI runners, primarily to measure the
+# matrix-level kernel assembly path on x86_64 (wider SIMD than Apple
+# NEON) with and without MKL.  Not part of regular CI: shared runners
+# are noisy, so treat results as indicative, and prefer the relative
+# per-pair vs matrix numbers over absolute timings.
+on:
+  push:
+    branches:
+      - 'feat/matrix-kernel-api'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  kernel-assembly:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        flavor:
+          - { name: "eigen", flags: "-c opt" }
+          - { name: "mkl", flags: "--config=mkl" }
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v6
+        with:
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_KEY }}
+
+      - uses: bazel-contrib/setup-bazel@0.19.0
+
+      - uses: extractions/netrc@v3
+        with:
+          machine: raw.githubusercontent.com
+          username: ${{ secrets.GH_NAME }}
+          password: ${{ secrets.GH_TOKEN }}
+
+      - name: Mount bazel cache
+        uses: actions/cache@v5
+        with:
+          path: "~/.cache/bazel"
+          key: bazel-bench-${{ matrix.flavor.name }}
+
+      - name: CPU info
+        run: |
+          lscpu | grep -E 'Model name|^CPU\(s\)' || true
+          grep -o -m1 -E 'avx512[a-z]*|avx2|fma' /proc/cpuinfo | sort -u | tr '\n' ' '; echo
+
+      - name: Run kernel assembly benchmark (${{ matrix.flavor.name }})
+        run: |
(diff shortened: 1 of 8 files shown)


armallen:
Linux CI results (benchmarks workflow, ubuntu-22.04, AMD EPYC 9V74, AVX2+FMA, medians of 5):

| n=m | assembly Eigen | assembly MKL | GP mean predict Eigen | GP mean predict MKL |
|---|---|---|---|---|
| 256 | 625 µs → 419 µs (**1.49x**) | 711 µs → 354 µs (**2.01x**) | 0.648 ms → 0.440 ms (**1.47x**) | 0.714 ms → 0.375 ms (**1.90x**) |
| 512 | 2.51 ms → 1.67 ms (**1.50x**) | 2.86 ms → 1.39 ms (**2.06x**) | 2.58 ms → 1.73 ms (**1.49x**) | 2.86 ms → 1.47 ms (**1.95x**) |
| 1024 | 10.2 ms → 9.4 ms (1.08x) | 11.5 ms → 8.5 ms (1.36x) | 10.5 ms → 6.9 ms (**1.52x**) | 11.5 ms → 5.9 ms (**1.96x**) |
| 2048 | 42.9 ms → 32.4 ms (1.32x) | 50.0 ms → 30.8 ms (1.62x) | 44.0 ms → 33.5 ms (1.31x) | 50.5 ms → 30.9 ms (1.63x) |

Conclusions:
- Realistic gain is **~1.5x (plain Eigen/AVX2) to ~2x (MKL)** — the computation is exp- and memory-bound, not GEMM-bound, so the theoretical 10x for pure-GEMM workloads doesn't apply here. MKL only accelerates the distance GEMM; Eigen's array exp does not route through MKL VML.
- Large-n ratios dip because the matrix path's extra n² temporaries exceed shared-runner cache; per-pair writes K only once.
- Linux crossover is n≈16, same as the macOS-derived threshold already in the code.

.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,67 @@
+name: Benchmarks
```
armallen, line 1: Temporary addition to run benchmarks on linux with and without MKL.

armallen:
Closing as we don't plan to use this

PR #600, opened 2026-08-04
Use format&lint from `rules_lint` and bump deps

martin4861:
Update to newer version of `rules_swiftnav`, which drops support for linting and formatting. Instead use clang-format and tidy from `rules_lint`.


armallen (approved):
🥇

